Issue #2, opened 2026-01-15
Setup PR Test CI workflow


slin1237:
## Description
Create GitHub Actions workflow for PR testing.

## Actions
- [ ] Create `.github/workflows/pr-test-rust.yml`
- [ ] Configure paths filter for root-level structure
- [ ] Set repository check to `lightseekorg/smg`
- [ ] Configure working directories for smg structure
- [ ] Configure artifact paths
- [ ] Review runner requirements (`4-gpu-a10`)
- [ ] Configure dependency install script path

## Workflow Features
- Build wheel
- Python unit tests
- Rust unit tests (lint, fmt, cargo test)
- E2E tests (benchmarks, responses, router, chat-completions)
- Docker build test

PR #36, opened 2026-01-16
Add ci workflow

key4ng:
https://github.com/lightseekorg/smg/issues/2

### DIFF
--- a/.github/workflows/pr-test-rust.yml
+++ b/.github/workflows/pr-test-rust.yml
@@ -208,9 +208,14 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Clone SGLang repository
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,79 @@
+######################## BASE IMAGE ##########################
```
CatherineSue, line 1: Do we have to add dockerfile in this PR? If not, let's move dockerfile to another PR

CatherineSue: And I wonder if we can simply call it Dockerfile as we don't have other dockerfile in this folder yet.

key4ng: the docker build test in the CI workflow depends on this file. sure I'll rename it

PR #114, opened 2026-01-20
Use request-scoped MCP contexts in gRPC responses

zhaowenzi:
Follow up on PR https://github.com/lightseekorg/smg/pull/43
- Cache per-request server_keys inside RequestMcpContext and return them by reference.
- Remove gRPC’s shared requested_servers state and propagate request-scoped MCP context through regular/harmony paths.
- Update tool listing/calls to use request-scoped keys and adjust OpenAI router to the new server_keys() signature.

### DIFF
--- a/model_gateway/src/routers/grpc/regular/responses/non_streaming.rs
+++ b/model_gateway/src/routers/grpc/regular/responses/non_streaming.rs
@@ -20,6 +20,7 @@ use super::{
     conversions,
 };
 use crate::{
+    mcp::RequestMcpContext,
     observability::metrics::{metrics_labels, Metrics},
     protocols::responses::{ResponseStatus, ResponsesRequest, ResponsesResponse},
     routers::{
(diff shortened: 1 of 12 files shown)


model_gateway/src/routers/grpc/regular/responses/non_streaming.rs
```diff
@@ -158,6 +153,7 @@ pub(super) async fn execute_tool_loop(
     headers: Option<http::HeaderMap>,
     model_id: Option<String>,
     response_id: Option<String>,
+    active_mcp: &Arc<RequestMcpContext>,
```
CatherineSue, line 156: This is borderline of too many arguments. I don't like how we not utilize the pipeline context but use threaded arguments here.

model_gateway/src/routers/grpc/regular/responses/streaming.rs
```diff
@@ -716,8 +721,7 @@ async fn execute_tool_loop_streaming_internal(
                     tool_call.arguments
                 );
                 let tool_start = Instant::now();
-                let (output_str, success, error) = match ctx
-                    .mcp_manager
+                let (output_str, success, error) = match active_mcp
```
CatherineSue, line 724: I am not sure I am getting the design here. The previous impl is much cleaner than the API right now. And propagating through indicates it is a bad design.

model_gateway/src/routers/grpc/regular/responses/non_streaming.rs
zhaowenzi: Hi @CatherineSue  How about store `RequestMcpContext` in `RequestContext` (likely under ProcessingState)? ResponsesContext (ctx) is global/shared across requests, so it can’t safely hold request‑scoped `RequestMcpContext`.

CatherineSue: I don't think `RequestContext` is used even in responses pipeline. Doesn't that require another parameter here?

PR #158, opened 2026-01-22
ci: add vLLM gRPC e2e tests

key4ng:
## Description
Add end-to-end tests for vLLM gRPC backend to ensure gateway compatibility with vLLM's gRPC server.
### Problem
The gateway supports vLLM gRPC backend but lacked automated e2e tests to verify functionality and catch regressions.

### Solution
Add comprehensive test suite covering chat completions, streaming, structured output, and health checks for vLLM gRPC backend, with proper CI integration.

## Changes

- Add e2e_test/vllm_grpc/ test suite (basic, streaming, health tests)
- Add vLLM runtime support in runtime_registry.py
- Add vLLM model specs (llama-1b-vllm, llama-8b-vllm)
- Add vllm-grpc job to CI workflow with flash-attn/flashinfer dependencies
- Add request timeouts to prevent test hangs

## Test Plan

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/e2e_test/fixtures/hooks.py
+++ b/e2e_test/fixtures/hooks.py
@@ -119,12 +119,26 @@ def pytest_collection_modifyitems(
 
     for item in items:
         # Extract model from marker or use default
-        # Use get_closest_marker which correctly handles inheritance
-        # (returns marker from child class before parent class)
+        # Walk class MRO to prioritize child class markers over parent class
         model_id = None
-        model_marker = item.get_closest_marker(PARAM_MODEL)
-        if model_marker and model_marker.args:
-            model_id = model_marker.args[0]
+        if hasattr(item, "cls") and item.cls is not None:
+            # Walk the class MRO to find model marker
+            # MRO lists classes from most specific (child) to least specific (parent)
+            for cls in item.cls.__mro__:
+                if hasattr(cls, "pytestmark"):
+                    markers = cls.pytestmark if isinstance(cls.pytestmark, list) else [cls.pytestmark]
+                    for marker in markers:
+                        if marker.name == PARAM_MODEL and marker.args:
+                            model_id = marker.args[0]
+                            break
+                if model_id:
+                    break
+
+        # Fallback to get_closest_marker if not found via MRO
+        if model_id is None:
+            model_marker = item.get_closest_marker(PARAM_MODEL)
+            if model_marker and model_marker.args:
+                model_id = model_marker.args[0]
 
         # Check parametrize for model
         if model_id is None:


slin1237:
instead of another file
can we add additional mark to existing test case, such that it will default to SGL. And spin up vLLM as well if vLLM is defined 
another test file is hard to scale and maintain

e2e_test/vllm_grpc/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""vLLM gRPC E2E Tests.
```
CatherineSue: what is the reason that we can't simply use the same e2e tests for sglang?

e2e_test/infra/model_pool.py
```diff
@@ -226,7 +231,16 @@ def _grpc_health_check(self, timeout: float = 5.0) -> bool:
             finally:
                 channel.close()
         except grpc.RpcError as e:
-            # gRPC-specific errors (connection refused, deadline exceeded, etc.)
+            # Check if the error is UNIMPLEMENTED (server doesn't support health service)
+            if hasattr(e, "code") and e.code() == grpc.StatusCode.UNIMPLEMENTED:
```
CatherineSue, line 235: I am not following this check. We have health check for vllm

key4ng: synced offline and decided to bypass it for now. we will have to add the standard health check support in the vllm repo. refer: https://github.com/sgl-project/sglang/pull/11955

.github/workflows/pr-test-rust.yml
```diff
@@ -202,27 +202,52 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v6
 
-      - name: Clone SGLang repository
+      - name: Install inference backend (SGLang or vLLM)
         run: |
-          git clone https://github.com/sgl-project/sglang.git
+          if [ "${{ matrix.setup_vllm }}" == "true" ]; then
+            echo "Installing vLLM for gRPC testing..."
+            python3 -m pip install vllm
```
CatherineSue, line 234: do we install the latest version of vllm here?

e2e_test/chat_completions/test_function_calling.py
```diff
@@ -536,6 +538,10 @@ def test_streaming_multiple_choices_finish_reason(self, setup_backend):
 
         This tests the fix for the bug where only the last index got a finish_reason chunk.
         """
+
+        if is_vllm():
```
CatherineSue: can we have a configuration somewhere on tests that each backend supports vs doesn't support yet? 

It is a bit hard to manage through if-else in each test case

.github/workflows/pr-test-rust.yml
key4ng: yeah it will install the latest stable version of vllm

e2e_test/chat_completions/test_function_calling.py
key4ng: good point. I removed the flag and try to centralize the skipped tests here:
https://github.com/lightseekorg/smg/blob/cbca2ce5f30222e0e820417439251b7e97a54d3e/.github/workflows/pr-test-rust.yml#L220-L222

PR #192, opened 2026-01-27
run test workflow to cluster 

XinyueZhang369:
## Description

### Problem

Move gpu needed workflow to cluster

### Solution

Move gpu needed workflow to cluster

## Changes

Workflow (pr-test-rust.yml)                                                                                                                                                                                             
                                                                                                                                                                                                                                   
  - Runner migration: Changed all GPU jobs from 4-gpu-a10 to k8s-runner-gpu                                                                                                                                                        
  - Python environment: Added explicit pip and venv installation since k8s runners don't have these pre-installed                                                                                                                  
  - Manylinux compliance: Added --zig flag to maturin build for manylinux_2_28 compatibility                                                                                                                                       
  - CUDA toolkit setup: Added nvcc installation and symlink creation for flash-attn builds on vLLM tests                                                                                                                           
  - Environment variables: Added HF_TOKEN and BRAVE_API_KEY secrets to relevant jobs                                                                                                                                               
  - SGLang installation: Added CUDA_HOME and IS_BLACKWELL=1 flags for proper dependency installation                                                                                                                               
  - Dependencies: Added pandas to e2e test dependencies                                                                                                                                                                            
  - Path fixes: Changed hardcoded /home/ubuntu/ paths to use $HOME for portability                                                                                                                                                 
                                                                                                                                                                                                                                   
  Go Bindings (batch_postprocessor.go)                                                                                                                                                                                             
                                                                                                                                                                                                                                   
  - Added mutex synchronization to prevent concurrent FFI calls causing memory corruption                                                                                                                                          
  - Protected AddChunk, Flush, and Reset methods with mutex locks                                                                                                                                                                  
                                                                                                                                                                                                                                   
  Tokenizer (hub.rs, tokenizer.rs)                                                                                                                                                                                                 
                                                                                                                                                                                                                                   
  - Fixed HuggingFace Hub token handling: explicitly read HF_TOKEN env var since hf-hub crate's from_env() doesn't read it directly                                                                                                
  - Changed Go FFI tokenizer to use create_tokenizer instead of create_tokenizer_from_file to support HF model IDs                                                                                                                 
  - Fixed doc comment indentation to pass clippy lint                                                                                                                                                                              
                                                                                                                                                                                                                                   
  GPU Monitor (gpu_monitor.py)                                                                                                                                                                                                     
                                                                                                                                                                                                                                   
  - Added configurable trim_start and trim_end parameters to exclude startup/shutdown periods from GPU utilization stats                                                                                                           
  - Moved trimming logic to _read_result() so stats are trimmed once when reading, not repeatedly in assertions                                                                                                                    
                                                                                                                                                                                                                                   
  Benchmarks (test_pd_perf.py)                                                                                                                                                                                                     
                                                                                                                                                                                                                                   
  - Increased max_requests_per_run to 200 for longer benchmark runs                                                                                                                                                                
  - Changed GPU utilization threshold from gpu_util_p50_min: 99 to gpu_util_mean_min: 30

## Test Plan

Workflow passed 

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/.github/workflows/pr-test-rust.yml
+++ b/.github/workflows/pr-test-rust.yml
@@ -44,47 +44,13 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Install pip & venv
-        run: |
-          sudo apt update
-          sudo apt install -y python3-pip python3-venv
-
       - name: Setup Rust
         uses: ./.github/actions/setup-rust
 
-      - name: Build Python wheel and Go FFI library (parallel)
+      - name: Build Python wheel and Go FFI library
         run: |
-          source "$HOME/.cargo/env"
-          export RUSTC_WRAPPER=sccache
-          
-          # Create venv
-          python3 -m venv .venv
-          source .venv/bin/activate
-
-          # Install maturin and zig for manylinux cross-compilation
-          python3 -m pip install --upgrade pip maturin ziglang
-
-          # Start Go FFI build in background
-          echo "Starting Go FFI build in background..."
-          (cd bindings/golang && make build && echo "Go FFI: OK" && ls -la target/release/libsmg_go.*) &
-          GO_PID=$!
-
-          # Build Python wheel in foreground
-          echo "Building Python wheel..."
-          cd bindings/python
-          maturin build --profile ci --features vendored-openssl --manylinux 2_28 --zig --out dist
-          echo "Python wheel: OK"
-          ls -lh dist/
-
-          # Wait for Go build to complete
-          echo "Waiting for Go FFI build..."
-          wait $GO_PID
-          GO_EXIT=$?
-          if [ $GO_EXIT -ne 0 ]; then
-            echo "Go FFI build failed with exit code $GO_EXIT"
-            exit $GO_EXIT
-          fi
-          echo "Both builds completed successfully"
+          bash scripts/ci_setup_python_venv.sh
+          bash scripts/ci_build_wheel.sh
 
       - name: Upload wheel artifact
         uses: actions/upload-artifact@v6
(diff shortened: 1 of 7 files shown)


tokenizer/src/hub.rs
```diff
@@ -2,6 +2,22 @@ use std::path::{Path, PathBuf};
...
 use hf_hub::api::tokio::ApiBuilder;
 
+/// Environment variable for HuggingFace token
+/// Note: The hf-hub crate's from_env() doesn't read HF_TOKEN directly,
+/// it only reads from the token file. We explicitly read this env var
+/// to support CI environments where the token is set as an environment variable.
+const HF_TOKEN_ENV: &str = "HF_TOKEN";
+
+/// Build an ApiBuilder with token from HF_TOKEN environment variable if set
+fn build_api() -> anyhow::Result<hf_hub::api::tokio::Api> {
```
CatherineSue, line 12: I'm not sure I like this change. This is changing the behavior of `download_tokenizer_from_hf`.

XinyueZhang369: This should be just passing HF_TOKEN env var to the builder otherwise the tests will fail due to 401 from HF, or maybe I miss something, do you mind helping me understand what behavior will be changed?

.github/workflows/pr-test-rust.yml
```diff
@@ -249,6 +268,46 @@ jobs:
       - name: Install flash-attn for vLLM
         if: matrix.setup_vllm
         run: |
+          # Ensure CUDA toolkit (nvcc) is available for building flash-attn
```
key4ng: this looks complicated in order to install the flash-attn for vllm. We can keep it for now, but I'll investigate if we can get rid of installing it

XinyueZhang369: Cool thanks

tokenizer/src/hub.rs
XinyueZhang369: After sync up, realized when the HF_TOKEN env var is empty, the behavior could be impacted, added a check to handle that case, thanks

PR #227, opened 2026-01-29
feat: Add Messages API foundation

key4ng:
### Problem

SMG currently only supports OpenAI-compatible APIs. Users wanting to use Anthropic's Claude models with the Messages API have no native support, limiting SMG's applicability for Claude-first workflows and missing unique features like extended thinking, citations, and native MCP integration.                                                          

### Solution

This PR adds the foundation for Messages API support by:                                                                                                                       
  - Extending `RouterTrait` with a `route_messages()` method                                                                                                                     
  - Creating the `messages/` module structure under the OpenAI router                                                                                                            
  - Registering the `/v1/messages` endpoint                                                                                                                                      
  - Implementing stub handlers that return 501 Not Implemented                                                                                                                   
  - Adding comprehensive protocol deserialization tests  
This foundation enables subsequent PRs to implement the full Messages API functionality while maintaining backward compatibility with existing routes. 
## Changes
  - **RouterTrait Extension**: Added `route_messages()` method with default 501 implementation                                                                                   
  - **Messages Module**: Created `routers/openai/messages/` with handler, streaming, and tools modules                                                                           
  - **Endpoint Registration**: Registered `/v1/messages` POST route in server                                                                                                    
  - **OpenAI Router Integration**: Implemented routing to streaming/non-streaming handlers based on request                                                                      
  - **Foundation Tests**: Added 9 tests covering request deserialization and tool extraction                                                                                     
  - **Documentation**: Updated CLAUDE.md with Messages API implementation status 

## Test Plan

 cargo test --test messages_test   

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/model_gateway/src/routers/openai/router.rs
+++ b/model_gateway/src/routers/openai/router.rs
@@ -37,7 +37,6 @@ use crate::{
     observability::metrics::{bool_to_static_str, metrics_labels, Metrics},
     protocols::{
         chat::ChatCompletionRequest,
-        messages::CreateMessageRequest,
         responses::{
             generate_id, ResponseContentPart, ResponseInput, ResponseInputOutputItem,
             ResponsesGetParams, ResponsesRequest,
@@ -1037,21 +1036,6 @@ impl crate::routers::RouterTrait for OpenAIRouter {
                 warn!("Failed to retrieve input items for {}: {}", response_id, e);
                 error_responses::internal_error(format!("Failed to retrieve input items: {}", e))
             }
-        }
-    }
-
-    async fn route_messages(
-        &self,
-        headers: Option<&HeaderMap>,
-        body: &CreateMessageRequest,
-        _model_id: Option<&str>,
-    ) -> Response {
-        use crate::routers::anthropic::messages::MessagesHandler;
-
-        if body.stream.unwrap_or(false) {
-            MessagesHandler::handle_streaming(headers, body).await
-        } else {
-            MessagesHandler::handle_non_streaming(headers, body).await
         }
     }
 


model_gateway/src/routers/anthropic/messages/handler.rs
```diff
@@ -0,0 +1,80 @@
+//! Messages API request handler (non-streaming)
```
slin1237, line 1: OpenAI doesn't have message api
message api should be its own folder
for gRPC. It can be part of gRPC folder.

model_gateway/src/routers/openai/router.rs
```diff
@@ -1039,6 +1040,21 @@ impl crate::routers::RouterTrait for OpenAIRouter {
         }
     }
 
+    async fn route_messages(
+        &self,
+        headers: Option<&HeaderMap>,
+        body: &CreateMessageRequest,
+        _model_id: Option<&str>,
+    ) -> Response {
+        use crate::routers::anthropic::messages::MessagesHandler;
```
CatherineSue: let's put imports at the top of the file

model_gateway/src/routers/openai/router.rs
```diff
@@ -1039,6 +1040,21 @@ impl crate::routers::RouterTrait for OpenAIRouter {
...
+
+        if body.stream.unwrap_or(false) {
+            MessagesHandler::handle_streaming(headers, body).await
```
CatherineSue: The dependency here is a bit overlapped. We have an import from anthropic folder?

model_gateway/src/routers/openai/router.rs
key4ng: removed

model_gateway/src/routers/openai/router.rs
key4ng: removed

key4ng:
> The structure is weird to me. Is the goal to support `/v1/message` in openai router?
> 
> Do we have a `anthropic` router now?

@CatherineSue thanks for pointing that. My previous idea was trying to support it in the openai router. But openai router should not support message api. I've created an anthropic router. pls have another look

PR #282, opened 2026-02-03
fix(grpc): correct completion_tokens counting for vLLM harmony streaming 

key4ng:
## Description                                                                                                                                        
                                                                                                                                                        
  ### Problem                                                                                                                                           
```
data: {"id":"chatcmpl-41072669-9437-43fb-a68b-4955bf8f8ec5","object":"chat.completion.chunk","created":1770087544,"model":"/home/ubuntu/models/openai/gpt-oss-20b","system_fingerprint":"default","choices":[],"usage":{"prompt_tokens":71,"completion_tokens":1,"total_tokens":72}}
```                                                                                                                                                        
The Harmony gRPC streaming router always reports `completion_tokens=1` in the usage field for vLLM backends. This occurs because the router uses the `completion_tokens` value from vLLM's `GenerateComplete` message, which is unreliable.                                                       
                                                                                                                                                        
  ### Solution                                                                                                                                          
                                                                                                                                                        
  Apply the same backend-aware token counting logic used in the regular gRPC streaming handler:                                                         
  - For vLLM: Accumulate tokens from `chunk.token_ids().len()` during streaming, ignore the `GenerateComplete` value                                    
  - For SGLang: Use the `completion_tokens` from `GenerateComplete` (SGLang sends correct cumulative values)                                            
                                                                                                                                                        
  ## Changes                                                                                                                                            
                                                                                                                                                        
  - Add vLLM detection in chunk handling to accumulate actual token counts                                                                              
  - Add vLLM detection in Complete handling to preserve accumulated count instead of overwriting                                                        
  - Apply fix to all 5 affected locations: single stream, dual stream, and Responses API streaming                                                      
                                                                                                                                                        
  ## Test Plan                                                                                                                                          
                                                                                                                                                        
  - Added `tests/grpc_harmony_streaming_test.rs` with 10 unit tests covering:                                                                           
    - vLLM chunk token accumulation                                                                                                                     
    - vLLM Complete preserving accumulated count                                                                                                        
    - SGLang using Complete message value                                                                                                               
    - Multi-index (n>1) streaming support                                                                                                               
    - Responses API single-value tracking

<details>
<summary>Checklist</summary>

</details>


model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -716,6 +737,12 @@ impl HarmonyStreamingProcessor {
 
             match response.into_response() {
                 ProtoResponseVariant::Chunk(chunk_wrapper) => {
+                    // Track token counts for vLLM (vLLM sends deltas)
+                    // For SGLang, skip (SGLang sends cumulative values in Complete)
+                    if chunk_wrapper.is_vllm() {
+                        completion_tokens += chunk_wrapper.token_ids().len() as u32;
```
CatherineSue, line 743: Why do we need to add completion_tokens here?

key4ng: This helps fix the response api token usage calculation. Same as the Chat Completion streaming paths - vLLM's GenerateComplete.completion_tokens field may be incorrect

without this change I got
`"usage":{"input_tokens":67,"output_tokens":0,"total_tokens":67}`

and with this change I got 
`"usage":{"input_tokens":67,"output_tokens":34,"total_tokens":101}`

PR #333, opened 2026-02-05
chore: simplify SGLang CI install with explicit dependencies  

key4ng:
### Problem

The previous `ci_install_sglang.sh` cloned the SGLang repo and ran its `ci_install_dependency.sh` script as a black box. This was fragile — whenever SGLang reorganized their repo or moved the script, our CI would break. It also made it unclear what was actually being installed and why, making failures hard to diagnose.

### Solution

Replace the clone-and-run approach with explicit `pip install` commands where every dependency is clearly documented:
- **`sglang[all]`** — installs SGLang with bundled `flashinfer_python` + `flashinfer_cubin` (attention kernels only)
- **`flashinfer-jit-cache`** — pre-compiled trtllm_comm kernels (fused allreduce + layernorm for multi-GPU communication), version auto-matched to `flashinfer_python`
- **`mooncake-transfer-engine`** — KV transfer for PD disaggregation
- **System libs** — InfiniBand/RDMA libraries for mooncake

We should not treat upstream CI scripts as a black box. Every dependency should be explicit so we understand exactly what is installed and why.

## Changes

- Replace git-clone-from-source install with `pip install "sglang[all]"`
- Add `flashinfer-jit-cache` install with version dynamically matched to `flashinfer_python` and CUDA version detected from torch
- Remove dependency on SGLang's `ci_install_dependency.sh`
- Remove old vLLM/xformers cleanup (no longer relevant)

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Chores**
  * Added sentence-transformers to test dependencies.
  * Simplified SGLang installation process using pip.
  * Added flashinfer-jit-cache and mooncake-transfer-engine as dependencies.
  * Updated system package requirements for Oracle Instant Client setup.

### DIFF
--- a/scripts/ci_install_sglang.sh
+++ b/scripts/ci_install_sglang.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
-# Install SGLang and its dependencies for CI
-# Clones SGLang repo and runs the CUDA dependency installation script
+# Install SGLang for CI
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


scripts/ci_install_sglang.sh
```diff
@@ -9,51 +9,16 @@ if [ -f ".venv/bin/activate" ]; then
...
-    LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
-    echo "Checking out latest SGLang tag: $LATEST_TAG"
-    git checkout "$LATEST_TAG"
-    cd -
+# Install uv for faster package management (10-100x faster than pip)
+if ! command -v uv &> /dev/null; then
+    echo "Installing uv..."
+    curl -LsSf https://astral.sh/uv/install.sh | sh
+    export PATH="$HOME/.local/bin:$PATH"
 fi
```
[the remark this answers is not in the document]
key4ng: this path already export in ci node

scripts/ci_install_sglang.sh
```diff
@@ -9,64 +8,32 @@ if [ -f ".venv/bin/activate" ]; then
...
-        git checkout "$DEFAULT_BRANCH"
-        git pull --ff-only
-    else
-        echo "WARNING: Could not determine default branch; staying on current HEAD"
-    fi
+pip install --upgrade pip
+
+# Install SGLang with all dependencies
+echo "Installing SGLang..."
+pip install "sglang[all]"
```
CatherineSue, line 15: With this change, we are not gonna use SGLang main branch from now on. Am I understanding it correctly?

key4ng: Yeah that's correct. it will use stable released version

PR #346, opened 2026-02-06
fix(trtllm): tokenize and inject user stop sequences for TRT-LLM requests

### LINKED ISSUE #221: TRT-LLM: stop sequences (stop strings) not mapped in chat completions
## Problem

In `grpc_client/src/trtllm_service.rs`, the `extract_stop_words()` method is a stub that returns `vec![]`, silently dropping all stop sequences from `ChatCompletionRequest.stop`.

```rust
fn extract_stop_words(&self, request: &ChatCompletionRequest) -> Vec<proto::TokenSequence> {
    // Note: This returns empty because stop words need to be tokenized
    let _ = request;
    vec![]
}
```

Both SGLang and vLLM properly extract stop strings via `extract_stop_strings()` and pass them as `Vec<String>` in their sampling params. TRT-LLM's proto uses `stop_words: repeated TokenSequence` (token ID sequences), which requires tokenization.

## Expected Behavior

Stop sequences from `ChatCompletionRequest.stop` should be tokenized and mapped to `Vec<proto::TokenSequence>` for TRT-LLM requests.

## Options

1. **Tokenize at the router level**: The router has access to the tokenizer. Convert stop strings to token ID sequences before building the TRT-LLM request.
2. **Add text-based stop field to TRT-LLM proto**: Add a `repeated string stop_strings` field so TRT-LLM server can handle tokenization.

## Files

- `grpc_client/src/trtllm_service.rs` — `extract_stop_words()` (lines 480-486)
- `grpc_client/src/trtllm_service.rs` — `build_generate_request_from_chat()` (line 270)

## Related

Part of #220

ppraneth:
### Problem

closes #221 

TRT-LLM requires stop words as tokenized `TokenSequence` entries (token IDs), unlike SGLang/vLLM which handle stop strings at the client level. Previously, the `extract_stop_words` method in `TrtllmServiceClient` was a no-op that always returned an empty vec, meaning user-provided stop sequences were silently ignored.

### **Solution Implementation**

Stop sequences are now tokenized and injected during the **Request Building Stage**, right after the proto request is constructed and before it is dispatched. A single helper function handles both tokenization and injection in one step, keeping the logic encapsulated and avoiding leaky abstractions.

**Key Changes:**

1. **Removed dead code in `grpc_client`**

   * Deleted the no-op `extract_stop_words` method from `TrtllmServiceClient` — it always returned `vec![]` with a comment saying "the router should handle this".
   * The call site now uses `vec![]` directly with a comment explaining that stop words are injected by the router post-build.

2. **Single-responsibility helper: `inject_trtllm_stop_words`**

   * Added `utils::inject_trtllm_stop_words(&mut req, tokenizer, stop)` that takes a mutable reference to the TRT-LLM request, tokenizes each stop string, and pushes `TokenSequence` entries directly — no intermediate allocations or conversions needed at the call site.

3. **Request Building Stages**

   * Both Chat and Generate request-building stages call `inject_trtllm_stop_words` after constructing the proto request, gated on `ProtoGenerateRequest::Trtllm` and the presence of a tokenizer + stop sequences.

4. **E2E Tests**

   * Added `test_stop_sequences` and `test_stop_sequences_stream` to the chat completions e2e test suite, verifying that stop sequences cause generation to halt (both non-streaming and streaming) and that the stop token does not appear in the output.

### **Verification**

* `cargo clippy --all-targets --all-features -- -D warnings` passes
* `cargo +nightly fmt` passes
* E2E tests added for stop sequences (non-streaming and streaming)

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/.github/actions/setup-trtllm/action.yml
+++ b/.github/actions/setup-trtllm/action.yml
@@ -14,8 +14,6 @@ runs:
       with:
         path: /tmp/trtllm-wheel
         key: trtllm-wheel-${{ runner.os }}-${{ hashFiles('scripts/ci_install_trtllm.sh') }}
-        restore-keys: |
-          trtllm-wheel-${{ runner.os }}-
 
     - name: Install TRT-LLM
       shell: bash
--- a/scripts/ci_install_trtllm.sh
+++ b/scripts/ci_install_trtllm.sh
@@ -5,6 +5,8 @@
 # so we build from source (main branch) which compiles the C++
 # extensions properly and includes the gRPC serve command.
 #
+# Cache version: 2 — rebuild to pick up gRPC stop_words fix (#11292)
+#
 # Prerequisites (expected on k8s-runner-gpu nodes):
 #   - NVIDIA driver 580+ (CUDA 13)
 #   - CUDA 13.0 toolkit at /usr/local/cuda-13.0


model_gateway/src/routers/grpc/harmony/stages/preparation.rs
```diff
@@ -123,6 +123,7 @@ impl HarmonyPreparationStage {
             selection_text: Some(build_output.selection_text),
             harmony_messages: Some(build_output.harmony_messages),
             harmony_stop_ids: Some(build_output.stop_token_ids),
+            harmony_stop_sequences: Some(build_output.additional_stop_sequences),
```
CatherineSue: I am not clear why we need this field.

model_gateway/src/routers/grpc/context.rs
```diff
@@ -123,6 +123,9 @@ pub(crate) struct PreparationOutput {
 
     /// Stop token IDs for Harmony models
     pub harmony_stop_ids: Option<Vec<u32>>,
+
+    /// Tokenized stop sequences for regular gRPC models (e.g. TRT-LLM)
+    pub regular_stop_sequences: Option<Vec<Vec<u32>>>,
```
CatherineSue: Why do we still need this in context?

CatherineSue: I don't think we need this field. The tokenizer is already available in `ctx.state.tokenizer` by the time request building runs (the preparation stage is what resolves and stores it there). We can just tokenize the stop sequences inline in the request building stage, right where the TRT-LLM request is being constructed.

Adding this to `PreparationOutput` forces every unrelated preparation stage (harmony, embedding, etc.) to set `regular_stop_sequences: None`, coupling them to a TRT-LLM-specific concern. There's no performance benefit either — tokenizing a few short stop strings is trivial and happens once per request regardless of where in the pipeline it's done.

ppraneth: Thanks for the catch
I've removed that field and moved the tokenization logic directly into ChatRequestBuildingStage and GenerateRequestBuildingStage. Since ctx.state.tokenizer is already available by that point, I’m now tokenizing the stop sequences on-the-fly right where the TRT-LLM request is built.

model_gateway/src/routers/grpc/regular/stages/chat/request_building.rs
```diff
@@ -76,6 +76,30 @@ impl PipelineStage for ChatRequestBuildingStage {
...
+        if let crate::routers::grpc::proto_wrapper::ProtoGenerateRequest::Trtllm(req) =
+            &mut proto_request
+        {
+            // Tokenize stop sequences on-the-fly
+            if let Some(tokenizer) = ctx.tokenizer_arc() {
+                if let Some(user_stops) = crate::routers::grpc::utils::tokenize_stop_sequences(
+                    tokenizer.as_ref(),
+                    &chat_request.stop,
+                ) {
+                    use crate::grpc_client::trtllm_proto::TokenSequence;
```
CatherineSue: Please put use at the top of the file.

model_gateway/src/routers/grpc/regular/stages/chat/request_building.rs
```diff
@@ -76,6 +76,30 @@ impl PipelineStage for ChatRequestBuildingStage {
                 error::bad_request("invalid_request_parameters", format!("Invalid request parameters: {}", e))
             })?;
 
+        // Inject user stop sequences for TRT-LLM (manual injection)
+        if let crate::routers::grpc::proto_wrapper::ProtoGenerateRequest::Trtllm(req) =
+            &mut proto_request
+        {
+            // Tokenize stop sequences on-the-fly
+            if let Some(tokenizer) = ctx.tokenizer_arc() {
+                if let Some(user_stops) = crate::routers::grpc::utils::tokenize_stop_sequences(
```
CatherineSue: Please don't use 4-layer of absolute imports here. Do imports at the top.

model_gateway/src/routers/grpc/regular/stages/generate/request_building.rs
```diff
@@ -75,6 +75,35 @@ impl PipelineStage for GenerateRequestBuildingStage {
...
 
+        // Inject user stop sequences for TRT-LLM (manual injection)
+        if let crate::routers::grpc::proto_wrapper::ProtoGenerateRequest::Trtllm(req) =
+            &mut proto_request
+        {
+            // Tokenize stop sequences on-the-fly
+            if let Some(tokenizer) = ctx.tokenizer_arc() {
+                let stop_sequences = generate_request
+                    .sampling_params
+                    .as_ref()
```
CatherineSue: Unnecessary `.clone()` — this allocates a new `StringOrArray` just to pass it by reference on the next line. Use `.as_ref().and_then(|p| p.stop.as_ref())` instead to avoid the allocation.

model_gateway/src/routers/grpc/regular/stages/generate/request_building.rs
```diff
@@ -75,6 +75,35 @@ impl PipelineStage for GenerateRequestBuildingStage {
...
+                if let Some(user_stops) = crate::routers::grpc::utils::tokenize_stop_sequences(
+                    tokenizer.as_ref(),
+                    &stop_sequences,
+                ) {
+                    use crate::grpc_client::trtllm_proto::TokenSequence;
+                    req.stop_words
+                        .extend(user_stops.iter().map(|seq| TokenSequence {
+                            token_ids: seq.clone(),
+                        }));
+                }
```
CatherineSue, line 89: This `else` branch is dead code. The preparation stage (which runs before request building) resolves the tokenizer and stores it in `ctx.state.tokenizer`. If the tokenizer can't be resolved, preparation fails and request building never executes. By the time we're here, the tokenizer is guaranteed to be present.

CatherineSue:
Hey @ppraneth, thanks for working on this — I appreciate the effort you put in to tackle #221.

After reviewing the changes, I have a few concerns I want to share:

### 1. Leaky abstraction with `tokenize_stop_sequences`

The current approach splits the operation into two parts: `tokenize_stop_sequences()` returns raw `Option<Vec<Vec<u32>>>`, and then every call site has to manually convert those into `TokenSequence` structs:

```rust
req.stop_words
    .extend(user_stops.iter().map(|seq| TokenSequence {
        token_ids: seq.clone(),
    }));
```

This forces the caller to know about `TokenSequence` internals and duplicates the conversion logic in both chat and generate request-building stages. A single helper that takes `&mut TrtllmGenerateRequest` directly (tokenize + inject in one step) would be cleaner — the caller shouldn't need to care about the intermediate representation.

### 2. Unnecessary allocation

`.iter().map(|seq| seq.clone())` copies every `Vec<u32>` for no reason. If the helper owned the injection end-to-end, it could just `push` each `TokenSequence` directly from the encoding result without the extra clone.

### 3. The `resolve_tokenizer` caching early-return

```rust
if let Some(tokenizer) = &ctx.state.tokenizer {
    return Ok(tokenizer.clone());
}
```

This is grafted onto `resolve_tokenizer()` as a cache check, but the function already sets `ctx.state.tokenizer` at the end of its normal flow. The simpler approach is to just access `ctx.state.tokenizer` directly at the injection site (it's already resolved by the preparation stage). No need to modify the shared utility.

---

I know this has taken a fair amount of your time already, and I really do appreciate you digging into this. I had actually been working on the same fix on a separate branch (`chang/trt-stop`) with a slightly different approach that addresses the points above, so I'm going to go ahead and land that version on this PR to keep things moving. Hope that's okay — happy to walk through the changes together if you'd like.

Issue #234, opened 2026-01-30
[Realtime Api] create protocol-realtime-session


pallasathena92:
# Description:
Create session configuration types including audio formats, voice options, modalities, turn detection, and tool definitions.

# File Path:
`protocols/src/realtime/session.rs`

# Prerequisites:
protocol-realtime-mod

# Key Types
- `AudioFormat` - pcm16, g711_ulaw, g711_alaw
- `Voice` - alloy, ash, ballad, coral, echo, sage, shimmer, verse
- `Modality` - text, audio
- `TurnDetection` - server_vad config or disabled
- `InputAudioTranscription` - whisper model config
- `RealtimeTool` - function tool definition with JSON schema
- `ToolChoice` - auto, none, required, or specific function
- `MaxTokens` - "inf" or number
- `SessionConfig` - all optional session settings
- `Session` - full session object from API responses

# Implementation:
```rust
use serde::{Deserialize, Serialize};

/// Audio format for input/output
#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
#[serde(rename_all = "snake_case")]
pub enum AudioFormat {
    #[default]
    Pcm16,
    G711Ulaw,
    G711Alaw,
}

/// Voice options for audio output
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum Voice {
    Alloy,
    Ash,
    Ballad,
    Coral,
    Echo,
    Sage,
    Shimmer,
    Verse,
}

/// Modality types
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum Modality {
    Text,
    Audio,
}

/// Turn detection configuration
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(tag = "type", rename_all = "snake_case")]
pub enum TurnDetection {
    ServerVad {
        #[serde(skip_serializing_if = "Option::is_none")]
        threshold: Option<f32>,
        #[serde(skip_serializing_if = "Option::is_none")]
        prefix_padding_ms: Option<u32>,
        #[serde(skip_serializing_if = "Option::is_none")]
        silence_duration_ms: Option<u32>,
        #[serde(skip_serializing_if = "Option::is_none")]
        create_response: Option<bool>,
    },
    Disabled,
}

/// Input audio transcription config
[...]

PR #364, opened 2026-02-07
feat(realtime api): realtime api session and transcription_session protocols

pallasathena92:
official api: https://platform.openai.com/docs/api-reference/realtime
Fix: https://github.com/lightseekorg/smg/issues/234
## Description
Add OpenAI Realtime Session API protocol types with strict type safety, aligned to the official OpenAPI spec (RealtimeSessionCreateRequestGA / RealtimeSessionCreateResponseGA / RealtimeTranscriptionSessionCreateRequestGA / RealtimeTranscriptionSessionCreateResponseGA).
### Problem
The gateway had no Rust types for the OpenAI Realtime Session API. These types are needed to deserialize customer requests, validate them at the gateway boundary, and forward them to upstream models.

## Changes

New file: `rotocols/src/realtime_session.rs`
#### Core session types

RealtimeSessionConfig — request body for creating a realtime session
RealtimeSessionObject — response with ephemeral client_secret (required per spec)
RealtimeTranscriptionSessionConfig — request for transcription sessions
RealtimeTranscriptionSessionObject — transcription session response
Typed enums for closed/const fields (fail-fast validation at gateway)

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive realtime session support: configurable realtime and transcription sessions; flexible audio input/output formats; live transcription, noise reduction, and turn-detection modes; selectable built-in or custom voices; multiple output modalities; token/truncation controls; integrated tool and connector selection; tracing and debugging options; prompt customization and secure secret handling for session payloads.

### DIFF
--- a/protocols/src/realtime_session.rs
+++ b/protocols/src/realtime_session.rs
@@ -1,0 +1,526 @@
+// OpenAI Realtime Session API types
+// https://platform.openai.com/docs/api-reference/realtime
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use crate::common::{Redacted, ResponsePrompt, ToolReference};
+
+// ============================================================================
+// Session Configuration
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionCreateRequest {
+    #[serde(rename = "type")]
+    pub r#type: RealtimeSessionType,
+    pub audio: Option<RealtimeAudioConfig>,
+    pub include: Option<Vec<RealtimeIncludeOption>>,
+    pub instructions: Option<String>,
+    pub max_output_tokens: Option<MaxOutputTokens>,
+    pub model: Option<String>,
+    #[serde(default = "audio")]
+    pub output_modalities: Option<Vec<OutputModality>>,
+    pub prompt: Option<ResponsePrompt>,
+    pub tool_choice: Option<RealtimeToolChoiceConfig>,
+    pub tools: Option<RealtimeToolsConfig>,
+    pub tracing: Option<RealtimeTracingConfig>,
+    pub truncation: Option<RealtimeTruncation>,
+}
+
+// ============================================================================
+// Session Object
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionCreateResponse {
+    pub client_secret: RealtimeSessionClientSecret,
+    #[serde(rename = "type")]
+    pub r#type: RealtimeSessionType,
+    pub audio: Option<RealtimeAudioConfig>,
+    pub include: Option<Vec<RealtimeIncludeOption>>,
+    pub instructions: Option<String>,
+    pub max_output_tokens: Option<MaxOutputTokens>,
+    pub model: Option<String>,
+    #[serde(default = "audio")]
+    pub output_modalities: Option<Vec<OutputModality>>,
+    pub prompt: Option<ResponsePrompt>,
+    pub tool_choice: Option<RealtimeToolChoiceConfig>,
+    pub tools: Option<Vec<RealtimeToolsConfig>>,
+    pub tracing: Option<RealtimeTracingConfig>,
+    pub truncation: Option<RealtimeTruncation>,
+}
+
+// ============================================================================
+// Transcription Session Configuration
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeTranscriptionSessionCreateRequest {
+    #[serde(rename = "type")]
+    pub r#type: RealtimeTranscriptionSessionType,
+    pub audio: Option<RealtimeTranscriptionSessionAudio>,
+    pub include: Option<Vec<RealtimeIncludeOption>>,
+}
+
+// ============================================================================
+// Transcription Session Object from Create Response
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeTranscriptionSessionCreateResponse {
+    pub id: String,
+    pub object: String,
+    #[serde(rename = "type")]
+    pub r#type: RealtimeTranscriptionSessionType,
+    pub audio: Option<RealtimeTranscriptionSessionResponseAudio>,
+    pub expires_at: Option<i64>,
+    pub include: Option<Vec<RealtimeIncludeOption>>,
+}
+
+// ============================================================================
+// Audio Formats
+// ============================================================================
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum RealtimeAudioFormats {
+    #[serde(rename = "audio/pcm")]
+    Pcm {
+        /// Sample rate. Only 24000 is supported.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        rate: Option<u32>,
+    },
+    #[serde(rename = "audio/pcmu")]
+    Pcmu,
+    #[serde(rename = "audio/pcma")]
+    Pcma,
+}
+
+// ============================================================================
+// Audio Transcription
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AudioTranscription {
+    pub language: Option<String>,
+    pub model: Option<String>,
+    pub prompt: Option<String>,
+}
+
+// ============================================================================
+// Noise Reduction
+// ============================================================================
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum NoiseReductionType {
+    NearField,
+    FarField,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NoiseReduction {
+    #[serde(rename = "type")]
+    pub r#type: NoiseReductionType,
+}
+
+// ============================================================================
+// Turn Detection
+// ============================================================================
+/// Used only for `semantic_vad` mode. The eagerness of the model to respond.
+/// `low` will wait longer for the user to continue speaking, `high` will respond more quickly.
+/// `auto` is the default and is equivalent to `medium`. `low`, `medium`, and `high` have max timeouts of 8s, 4s, and 2s respectively.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum SemanticVadEagerness {
+    Low,
+    Medium,
+    High,
+    #[default]
+    Auto,
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum TurnDetection {
+    #[serde(rename = "server_vad")]
+    ServerVad {
+        create_response: Option<bool>,
+        idle_timeout_ms: Option<u32>,
+        interrupt_response: Option<bool>,
+        prefix_padding_ms: Option<u32>,
+        silence_duration_ms: Option<u32>,
+        threshold: Option<f64>,
+    },
+    #[serde(rename = "semantic_vad")]
+    SemanticVad {
+        create_response: Option<bool>,
+        eagerness: Option<SemanticVadEagerness>,
+        interrupt_response: Option<bool>,
+    },
+}
+
+/// Turn detection for transcription sessions. Only `server_vad` is currently supported.
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum RealtimeTranscriptionSessionTurnDetection {
+    #[serde(rename = "server_vad")]
+    ServerVad {
+        prefix_padding_ms: Option<u32>,
+        silence_duration_ms: Option<u32>,
+        threshold: Option<f64>,
+    },
+}
+
+// ============================================================================
+// Voice
+// ============================================================================
+
+/// Built-in voice name (e.g. "alloy", "ash") or custom voice reference.
+///
+/// Variant order matters for `#[serde(untagged)]`: a bare JSON string (e.g. `"alloy"`)
+/// always deserializes as `BuiltIn`. A JSON object `{"id": "..."}` fails `BuiltIn`
+/// and falls through to `Custom`. The two forms are structurally distinct (string vs
+/// object) per the OpenAI spec, so there is no ambiguity.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Voice {
+    VoiceIDsShared(String),
+    Custom { id: String },
+}
+
+// ============================================================================
+// Output Modality
+// ============================================================================
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum OutputModality {
+    Text,
+    Audio,
+}
+
+fn audio() -> Option<Vec<OutputModality>> {
+    Some(vec![OutputModality::Audio])
+}
+
+// ============================================================================
+// Tracing
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TracingConfig {
+    pub group_id: Option<String>,
+    pub metadata: Option<Value>,
+    pub workflow_name: Option<String>,
+}
+
+/// The tracing mode. Always `"auto"`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum TracingMode {
+    #[serde(rename = "auto")]
+    Auto,
+}
+
+/// Either the string `"auto"` or a granular tracing configuration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum RealtimeTracingConfig {
+    Mode(TracingMode),
+    Config(TracingConfig),
+}
+
+// ============================================================================
+// Connector ID
+// ============================================================================
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ConnectorId {
+    ConnectorDropbox,
+    ConnectorGmail,
+    ConnectorGooglecalendar,
+    ConnectorGoogledrive,
+    ConnectorMicrosoftteams,
+    ConnectorOutlookcalendar,
+    ConnectorOutlookemail,
+    ConnectorSharepoint,
+}
+
+// ============================================================================
+// Tools
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum RealtimeToolsConfig {
+    #[serde(rename = "function")]
+    RealtimeFunctionTool {
+        description: Option<String>,
+        name: Option<String>,
+        parameters: Option<Value>,
+    },
+    #[serde(rename = "mcp")]
+    McpTool {
+        server_label: String,
+        allowed_tools: Option<McpAllowedTools>,
+        authorization: Option<Redacted>,
+        connector_id: Option<ConnectorId>,
+        headers: Option<HashMap<String, Redacted>>,
+        require_approval: Option<McpToolApproval>,
+        server_description: Option<String>,
+        server_url: Option<String>,
+    },
+}
+
+// ============================================================================
+// MCP Tool Filter
+// ============================================================================
+
+/// List of allowed tool names or a filter object.
+///
+/// Variant order matters for `#[serde(untagged)]`: serde tries `List` first
+/// (JSON array). A JSON object falls through to `Filter`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum McpAllowedTools {
+    List(Vec<String>),
+    Filter(McpToolFilter),
+}
+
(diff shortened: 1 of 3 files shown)


protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,488 @@
...
+
+// ============================================================================
+// Tool Choice
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum RealtimeToolChoiceSpecific {
+    #[serde(rename = "function")]
```
CatherineSue: do we need rename explicitly here? Can we use snake_case?

CatherineSue:
The fmt is broken. Can you run `cargo +nightly fmt`?

CatherineSue:
nit: add fixes issues in PR description

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+// ============================================================================
+// Session Configuration (RealtimeSessionCreateRequestGA)
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionConfig {
```
CatherineSue: The docs at L10 says this is `RealtimeSessionCreateRequest`, but the name here is `RealtimeSessionConfig`, are they two the same thing?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    #[serde(rename = "type")]
```
CatherineSue, line 18: The fields here are not following the API doc order nor alphabetical order.

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    pub r#type: RealtimeSessionType,
+    #[serde(default = "default_output_modalities")]
```
CatherineSue: The official API says it defaults to `"audio"`

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    Auto,
+    Required,
+}
+
+/// `"none"`, `"auto"`, `"required"`, or a specific function/MCP tool reference.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum RealtimeToolChoice {
+    Mode(RealtimeToolChoiceMode),
+    Specific(RealtimeToolChoiceSpecific),
```
CatherineSue: Where is `specific` in the offical API? I can't seem to find it. https://developers.openai.com/api/reference/resources/realtime#(resource)%20realtime%20%3E%20(model)%20realtime_tool_choice_config%20%3E%20(schema)

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    pub r#type: RealtimeSessionType,
+    #[serde(default = "default_output_modalities")]
+    pub output_modalities: Vec<OutputModality>,
+    pub model: Option<String>,
+    pub instructions: Option<String>,
+    pub audio: Option<RealtimeSessionAudioConfig>,
+    pub include: Option<Vec<RealtimeIncludeOption>>,
+    pub tracing: Option<Tracing>,
+    pub tools: Option<Vec<RealtimeTool>>,
+    pub tool_choice: Option<RealtimeToolChoice>,
```
CatherineSue: Should this be `RealtimeToolChoiceConfig`, as in the official API?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    #[serde(rename = "type")]
+    pub r#type: RealtimeSessionType,
+    #[serde(default = "default_output_modalities")]
+    pub output_modalities: Vec<OutputModality>,
+    pub model: Option<String>,
+    pub instructions: Option<String>,
+    pub audio: Option<RealtimeSessionAudioConfig>,
+    pub include: Option<Vec<RealtimeIncludeOption>>,
+    pub tracing: Option<Tracing>,
+    pub tools: Option<Vec<RealtimeTool>>,
```
CatherineSue: Should this be `RealtimeToolsConfig`? Do we want to be consistent with the official API definition or follow some other standard?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    #[serde(rename = "mcp")]
+    Mcp {
+        server_label: String,
+        server_url: Option<String>,
+        connector_id: Option<ConnectorId>,
+        authorization: Option<String>,
+        server_description: Option<String>,
+        headers: Option<HashMap<String, String>>,
+        allowed_tools: Option<Value>,
+        require_approval: Option<Value>,
```
CatherineSue: I assume the validation will be add in a follow-up PR, right? The official APIs has it to be `optional object { always, never } or "always" or "never"`

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    },
+}
+
+impl std::fmt::Debug for RealtimeTool {
```
CatherineSue: What is this function for?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionConfig {
+    #[serde(rename = "type")]
+    pub r#type: RealtimeSessionType,
+    #[serde(default = "default_output_modalities")]
+    pub output_modalities: Vec<OutputModality>,
+    pub model: Option<String>,
+    pub instructions: Option<String>,
+    pub audio: Option<RealtimeSessionAudioConfig>,
```
CatherineSue: Same here, in official API it is called `RealtimeAudioConfig`.

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+pub struct RealtimeAudioOutputConfig {
+    pub format: Option<RealtimeAudioFormat>,
+    pub voice: Option<Voice>,
+    pub speed: Option<f64>,
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionAudioConfig {
+    pub input: Option<RealtimeAudioInputConfig>,
```
CatherineSue: It seems the official API doc calls it `RealtimeAudioConfigInput`

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// ============================================================================
+// Audio Configuration
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeAudioInputConfig {
+    pub format: Option<RealtimeAudioFormat>,
```
CatherineSue: It seems the official name is `RealtimeAudioFormats`.

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    Auto,
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum TurnDetection {
+    #[serde(rename = "server_vad")]
+    ServerVad {
+        threshold: Option<f64>,
```
CatherineSue: nit: this field also has a valid range, can be added in a follow up PR.

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+// ============================================================================
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum SemanticVadEagerness {
+    Low,
+    Medium,
+    High,
+    #[default]
+    Auto,
```
CatherineSue, line 148: In the API note it says "`auto` is the fault and is equivalent to `medium`". Do we plan to handle it in router logic?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+    pub prompt: Option<Prompt>,
+}
+
+// ============================================================================
+// Session Object (RealtimeSessionCreateResponseGA)
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionObject {
```
CatherineSue: Same here: should this be `RealtimeSessionCreateResponse`?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+/// Built-in voice name (e.g. "alloy", "ash") or custom voice reference.
+///
+/// Variant order matters for `#[serde(untagged)]`: a bare JSON string (e.g. `"alloy"`)
+/// always deserializes as `BuiltIn`. A JSON object `{"id": "..."}` fails `BuiltIn`
+/// and falls through to `Custom`. The two forms are structurally distinct (string vs
+/// object) per the OpenAI spec, so there is no ambiguity.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Voice {
+    BuiltIn(String),
```
CatherineSue: Should we call this `VoiceIDsShared` and `ID` instead?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+// ============================================================================
+// Prompt
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Prompt {
+    pub id: String,
+    pub version: Option<String>,
+    pub variables: Option<Value>,
```
CatherineSue: Do we plan to add `ResponseInputText`, `ResponseInputImage`, and `ResponseInputFile`?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// ============================================================================
+// Transcription Session Configuration
+// (RealtimeTranscriptionSessionCreateRequestGA)
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeTranscriptionSessionConfig {
```
CatherineSue: Same here: should it be `RealtimeTranscriptionSessionCreateRequest`?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// ============================================================================
+// Transcription Session Object
+// (RealtimeTranscriptionSessionCreateResponseGA)
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeTranscriptionSessionObject {
```
CatherineSue: Same here, should this be `RealtimeTranscriptionSessionCreateResponse`?

protocols/src/realtime_session.rs
pallasathena92: yes, they are the same thing

protocols/src/realtime_session.rs
pallasathena92: i can update it to RealtimeSessionCreateResponse

protocols/src/realtime_session.rs
pallasathena92: they updated api spec recently. previous, the output modalities is an array rather than string

protocols/src/realtime_session.rs
pallasathena92: they updated the spec recently.

protocols/src/realtime_session.rs
pallasathena92: will take a note

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,630 @@
...
+#[serde_with::skip_serializing_none]
+#[derive(Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum RealtimeToolsConfig {
+    #[serde(rename = "function")]
+    RealtimeFunctionTool {
+        description: Option<String>,
+        name: Option<String>,
+        parameters: Option<Value>,
+    },
```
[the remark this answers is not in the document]
pallasathena92: definite it as option according api spec

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,630 @@
...
+        silence_duration_ms: Option<u32>,
+        threshold: Option<f64>,
+    },
+    #[serde(rename = "semantic_vad")]
+    SemanticVad {
+        create_response: Option<bool>,
+        eagerness: Option<SemanticVadEagerness>,
+        interrupt_response: Option<bool>,
+    },
+}
```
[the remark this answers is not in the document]
pallasathena92: the turn_detection is optional. it can be turn off with setting to "null"

protocols/src/realtime_session.rs
pallasathena92: updated

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,643 @@
...
+// Session Object (RealtimeSessionCreateResponseGA)
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionCreateResponse {
+    pub client_secret: RealtimeSessionClientSecret,
+    #[serde(rename = "type")]
+    pub r#type: RealtimeSessionType,
+    pub audio: Option<RealtimeAudioConfig>,
```
CatherineSue, line 44: any reason we replaced the previous flatten with the exact fields now?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,643 @@
...
+// Audio Configuration
+// ============================================================================
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RealtimeAudioConfigInput {
+    pub format: Option<RealtimeAudioFormats>,
+    pub noise_reduction: Option<NoiseReduction>,
+    pub transcription: Option<AudioTranscription>,
+    pub turn_detection: Option<TurnDetection>,
```
CatherineSue, line 457: I saw that there is a [`RealtimeTranscriptionSessionTurnDetection`](https://developers.openai.com/api/reference/resources/realtime#(resource)%20realtime.client_secrets%20%3E%20(model)%20realtime_transcription_session_turn_detection%20%3E%20(schema)) and a `RealtimeAudioInputTurnDetection`

They have different fields. 

It seems we treat them as the same thing here?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,643 @@
...
+// ============================================================================
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct RealtimeSessionClientSecret {
+    pub expires_at: i64,
+    pub value: String,
+}
+
+impl std::fmt::Debug for RealtimeSessionClientSecret {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
CatherineSue: Does this fmt serve for the data security concerns?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,643 @@
...
+}
+
+// ============================================================================
+// Prompt
+// ============================================================================
+
+/// Reference to a prompt template and its variables.
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ResponsePrompt {
```
CatherineSue: Can we not import something exist in responses protocol here?

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,643 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ToolChoiceFunctionType {
+    #[serde(rename = "function")]
+    Function,
+}
+
+/// Force the model to call a specific tool on a remote MCP server.
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ToolChoiceMcp {
```
CatherineSue: Same here. Do we have same or similar definitions in responses that we can reuse here? @key4ng Asking this because in OpenAI's official API docs, they seem to be reused.

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,643 @@
...
+                .field(
+                    "authorization",
+                    &authorization.as_ref().map(|_| "[REDACTED]"),
+                )
+                .field("server_description", server_description)
+                .field("headers", &headers.as_ref().map(|_| "[REDACTED]"))
+                .field("allowed_tools", allowed_tools)
+                .field("require_approval", require_approval)
+                .finish(),
+        }
```
CatherineSue: I wonder how many Redacted we need for Realtime APIs? Do similar `fmt` functions exist in other realtime API specs?

If so, can we add a new type based on string instead? The `fmt` function is a bit boilerplate.

```
/// A string that redacts itself in Debug output.                                                                                                                            
#[derive(Clone, Serialize, Deserialize)]                                                                                                                                   
pub struct Redacted(pub String);

impl std::fmt::Debug for Redacted {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str("[REDACTED]")
    }
}
```

protocols/src/realtime_session.rs
pallasathena92: what you mean previous flatten?
the audio only update the name from RealtimeSessionAudioConfig toRealtimeAudioConfig

protocols/src/realtime_session.rs
pallasathena92: I didn't find prompt in responses protocol, would you show me which one I can reuse here?

protocols/src/realtime_session.rs
pallasathena92: not too many. special for some secret or sensitive data

protocols/src/realtime_session.rs
pallasathena92:
```
pub enum ToolReference {
    /// Reference to a function tool
    #[serde(rename = "function")]
    Function { name: String },

    /// Reference to an MCP tool
    #[serde(rename = "mcp")]
    Mcp {
        server_label: String,
        #[serde(skip_serializing_if = "Option::is_none")]
        name: Option<String>,
    },
```
a little bit different here. the definitions in common doesn't include type

protocols/src/realtime_session.rs
CatherineSue: I saw they are used in the following PRs as well. Can we consider the benefit of introducing a new type? Because that would help to eliminate all the `fmt` functions like this one above.

protocols/src/realtime_session.rs
CatherineSue: https://github.com/lightseekorg/smg/compare/7c4fe0ca5b9a8c7badb0c11ae8d52f85f6cfe4d2..a68799f7785aab6b9524e93a436b6891bf33f3d6#diff-952d43dd0a9c67cae80f917275537e95d8d4a73a3fa8324926f45b3dfbb56b1bR69

`#[serde(flatten)]`

protocols/src/realtime_session.rs
key4ng: Yeah it looks the same. It will add the `type: mcp`  during serialization because of this line
https://github.com/lightseekorg/smg/blob/be80611686087e7d12a540939b3ca38287a767cf/protocols/src/common.rs#L310

pallasathena92: ohohoh, missing this line.
yeah, it is same. I can import it here

protocols/src/realtime_session.rs
pallasathena92: 1. to follow the api spec order.
2. if both the container and the flattened struct have a field with the same key, behavior is undefined.

protocols/src/realtime_session.rs
pallasathena92: there are four turn_detection in api spec

- RealtimeSessionCreateRequest:

      - audio:
           - optional RealtimeAudioInputTurnDetection
- RealtimeTranscriptionSessionCreateRequest：

      - audio:
            - optional RealtimeTranscriptionSessionAudioInputTurnDetection`
- RealtimeSessionCreateResponse：

      -  audio:
             - optional object { type, create_response, idle_timeout_ms, 4 more }  or object { type, create_response, eagerness, interrupt_response }
- RealtimeTranscriptionSessionCreateResponse：

      - audio:
             - optional RealtimeTranscriptionSessionTurnDetection

only `RealtimeTranscriptionSessionTurnDetection` is different with others.
I plan to treat `RealtimeAudioInputTurnDetection`, `RealtimeTranscriptionSessionAudioInputTurnDetection` as the same. Meanwhile, treat `RealtimeTranscriptionSessionTurnDetection` differently.

pallasathena92: for current update.
the audio in `RealtimeSessionCreateRequest`,` RealtimeTranscriptionSessionCreateRequest`, `RealtimeSessionCreateResponse` use TurnDetection
the audio in `RealtimeTranscriptionSessionCreateResponse` use RealtimeTranscriptionSessionTurnDetection https://github.com/lightseekorg/smg/pull/364/changes#diff-952d43dd0a9c67cae80f917275537e95d8d4a73a3fa8324926f45b3dfbb56b1bR506

protocols/src/realtime_session.rs
```diff
@@ -0,0 +1,539 @@
...
+    Low,
+    Medium,
+    High,
+    #[default]
+    Auto,
+}
+
+impl SemanticVadEagerness {
+    /// Resolves `Auto` to `Medium` per the OpenAI spec.
+    pub fn resolve(&self) -> &Self {
```
CatherineSue: Should we let this be handled by openai engine backend? Do we plan to let openai router calls this function?

pallasathena92: have considered it.
as your suggestion, I will let openai engine backed handle this

Issue #238, opened 2026-01-30
[Realtime Api] create protocol-realtime-server-events and protocol-realtime-client-event


pallasathena92:
# Description:
Create all 28 server event types received from the Realtime API.

# File Path:
`protocols/src/realtime_events.rs`

# Prerequisites:
protocol-realtime-session, protocol-realtime-conversation, protocol-realtime-response

# Event Categories:
- Error (1): `error`
- Session (2): `session.created`, `session.updated`
- Conversation (5): `conversation.created`, `conversation.item.created`, `conversation.item.input_audio_transcription.completed`, `conversation.item.input_audio_transcription.failed`, `conversation.item.truncated`, `conversation.item.deleted`
- Input Audio Buffer (4): `input_audio_buffer.committed`, `input_audio_buffer.cleared`, `input_audio_buffer.speech_started`, `input_audio_buffer.speech_stopped`
- Response (6): `response.created`, `response.done`, `response.output_item.added`, `response.output_item.done`, `response.content_part.added`, `response.content_part.done`
- Streaming Deltas (8): `response.text.delta`, `response.text.done`, `response.audio_transcript.delta`, `response.audio_transcript.done`, `response.audio.delta`, `response.audio.done`, `response.function_call_arguments.delta`, `response.function_call_arguments.done`
- Rate Limits (1): `rate_limits.updated`

# Implementation:
```rust
use serde::{Deserialize, Serialize};
use super::{Session, ConversationItem, Response, ContentPart};

/// API error details
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct ApiError {
    #[serde(rename = "type")]
    pub error_type: String,
    pub code: Option<String>,
    pub message: String,
    pub param: Option<String>,
    pub event_id: Option<String>,
}

/// Conversation object
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct Conversation {
    pub id: String,
    pub object: String,
}

/// Rate limit info
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct RateLimit {
    pub name: String,
    pub limit: u32,
    pub remaining: u32,
    pub reset_seconds: f32,
}

/// All server events received from the Realtime API
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(tag = "type")]
pub enum RealtimeServerEvent {
    // === Error ===
    #[serde(rename = "error")]
    Error {
        event_id: String,
        error: ApiError,
    },

    // === Session Events ===
[...]

PR #389, opened 2026-02-09
feat(realtime-api): realtime api wire-formated event protocol

pallasathena92:
official api spec: https://developers.openai.com/api/reference/resources/realtime
Fix: https://github.com/lightseekorg/smg/issues/238
### Problem

The Realtime API protocol layer currently defines conversation items, response objects, and session configuration, but lacks the wire-format event types needed to actually exchange messages over WebSocket/WebRTC/SIP connections. Without these, there is no typed representation for the client-to-server and server-to-client event framing that the OpenAI Realtime API requires.

**Key properties:**
Bidirectional — you need to both serialize (server→client) and deserialize (client→server). With json!() you can only serialize; you'd still need structs for parsing incoming client events.
The server must parse incoming events — session.update, input_audio_buffer.append, conversation.item.create all have different payload shapes. We need typed deserialization to dispatch on them.
Output shapes vary wildly — session.created carries a full session object, output_text.delta carries a string delta, rate_limits.updated carries rate limit info, error carries an error object. These are not variations of one struct — they're completely different payloads.
Dispatch is required — the WebSocket handler receives arbitrary messages and must pattern-match: "is this a session update? an audio buffer append? a response cancel?" Without typed events, we'd be doing json["type"].as_str() string matching everywhere.
Scale — you have ~40+ event types across RealtimeClientEvent and RealtimeServerEvent. Ad-hoc JSON at that scale becomes a maintenance burden where typos and missing fields are only caught at runtime.

### Solution

Add a complete, spec-faithful implementation of the OpenAI Realtime API wire-format event protocol (realtime_events.rs), covering all 12 client events and 46 server events defined in the [OpenAI Realtime API reference](https://platform.openai.com/docs/api-reference/realtime). Events are modeled as serde-tagged enums discriminated by the JSON type field, with ClientEvent deriving Deserialize and ServerEvent deriving Serialize to match their data-flow direction.
## Changes
- realtime_events.rs (new, 902 lines): Core event protocol definitions including:
  - ClientEvent enum (12 variants): session.update, conversation.item.{create,delete,retrieve,truncate}, input_audio_buffer.{append,clear,commit}, output_audio_buffer.clear, response.{cancel,create}, transcription_session.update
  - ServerEvent enum (46 variants): session lifecycle, conversation item CRUD, input audio transcription (completed/delta/failed/segment), input/output audio buffer events, response lifecycle and streaming (text, audio, transcript, function calls, MCP calls), rate limits, and errors
  - SessionConfig tagged union ("realtime" / "transcription") for session.update/session.created/session.updated
  - Supporting types: LogProb, TranscriptionUsage (tokens/duration), TranscriptionError, RateLimitInfo, RealtimeError, ResponseContentPart, ConversationInfo
  - event_type() helper and From impl for mapping to event_types constants
  - Large payloads (SessionConfig, RealtimeResponse, ResponseCreateParams) are Box-ed to keep enum size ~224 bytes

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Full OpenAI Realtime API support: bidirectional conversations, live event streaming, structured realtime responses, and session management.
  * Multi-modal content (text, audio, images), advanced audio/transcription controls (noise reduction, turn detection), configurable output modalities and voices.
  * Tool/MCP integration and flexible tool selection, detailed response usage and token/truncation controls, and safer logging with redacted secrets.

### DIFF
--- a/protocols/src/event_types.rs
+++ b/protocols/src/event_types.rs
@@ -337,7 +337,6 @@ pub enum RealtimeClientEvent {
     ResponseCancel,
     ResponseCreate,
     SessionUpdate,
-    TranscriptionSessionUpdate,
 }
 
 impl RealtimeClientEvent {
@@ -352,7 +351,6 @@ impl RealtimeClientEvent {
     pub const RESPONSE_CANCEL: &'static str = "response.cancel";
     pub const RESPONSE_CREATE: &'static str = "response.create";
     pub const SESSION_UPDATE: &'static str = "session.update";
-    pub const TRANSCRIPTION_SESSION_UPDATE: &'static str = "transcription_session.update";
 
     pub const fn as_str(self) -> &'static str {
         match self {
@@ -367,7 +365,6 @@ impl RealtimeClientEvent {
             Self::ResponseCancel => Self::RESPONSE_CANCEL,
             Self::ResponseCreate => Self::RESPONSE_CREATE,
             Self::SessionUpdate => Self::SESSION_UPDATE,
-            Self::TranscriptionSessionUpdate => Self::TRANSCRIPTION_SESSION_UPDATE,
         }
     }
 }
--- a/protocols/src/realtime_events.rs
+++ b/protocols/src/realtime_events.rs
@@ -1,0 +1,875 @@
+// OpenAI Realtime API wire-format event types
+// https://platform.openai.com/docs/api-reference/realtime
+//
+// This module defines the serializable/deserializable event structures
+// for both client-to-server and server-to-client messages sent over
+// WebSocket, WebRTC, or SIP connections.
+//
+// Session configuration types live in `realtime_session`.
+// Conversation item types live in `realtime_conversation`.
+// Response and usage types live in `realtime_response`.
+// Event type string constants live in `event_types`.
+
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    event_types::{RealtimeClientEvent, RealtimeServerEvent},
+    realtime_conversation::RealtimeConversationItem,
+    realtime_response::{RealtimeResponse, RealtimeResponseCreateParams},
+    realtime_session::{RealtimeSessionCreateRequest, RealtimeTranscriptionSessionCreateRequest},
+};
+
+// ============================================================================
+// Client Events
+// ============================================================================
+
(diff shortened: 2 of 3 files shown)


protocols/src/event_types.rs
```diff
@@ -352,7 +351,6 @@ impl RealtimeClientEvent {
     pub const RESPONSE_CANCEL: &'static str = "response.cancel";
     pub const RESPONSE_CREATE: &'static str = "response.create";
     pub const SESSION_UPDATE: &'static str = "session.update";
-    pub const TRANSCRIPTION_SESSION_UPDATE: &'static str = "transcription_session.update";
```
pallasathena92, line 355: this event didn't exist in current api spec

protocols/src/realtime_events.rs
```diff
@@ -0,0 +1,875 @@
...
+// ============================================================================
+// Session Config Union
+// ============================================================================
+
+/// Union of realtime and transcription session configurations.
+///
+/// Discriminated by the `type` field: `"realtime"` or `"transcription"`.
+/// Used by `session.update`, `session.created`, and `session.updated` events.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
```
key4ng, line 775: will this produces duplicate "type" keys?

RealtimeSessionCreateRequest already has `pub r#type: RealtimeSessionType` which serializes as `"type": "realtime"`. Adding `#[serde(tag = "type")]` here causes serde to inject `"type": "realtime"` a second time. pls double check

pallasathena92: #[serde(tag = "type")] on SessionConfig means serde uses the type field to decide which variant (Realtime or Transcription) to pick. Once it picks, "Realtime", it then deserializes the remaining fields of that same JSON object into RealtimeSessionCreateRequest.

The JSON always has a single "type" key. Serde's internally tagged enum reads it non-destructively — it peeks at "type" to choose the variant, then passes the full object (with "type" still present) to the inner struct's deserializer.

For serialization it's the same — serde writes "type" once from the tag, and the inner struct's r#type field is effectively shadowed (the tag wins). So the output JSON still has exactly one "type" key.

PR #405, opened 2026-02-11
fix(openai): /v1/models returns empty for external workers with lazy discovery

CatherineSue:
### Problem

When using the OpenAI backend (`--backend openai`), the `/v1/models` endpoint returns `{"object":"list","data":[]}` even though model refresh successfully discovers 117 models from the upstream OpenAI API.

**Root cause**: `get_models()` in the OpenAI router calls `refresh_external_models()` which fetches models from the backend and stores them via `worker.set_models()` into `models_override` (a `RwLock`). However, it then iterates `worker.models()` to build the response, which reads from `metadata.models` (always empty for wildcard/external workers) — not from `models_override`.

This bug only affects the OpenAI router. gRPC/HTTP routers proxy `/v1/models` directly to the backend, and their workers have `metadata.models` populated at startup during model discovery.

### Solution

Return discovered models directly from `refresh_worker_models()` and `refresh_external_models()` instead of reading them back via `worker.models()`. The `get_models()` handler now uses the returned models to build the response. `set_models()` is still called so that `supports_model()` works correctly for routing.

No trait changes needed — the fix is entirely localized to the OpenAI router.

## Changes

- `refresh_worker_models()`: returns `Vec<ModelCard>` instead of `bool`
- `refresh_external_models()`: collects and returns all discovered models from workers
- `get_models()`: iterates returned models directly instead of reading from `worker.models()`

## Test Plan

- Verified locally with `--backend openai --worker-urls https://api.openai.com --disable-health-check`:
  - **Before fix**: `curl /v1/models` returns `{"data":[]}` despite logs showing `Model refresh: found 117 models`
  - **After fix**: `curl /v1/models` returns all 117 models

Screenshot 2026-02-11 at 1 36 48 PM

- E2E test coverage needed: add an e2e test for `/v1/models` with the OpenAI backend to prevent regression

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Refactor**
  * Optimized model listing and fetching with improved deduplication by model ID and aliases
  * Added primary_model field to model responses for enhanced model relationship clarity

### DIFF
--- a/model_gateway/src/routers/openai/router.rs
+++ b/model_gateway/src/routers/openai/router.rs
@@ -140,7 +140,7 @@ impl OpenAIRouter {
         &self,
         worker: &Arc<dyn Worker>,
         auth_header: Option<&HeaderValue>,
-    ) -> bool {
+    ) -> Vec<ModelCard> {
         let url = format!("{}/v1/models", worker.url());
         let mut backend_req = self.shared_components.client.get(&url);
         if let Some(auth) = auth_header {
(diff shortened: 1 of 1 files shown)


CatherineSue:
Will wait for worker refactor. Close this PR for now.

PR #455, opened 2026-02-18
fix: gRPC channel pool to eliminate single-connection bottleneck

ppraneth:
### Problem

Each gRPC client (`SglangSchedulerClient`, `VllmEngineClient`, `TrtllmServiceClient`) uses a single `tonic::Channel`  meaning one TCP connection per client. Under high concurrency, this causes:

* **HTTP/2 Head-of-Line blocking** — multiplexed streams on one connection contend for the same TCP send buffer
* **TCP congestion window limits** — a single connection's cwnd caps( congestion window ) throughput regardless of available bandwidth
* **Flow control contention** — HTTP/2 connection-level flow control becomes a bottleneck when many streams compete

---

### Solution

Replace the single `tonic::Channel` with a **pool of 4 channels** (4 independent TCP connections) per client, using lock-free round-robin (`AtomicUsize`) distribution. Each RPC call picks the next channel from the pool, spreading concurrent requests across multiple TCP connections.

* Zero public API changes the pool is entirely internal
* All clones of a client share the same pool via `Arc<ChannelPool>`
* `AbortOnDropStream` correctly handles the shared pool on drop

---

## Changes

* Added `ChannelPool` struct
* Replaced `client` field with `Arc<ChannelPool>`
* Added `next_client()` round-robin selector
* Updated all RPC methods across the three gRPC clients
* Added new benchmark with in-process mock gRPC server for measuring throughput/latency
* Added `tokio-stream` dev-dependency and benchmark entry

---

### Benchmark

`cargo bench --package smg --bench grpc_channel_bench`

---

### Baseline (single channel)

| Concurrency | Total Requests | Throughput (req/s) | p50 (ms) | p95 (ms) | p99 (ms) | Errors | Max (ms) |
| :---------: | :------------: | :----------------: | :------: | :------: | :------: | :----: | :------: |
|      1      |       20       |       922.70       |   1.08   |   1.13   |   1.13   |    0   |   1.13   |
|      10     |       200      |       3789.41      |   2.57   |   3.06   |   3.40   |    0   |   3.44   |
|      50     |      1000      |       5765.98      |   8.24   |   11.55  |   14.89  |    0   |   16.33  |
|     100     |      2000      |       5894.58      |   16.04  |   23.05  |   32.35  |    0   |   36.42  |
|     200     |      4000      |       5765.76      |   33.67  |   43.93  |   50.50  |    0   |   56.06  |
|     500     |      10000     |       5534.75      |   86.49  |  109.73  |  125.73  |    0   |  142.30  |

---

### Optimized (4-channel pool)

| Concurrency | Total Requests | Throughput (req/s) | p50 (ms) | p95 (ms) | p99 (ms) | Errors | Max (ms) |
| :---------: | :------------: | :----------------: | :------: | :------: | :------: | :----: | :------: |
|      1      |       20       |       943.65       |   1.06   |   1.10   |   1.10   |    0   |   1.10   |
|      10     |       200      |       4197.76      |   2.31   |   2.79   |   2.87   |    0   |   2.88   |
|      50     |      1000      |       9498.41      |   5.00   |   6.95   |   7.82   |    0   |   8.31   |
|     100     |      2000      |      12067.40      |   7.90   |   11.04  |   11.29  |    0   |   12.32  |
|     200     |      4000      |      17693.47      |   10.68  |   14.78  |   15.62  |    0   |   17.54  |
|     500     |      10000     |      13362.01      |   34.78  |   50.42  |   60.02  |    0   |   69.98  |

---

## Key Improvements

| Metric          | 100 concurrency | 200 concurrency | 500 concurrency |
| --------------- | :-------------: | :-------------: | :-------------: |
| **Throughput**  |      +2.0×      |      +3.1×      |      +2.4×      |
| **p50 latency** |       −51%      |       −68%      |       −60%      |
| **p99 latency** |       −65%      |       −69%      |       −52%      |

---

At higher concurrency levels (100–500), the channel pool significantly improves throughput and reduces tail latency, while maintaining negligible overhead at low concurrency.

<details>
<summary>Checklist</summary>

</details>

## Release Notes

* **New Features**
  * Implemented gRPC channel pooling across scheduler clients to distribute load across multiple TCP connections, improving throughput and reducing latency.
  * Added async constructors for gRPC client initialization.

* **Tests**
  * Added gRPC channel benchmark tool for performance evaluation and monitoring.

### DIFF
--- a/model_gateway/benches/grpc_channel_bench.rs
+++ b/model_gateway/benches/grpc_channel_bench.rs
@@ -156,7 +156,6 @@ struct BenchResult {
     latency_p50_ms: f64,
     latency_p95_ms: f64,
     latency_p99_ms: f64,
-    latency_min_ms: f64,
     latency_max_ms: f64,
     errors: u64,
 }
@@ -289,8 +288,8 @@ async fn run_benchmark(
     let mut latency_values = latencies.lock().await;
     latency_values.sort_by(|a, b| a.partial_cmp(b).unwrap());
 
-    let (p50, p95, p99, min_lat, max_lat) = if latency_values.is_empty() {
-        (0.0, 0.0, 0.0, 0.0, 0.0)
+    let (p50, p95, p99, max_lat) = if latency_values.is_empty() {
+        (0.0, 0.0, 0.0, 0.0)
     } else {
         let len = latency_values.len();
         let p50_idx = (len as f64 * 0.50) as usize;
@@ -300,7 +299,6 @@ async fn run_benchmark(
             latency_values[p50_idx],
             latency_values[p95_idx.min(len - 1)],
             latency_values[p99_idx],
-            latency_values[0],
             latency_values[len - 1],
         )
     };
@@ -312,7 +310,6 @@ async fn run_benchmark(
         latency_p50_ms: p50,
         latency_p95_ms: p95,
         latency_p99_ms: p99,
-        latency_min_ms: min_lat,
         latency_max_ms: max_lat,
         errors,
     }


slin1237:
## Review: gRPC Channel Pool

### Summary

The PR adds a pool of 4 gRPC channels (TCP connections) per backend client to distribute concurrent RPCs and reduce HTTP/2 head-of-line blocking. While the theoretical motivation is sound, I don't think this optimization is warranted for our workload, and the implementation has several issues.

### Why this optimization is premature for LLM gateway traffic

The core premise — HTTP/2 HoL blocking on a single TCP connection — is a real problem for high-throughput, low-latency gRPC services. But our gRPC clients connect to **LLM inference backends** (sglang, vllm, trtllm), where:

1. **Inference is the bottleneck, not the channel.** A single generation request takes 1-30+ seconds. At realistic concurrency levels where the backend GPU isn't already saturated, a single HTTP/2 connection can easily multiplex those streams.

2. **Concurrency per backend is naturally bounded.** Even if the gateway handles thousands of requests, our load balancer distributes them across many worker pods. Each individual gRPC connection to a worker sees a manageable number of concurrent streams.

3. **We already set generous window sizes** (`initial_stream_window_size = 16MB`, `initial_connection_window_size = 32MB`) and `http2_adaptive_window(true)` — these already mitigate the flow-control contention the PR describes.

### Benchmark doesn't reflect production workload

The benchmark uses **100µs chunk delay** and measures throughput at **500 concurrent requests to a single endpoint**. In production:
- Inter-token latency is **10-100ms** (100-1000x slower)
- A single backend worker handles **tens** of concurrent requests, not hundreds
- The CPU time to serialize/deserialize protobuf messages is negligible vs inference time

The impressive 3.1x throughput improvement at 200 concurrency is measuring the mock server's ability to handle connections, not a bottleneck that exists in production. To justify this change, we'd need benchmarks showing channel saturation under **realistic** LLM inference latencies and concurrency levels.

### Pool size of 4 is unjustified

`DEFAULT_CHANNEL_POOL_SIZE = 4` appears in 3 files with no explanation for why 4. Not 2, not 8, not configurable. If this were needed, it should at minimum be:
- Configurable (env var or constructor parameter)
- Justified with analysis of when a single channel saturates

### Code duplication

The identical `ChannelPool` struct, `DEFAULT_CHANNEL_POOL_SIZE` constant, `next_client()` method, and pool initialization logic is copy-pasted across all 3 client files. If we were to pursue this, it should be a shared generic utility.

### Resource cost

4x TCP connections per worker means 4x the server-side connection load, 4x the keepalive traffic, and 4x the startup time (all `connect()` calls are sequential and eager). For workers that may be slow to start (loading model weights), this could be problematic.

### Recommendation

I'd suggest closing this PR unless there's evidence of channel saturation in production (e.g., gRPC deadline exceeded errors under load that correlate with a single-connection bottleneck, or `GOAWAY` frames from the server). The current single-channel setup with 16MB/32MB window sizes and adaptive windowing should be more than sufficient for LLM inference traffic.

If we do see channel pressure in the future, the right approach would likely be a shared, configurable `ChannelPool<T>` utility in `grpc_client/src/channel_pool.rs` rather than copy-pasted logic in each client.

PR #463, opened 2026-02-18
feat(tokenizer): Enrich tokenizer metadata with backend, vocab_size, and creation time

### LINKED ISSUE #28: Enrich TokenizerEntry with metadata
## Description
Add missing metadata to `TokenizerEntry` in the tokenizer registry.

## Current State
```rust
struct TokenizerEntry {
    id: String,
    name: String,
    source: String,
    tokenizer: Arc<dyn Tokenizer>,
}
```

## Proposed Changes
```rust
pub struct TokenizerEntry {
    pub id: String,
    pub name: String,
    pub source: String,
    pub tokenizer: Arc<dyn Tokenizer>,
    // New fields:
    pub vocab_size: usize,
    pub chat_template_path: Option<String>,
    pub backend: TokenizerBackend,
    pub created_at: DateTime<Utc>,
}

pub enum TokenizerBackend {
    HuggingFace,
    Tiktoken,
    Mock,
}
```

## Actions
- [ ] Add `TokenizerBackend` enum
- [ ] Add new fields to `TokenizerEntry`
- [ ] Update `load()` and `register()` to populate metadata
- [ ] Update API responses to include new metadata
- [ ] Add tests for metadata population

## Files
- `src/tokenizer/registry.rs`
- `src/tokenizer/mod.rs`
- `src/protocols/tokenize.rs`

ppraneth:
### Problem

Closes #28 

### Solution

This PR enriches the `TokenizerEntry` structure and the corresponding `TokenizerInfo` API response with comprehensive metadata.

* **Backend Identification**: Introduced a `TokenizerBackend` enum and a `backend()` method on the `Tokenizer` trait.
* **Automatic Population**: Updated `TokenizerRegistry::load` to automatically capture:
  * `vocab_size`
  * detected `backend`
  * `created_at` timestamp
* **API Exposure**: Updated the `TokenizerInfo` protocol to expose these new fields to API clients.

---

### **tokenizer crate**

* Added `TokenizerBackend` enum in `traits.rs`.
* Extended `TokenizerEntry` with:

  * `vocab_size`
  * `chat_template_path`
  * `backend`
  * `created_at`
* Implemented `backend()` for:

  * `HuggingFaceTokenizer`
  * `TiktokenTokenizer`
  * `MockTokenizer`
  * `CachedTokenizer`
* Updated `TokenizerRegistry::load` to populate metadata automatically.
* Re-exported `TokenizerBackend` in `lib.rs`.

### **protocols crate**

* Updated `TokenizerInfo` to include the new metadata fields.
* Enabled the `serde` feature for the `chrono` dependency.

### **model_gateway crate**

* Updated `LoadTokenizerStep` to pass chat template paths correctly.
* Updated handlers to map `TokenizerEntry` metadata to `TokenizerInfo`.

### **multimodal crate**

* Updated `TestTokenizer` to implement the new `backend()` trait method.

---

## Test Plan

Added new unit tests in `tokenizer/src/registry.rs` to verify metadata population and registry behavior.

### 1. Verify Metadata Population

```bash
cargo test --package llm-tokenizer --lib registry::tests::test_metadata_population
```

*Asserts that vocab_size is correctly captured, backend is identified as Mock, and timestamps are valid.*

### 2. Verify Registry Immutability

```bash
cargo test --package llm-tokenizer --lib registry::tests::test_load_ignores_updates
```

*Asserts that attempting to reload a tokenizer with different parameters does not corrupt the existing entry.*

### 3. General Regression

Ran full test suite for tokenizer crate:

```bash
cargo test --package llm-tokenizer
```

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Tokenizer endpoints return richer metadata: chat template path, backend type (HuggingFace, Tiktoken, Mock), and creation timestamp.
  * Tokenizers expose their backend type at runtime.
  * Tokenizer loading accepts an optional chat-template context for deduplication/association.
  * Vocabulary size is now pre-calculated and returned for improved performance.


slin1237:
Thanks for the work here — the metadata enrichment is useful and the implementation is clean. But I have a fundamental concern with the API design.

## `load()` shouldn't accept metadata the registry can derive itself

The registry's job is: take an ID, name, source, and a loader function → load the tokenizer → store it. After loading, it should derive all metadata from the tokenizer object itself.

Right now `chat_template_path` is threaded through `load()` purely to store it as metadata — it's not used during loading at all (the loader closure already handles the actual template). This is the wrong direction:

```rust
// Current: caller supplies metadata TO the registry
registry.load(&id, &name, &source, chat_template_path, || async { ... })

// What it should be: registry loads, derives everything from the tokenizer
registry.load(&id, &name, &source, || async { ... })
```

`vocab_size`, `backend`, `created_at` — you're already deriving these from the tokenizer in `TokenizerEntry::new()`. That's correct. But `chat_template_path` breaks the pattern by requiring the caller to pass it in.

## The fix

If `chat_template_path` is useful metadata, it should be **queryable from the tokenizer object itself** — either:

1. Add it to the `Tokenizer` trait: `fn chat_template_path(&self) -> Option<&str>`
2. Or store it internally in each tokenizer impl and expose it

Then the registry derives it the same way it derives `vocab_size` and `backend` — from the loaded tokenizer, not from the caller.

A tokenizer is a tokenizer. The registry shouldn't need to know about chat templates, vocab sizes, or backends as load parameters. It loads a tokenizer, then introspects it for metadata.

## Other notes

- `TokenizerBackend::Display` uses `{:?}` (PascalCase: `"HuggingFace"`) — most APIs use snake_case for enum values
- `TokenizerEntry::new()` should be `pub(crate)` not `pub`
- Missing trailing newline in `Cargo.toml`
- Consider `Serialize`/`Deserialize` on `TokenizerBackend` directly instead of converting to `String` in handlers

PR #465, opened 2026-02-18
fix: default tokenizer management

### LINKED ISSUE #29: Add unified resolve() method to TokenizerRegistry
## Description
Add a unified lookup method to `TokenizerRegistry` that handles all lookup patterns consistently.

## Problem
Currently there are multiple lookup patterns:
- HTTP handlers: try model → "default" → first available
- gRPC utils: only by model_id, no fallback
- Registry has: `get()`, `get_by_id()`, `get_by_name()` (inconsistent usage)

## Proposed Changes
```rust
impl TokenizerRegistry {
    /// Unified lookup: tries id → name → "default" → first available
    pub fn resolve(&self, key: Option<&str>) -> Option<Arc<dyn Tokenizer>> {
        match key {
            Some(k) => self.get_by_id(k)
                .or_else(|| self.get_by_name(k)),
            None => self.get_by_name("default")
                .or_else(|| self.first()),
        }
    }

    /// Get first available tokenizer
    pub fn first(&self) -> Option<Arc<dyn Tokenizer>>;

    /// Set default tokenizer by id
    pub fn set_default(&self, id: &str) -> Result<()>;
}
```

## Actions
- [ ] Add `resolve()` method to registry
- [ ] Add `first()` method to registry
- [x] Add `set_default()` method and internal default tracking
- [ ] Add tests for resolve fallback behavior

## Files
- `src/tokenizer/registry.rs`

ppraneth:
### Problem

closes #29 

## Changes

* **Added resolve()**: unified lookup method that tries:

  1. **ID/Name**: if a key is provided.
  2. **Explicit Default**: if a master default is set via **set_default()**.
  3. **"default"**: looks for a tokenizer named "default".
  4. **First Available**: falls back to any loaded tokenizer.

* **Added set_default()**: thread-safe method to pin a specific tokenizer as the default. Includes concurrency safety to prevent race conditions during setting.

* **Added first()**: helper to retrieve the first available tokenizer.

* **Added Tests**: `test_resolve_and_defaults` verifies the priority order and fallback mechanisms.

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added ability to set and manage a default tokenizer.
  * Enhanced tokenizer resolution with flexible fallback logic and improved retrieval mechanisms.

### DIFF
--- a/tokenizer/src/registry.rs
+++ b/tokenizer/src/registry.rs
@@ -16,7 +16,7 @@
 //! - `remove(name)`: Removes by name
 //! - `remove_by_id(id)`: Removes by ID
 
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 
 use dashmap::DashMap;
 use thiserror::Error;
@@ -88,6 +88,8 @@ pub struct TokenizerRegistry {
     tokenizers: DashMap<String, TokenizerEntry>,
     /// Secondary index: name -> UUID for lookup
     name_to_id: DashMap<String, String>,
+    /// Default tokenizer ID (optional)
+    default_id: RwLock<Option<String>>,
     /// Per-key locks to prevent duplicate loading
     loading_locks: DashMap<String, Arc<Mutex<()>>>,
 }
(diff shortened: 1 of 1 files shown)


slin1237:
Thanks for picking this up! The default tokenizer management is something we need, but I think we should take this opportunity to clean up the registry API rather than add more methods on top.

## Core concern: overlapping lookup methods

After this PR, there are 5+ ways to retrieve a tokenizer, with inconsistent behavior:

| Method | Lookup order | Returns |
|--------|-------------|---------|
| `get("foo")` | name → ID | `Option<Arc<dyn Tokenizer>>` |
| `get_by_name("foo")` | exact name | `Option<TokenizerEntry>` |
| `get_by_id("foo")` | exact ID | `Option<TokenizerEntry>` |
| `resolve(Some("foo"))` | **ID → name** | `Option<Arc<dyn Tokenizer>>` |
| `resolve(None)` | default → "default" → first | `Option<Arc<dyn Tokenizer>>` |
| `first()` | arbitrary DashMap order | `Option<Arc<dyn Tokenizer>>` |

`get()` and `resolve()` try name/ID in **opposite order** — that's a trap. Return types are inconsistent (`TokenizerEntry` vs bare `Arc<dyn Tokenizer>`). Callers won't know which method to use.

## What I'd like to see instead

**Consolidate the lookup API** — `resolve()` should be the single entry point, replacing `get()` and `first()`:

```rust
/// Single entry point for tokenizer lookup (replaces get, first)
pub fn resolve(&self, key: Option<&str>) -> Option<Arc<dyn Tokenizer>>

/// Same, but returns full entry with metadata
pub fn resolve_entry(&self, key: Option<&str>) -> Option<TokenizerEntry>
```

Then **deprecate or remove** `get()` and `first()` since `resolve()` subsumes both. Keep `get_by_id()` / `get_by_name()` for when callers need exact lookups.

## Bugs to fix

1. **`first()` is non-deterministic** — DashMap iteration order is hash-based, not insertion-ordered. The "first" tokenizer changes unpredictably. Bad for a default fallback.

2. **Stale default after remove** — `remove_by_id()`, `remove()`, and `clear()` don't reset `default_id`. After removing the default tokenizer, `resolve(None)` silently falls through instead of cleaning up.

3. **No `clear_default()`** — once set, the default can't be unset.

4. **`set_default` returns `Result<(), String>`** — should use a proper error type consistent with `LoadError`.

5. **Test doesn't verify priority** — it asserts `.is_some()` but never checks *which* tokenizer was returned. The whole point of `resolve()` is the fallback chain, and the test should verify the ordering.

## Suggested approach

1. Add `resolve()` / `resolve_entry()` as the unified lookup
2. Update `get()` to delegate to `resolve(Some(key))` (or deprecate it)
3. Remove standalone `first()`
4. Have `remove*()` and `clear()` clean up `default_id`
5. Add `clear_default()` method
6. Tests that verify the exact priority chain (explicit default > "default" name > insertion-ordered fallback)

Happy to discuss the design further before you rework this!

PR #481, opened 2026-02-20
feat(benchmark): add vLLM HTTP to nightly benchmark

paxiaatucsdedu:
### Problem
The nightly benchmark only ran vLLM over gRPC, without vLLM HTTP performance benchmark.

### Solution
Remove the `grpc_only: "true"` restriction on the vLLM variant in the nightly benchmark workflow.

## Changes

- ` .github/workflows/nightly-benchmark.yml`: Set `grpc_only: "false"` for the vLLM variant in all three jobs (single-worker, multi-worker, single-worker-h200). 
- `e2e_test/benchmarks/test_nightly_perf.py`: Updated module docstring to reflect that both protocols now run. 

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Tests**
  * Nightly benchmark workflow now runs both HTTP and gRPC for all runtime variants (vLLM no longer restricted to gRPC).
  * Added a pull-request trigger for nightly benchmark runs alongside scheduled runs.
  * Enabled test mode for nightly benchmarks to use constrained, faster single-concurrency test runs.

* **Bug Fixes**
  * Adjusted model launch logic so vLLM HTTP runs use vLLM-specific launch arguments while non-vLLM HTTP/GRPC retain prior argument handling.

### DIFF
--- a/e2e_test/infra/model_pool.py
+++ b/e2e_test/infra/model_pool.py
@@ -38,6 +38,14 @@ from .process_utils import detect_ib_device
 
 logger = logging.getLogger(__name__)
 
+# Shared vLLM defaults for both gRPC and HTTP modes.
+_VLLM_DEFAULT_ARGS = [
+    "--max-model-len",
+    "16384",
+    "--gpu-memory-utilization",
+    "0.9",
+]
+
 
 @dataclass(frozen=True)
 class WorkerIdentity:
@@ -524,11 +532,9 @@ class ModelPool:
 
         # vLLM HTTP: use dedicated vLLM OpenAI-compatible API server
         if mode == ConnectionMode.HTTP and is_vllm():
-            runtime_label = RUNTIME_LABELS.get(get_runtime(), "vLLM")
             logger.info(
-                "HTTP worker requested for %s: E2E_RUNTIME=vllm, routing to %s HTTP backend",
+                "HTTP worker requested for %s: routing to vLLM HTTP backend",
                 model_id,
-                runtime_label,
             )
             spec = get_model_spec(model_id)
             assert gpu_slot is not None
@@ -1191,11 +1197,8 @@ class ModelPool:
                 str(port),
                 "--tensor-parallel-size",
                 str(tp_size),
-                "--max-model-len",
-                "16384",
-                "--gpu-memory-utilization",
-                "0.9",
             ]
+            cmd.extend(_VLLM_DEFAULT_ARGS)
             extra = model_spec.get("vllm_args", [])
         elif runtime == "trtllm":
             cmd = [
@@ -1254,11 +1257,8 @@ class ModelPool:
             str(port),
             "--tensor-parallel-size",
             str(tp_size),
-            "--max-model-len",
-            "16384",
-            "--gpu-memory-utilization",
-            "0.9",
         ]
+        cmd.extend(_VLLM_DEFAULT_ARGS)
         extra = model_spec.get("vllm_args", [])
         if extra:
             cmd.extend(extra)
@@ -1489,10 +1489,7 @@ class ModelPool:
 
         cmd = self._build_vllm_http_cmd(model_path, DEFAULT_HOST, port, tp_size, model_spec)
 
-        if instance_key:
-            key = instance_key
-        else:
-            key = f"{model_id}:http"
+        key = instance_key or f"{model_id}:http"
         logger.info(
             "Launching vLLM HTTP worker %s on GPUs %s port %d",
             key,


.github/workflows/nightly-benchmark.yml
```diff
@@ -3,6 +3,9 @@ name: Nightly Benchmark
 on:
   schedule:
     - cron: '0 8 * * *'
+  pull_request:
```
slin1237, line 6: instead of branch matching
just run it during pull request
then revert this later

paxiaatucsdedu: Sure thanks

.github/workflows/nightly-benchmark.yml
```diff
@@ -3,6 +3,7 @@ name: Nightly Benchmark
...
```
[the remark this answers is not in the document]
CatherineSue: codex is right here. Why are we having the scope on pull_request?

e2e_test/infra/model_pool.py
```diff
@@ -531,47 +531,71 @@ def _launch_model(
...
+                "-m",
+                "vllm.entrypoints.openai.api_server",
+                "--model",
+                model_path,
+                "--host",
+                DEFAULT_HOST,
+                "--port",
+                str(port),
+                "--tensor-parallel-size",
+                str(tp_size),
```
CatherineSue: Why do we add these commands here? I think a better way is to consolidate this based on model instead of in model_pool.py

CatherineSue (changes requested):
The approach of inlining vLLM-specific command construction inside `_launch_model` doesn't follow the existing pattern. gRPC already handles runtime dispatch cleanly via `_build_grpc_cmd` / `_launch_grpc_worker` — HTTP should follow the same structure rather than scattering `if is_vllm()` branches in the shared launch path.

This also introduces several issues: `deep_health_check` will break (vLLM doesn't serve `/health_generate`), hardcoded defaults are duplicated across two locations, and SGLang-specific flags (embedding, PD disagg, worker_args) are silently skipped for vLLM HTTP without explicit intent.

e2e_test/infra/model_pool.py
paxiaatucsdedu: Thank you for pointing this out. These are the same defaults used by `_build_grpc_cmd`
 for vLLM gRPC (line 1194-1197 on main). Would you like me to consolidate this in this PR? Or maybe I can refactor this in a seperate PR?

paxiaatucsdedu:
> The approach of inlining vLLM-specific command construction inside `_launch_model` doesn't follow the existing pattern. gRPC already handles runtime dispatch cleanly via `_build_grpc_cmd` / `_launch_grpc_worker` — HTTP should follow the same structure rather than scattering `if is_vllm()` branches in the shared launch path.
> 
> This also introduces several issues: `deep_health_check` will break (vLLM doesn't serve `/health_generate`), hardcoded defaults are duplicated across two locations, and SGLang-specific flags (embedding, PD disagg, worker_args) are silently skipped for vLLM HTTP without explicit intent.

@CatherineSue Of course. Let me use the same structure for HTTP VLLM.

> codex is right here. Why are we having the scope on pull_request?

This line is purely for testing if the code works. I will remove it before merging.

e2e_test/infra/model_pool.py
CatherineSue: This should be consolidated in this PR, not deferred to a separate one. The correct approach is to follow the existing `_build_grpc_cmd` / `_launch_grpc_worker` pattern:

1. Add a `_build_vllm_http_cmd` static method (mirrors `_build_grpc_cmd` for the vllm branch but uses `vllm.entrypoints.openai.api_server`)
2. Add a `_launch_vllm_http_worker` method (mirrors `_launch_grpc_worker`)
3. At the top of `_launch_model`, dispatch `HTTP + is_vllm()` to the new launcher — same as how `GRPC + (is_vllm() or is_trtllm())` dispatches to `_launch_grpc_worker`

This keeps `_launch_model` as the SGLang-only path without scattering `if is_vllm()` branches in it. The hardcoded defaults (`--max-model-len`, `--gpu-memory-utilization`) should live in one place, not duplicated.

.github/workflows/nightly-benchmark.yml
CatherineSue: The latest push still has `pull_request:` in the workflow. Please remove it before pushing for review — reviewers should not need to mentally track which lines are "temporary". Same applies to `_TEST_MODE = True` in `test_nightly_perf.py`.

paxiaatucsdedu: I'll keep `pull_request:` and `_TEST_MODE = True` while iterating on the PR so I can validate changes on each push without waiting for the nightly schedule. I'll remove both in the final commit before merge. I will squash everything into a clean commit at that point.

.github/workflows/nightly-benchmark.yml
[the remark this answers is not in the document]
paxiaatucsdedu: I'll keep pull_request: and _TEST_MODE = True while iterating on the PR so I can validate changes on each push without waiting for the nightly schedule. I'll remove both in the final commit before merge, and will squash everything into a clean commit at that point.

Does this sound like good plan?

e2e_test/infra/model_pool.py
paxiaatucsdedu: I pushed the refactored version. vLLM HTTP now follows the same `_build_grpc_cmd` and `_launch_grpc_worker` pattern. Thanks

e2e_test/infra/model_pool.py
```diff
@@ -186,13 +186,18 @@ def _http_health_check(self, timeout: float = 5.0) -> bool:
...
+        For vLLM HTTP workers, falls back to /health (no /health_generate).
         For gRPC workers, falls back to standard health check.
         """
         if self.mode == ConnectionMode.GRPC:
             # For gRPC, use standard health check (no /health_generate equivalent)
             return self._grpc_health_check(timeout)
 
+        # vLLM HTTP does not support /health_generate, use /health instead
+        if self.mode == ConnectionMode.HTTP and is_vllm():
+            return self._http_health_check(timeout)
```
CatherineSue, line 207: Can we double check on vllm's health check endpoint? I remember they have something similar to /health_generate

e2e_test/infra/model_pool.py
```diff
@@ -517,6 +522,26 @@ def _launch_model(
                 assert instance is not None
                 return instance
 
+        # vLLM HTTP: use dedicated vLLM OpenAI-compatible API server
+        if mode == ConnectionMode.HTTP and is_vllm():
+            runtime_label = RUNTIME_LABELS.get(get_runtime(), "vLLM")
```
CatherineSue: Do we need `runtime = get_runtime()` here?

e2e_test/infra/model_pool.py
```diff
@@ -1197,6 +1222,48 @@ def _build_grpc_cmd(
...
+            "--model",
+            model_path,
+            "--host",
+            host,
+            "--port",
+            str(port),
+            "--tensor-parallel-size",
+            str(tp_size),
+            "--max-model-len",
+            "16384",
```
CatherineSue: This override is repeatitive everywhere. Is this specific to vLLM? If it does, how does it work with different models? Do any of them have specific overrides?

Is this required by both gRPC and http mode for vLLM? If it does, why don't we introduce a new function to consolidate it insead of having it spreaded everywhere?

e2e_test/infra/model_pool.py
```diff
@@ -1392,6 +1459,102 @@ def get_grpc_worker(
...
+        tp_size = model_spec.get("tp", 1)
+        port = gpu_slot.port
+        assert port is not None
+
+        env = os.environ.copy()
+        env["CUDA_VISIBLE_DEVICES"] = gpu_slot.cuda_visible_devices()
+
+        cmd = self._build_vllm_http_cmd(model_path, DEFAULT_HOST, port, tp_size, model_spec)
+
+        if instance_key:
```
CatherineSue: You can simply use a `key = instance_key or xxx` for this, instead of having a 4-line if-else.

e2e_test/infra/model_pool.py
paxiaatucsdedu: Sure. I searched `/health` key word in vllm code. Did not find something similar to `/health_generate`.

https://github.com/search?q=repo%3Avllm-project%2Fvllm+%22%2Fhealth%22&type=code&p=1

e2e_test/infra/model_pool.py
paxiaatucsdedu: Sure let me remove it.

e2e_test/infra/model_pool.py
paxiaatucsdedu: > Is this specific to vLLM? 

Yes.

> If it does, how does it work with different models? 

Models that need something different override it via `vllm_args` in `model_specs.py`.

> Do any of them have specific overrides?

Yes. The specific overrides are in `model_specs.py`.

> Is this required by both gRPC and http mode for vLLM? If it does, why don't we introduce a new function to consolidate it insead of having it spreaded everywhere?

Yes. The defaults were already hardcoded in `_build_grpc_cmd` on main. I will put them into a shared `_VLLM_DEFAULT_ARGS` constant in this PR.

e2e_test/infra/model_pool.py
paxiaatucsdedu: Of course. I see that `_launch_grpc_worker` uses the same 4-line if/else pattern in the main branch. Should I update that one as well in this PR?

PR #505, opened 2026-02-22
fix(data-connector): harden storage backends against data corruption and races

slin1237:
## Summary

Fixes 5 production-readiness issues in the `data_connector` storage backends (Postgres, Memory, Redis) identified during code review of #504 (commit `363bbe25`).

Refs: `363bbe25` (refactor(data-connector): comprehensive code quality improvements)

### `data_connector/src/postgres.rs`
- **Fix 1**: Replaced inline `parse_metadata` with delegation to the shared `crate::common::parse_conversation_metadata` helper, matching Oracle and Redis backends. The Postgres-specific version had subtly different whitespace/null handling.

### `data_connector/src/memory.rs`
- **Fix 2**: Consolidated `get_response_chain` from two separate lock acquisitions into a single read-lock scope. The original code collected chain IDs under one lock, dropped it, then re-acquired to fetch responses — allowing concurrent writers to delete entries between locks and silently lose chain data.

### `data_connector/src/redis.rs`
- **Fix 3**: Added `build_item_from_map` helper with proper error propagation for corrupted `content` (serde failures) and missing/invalid `created_at` timestamps. Previously, malformed data was silently replaced with `Value::Null` or `Utc::now()`, masking data corruption. Also hardened `build_response_from_map` with the same `created_at` error handling.
- **Fix 4**: Replaced sequential `HGET` + `DEL` in `delete_response` with a Redis pipeline for atomic read+delete. Eliminates a TOCTOU race where the `safety_identifier` could change between the read and the delete.
- **Fix 5**: Fixed cursor pagination in `list_items` to use inclusive score bounds with post-filtering by `item_id`, matching the composite `(added_at, item_id)` cursor semantics of Postgres and Oracle. The exclusive-bound approach silently skipped items sharing the same millisecond timestamp as the cursor.

## Why

The original refactor commit introduced several patterns that could cause silent data loss or corruption in production:
- Double lock in memory backend creates a TOCTOU window for chain data loss
- Silent fallback to `Utc::now()` and `Value::Null` in Redis masks corrupted records
- Non-atomic read+delete in Redis allows race conditions under concurrent access
- Inconsistent cursor pagination between Redis and SQL backends causes different query results depending on storage backend

## How

- Postgres: one-line delegation to existing shared helper
- Memory: restructured to single lock scope with in-place chain collection and post-reversal
- Redis items: extracted `build_item_from_map` that propagates `SerializationError` and `StorageError` instead of substituting defaults
- Redis responses: `build_response_from_map` now errors on missing/invalid `created_at`
- Redis delete: `redis::pipe()` batches `HGET` + `DEL` into one round-trip
- Redis pagination: inclusive `ZRANGEBYSCORE`/`ZREVRANGEBYSCORE` bounds with over-fetch (+32) and post-filter to skip the cursor item, then `take(limit)`

## Test plan

- All 106 existing unit tests pass: `cargo test -p data-connector`
- Clean build: `cargo check -p data-connector`
- No new warnings

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive README for the data-connector covering architecture, backends, configuration, usage, data models, and testing.

* **Improvements**
  * Clearer Postgres/Redis validation messages.
  * Centralized metadata parsing and more consistent storage handling.
  * Performance and atomicity improvements in memory and Redis flows.
  * Reduced public visibility for Redis storage types.

* **Bug Fixes**
  * More robust retrieval and atomic deletions to prevent inconsistent state.

* **Tests**
  * Broadly expanded unit test coverage across all storage backends.

* **Chores**
  * Removed a workspace flag from a dependency declaration.

### DIFF
--- a/data_connector/src/redis.rs
+++ b/data_connector/src/redis.rs
@@ -259,7 +259,15 @@ impl RedisConversationItemStorage {
                 .unwrap_or_else(|| fallback_id.to_string()),
         );
         let response_id = map.get("response_id").cloned();
-        let item_type = map.get("item_type").cloned().unwrap_or_default();
+        let item_type = map
+            .get("item_type")
+            .filter(|s| !s.is_empty())
+            .cloned()
+            .ok_or_else(|| {
+                ConversationItemStorageError::StorageError(format!(
+                    "item {fallback_id} missing item_type"
+                ))
+            })?;
         let role = map.get("role").cloned();
         let status = map.get("status").cloned();
 


data_connector/src/redis.rs
```diff
@@ -360,33 +402,68 @@ impl ConversationItemStorage for RedisConversationItemStorage {
...
+                    // the boundary of the result set (first items for ASC, last
+                    // for DESC). We conservatively filter by ID comparison for
+                    // all items — the string comparison is cheap.
+                    id != c_id
+                })
+                .take(params.limit)
+                .collect()
+        } else {
+            item_ids.into_iter().take(params.limit).collect()
         };
```
slin1237, line 467: The `skip_while` approach is correct here. Redis guarantees that members with the same score are ordered lexicographically (from the docs: "When two elements have the same score, they are ordered lexicographically"). This means:

- **ASC** (`ZRANGEBYSCORE`): same-score members come in lex order → `skip_while(|id| id \!= c_id)` skips all lex-predecessors, then `skip(1)` drops the cursor itself
- **DESC** (`ZREVRANGEBYSCORE`): same-score members come in reverse-lex order → same logic applies correctly

The WITHSCORES approach would also work but adds unnecessary complexity (raw `redis::cmd` instead of typed `AsyncCommands` methods) for no correctness gain given the lex-ordering guarantee.

The `+32` over-fetch heuristic is acknowledged — in practice, >32 items sharing the exact same millisecond timestamp would be extremely unusual for conversation items, but if this becomes a concern we can switch to ZRANGEBYLEX within the score bucket.

PR #585, opened 2026-03-03
feat(epd): add gRPC encode-prefill-decode routing mode [WIP]

chenzongyao200127:
### Backgroud

SGLang already supports EPD (Encode-Prefill-Decode) mode to split work across dedicated encode, prefill, and decode workers.
Our gateway did not have a complete gRPC EPD routing path, so it could not properly route requests for this deployment model.

### Solution

This PR adds a dedicated gRPC EPD router path and the supporting worker/discovery logic.
It introduces encode-stage routing, updates worker detection/health handling for EPD workers, and wires EPD mode through router selection and request flow so EPD clusters can run end-to-end.

## Changes

- Added gRPC EPD router implementation and integrated it into router factory/pipeline.
- Added encode worker support across worker selection and routing flow.
- Updated worker connection/backend detection for EPD-related gRPC behavior.
- Extended service discovery config/selectors to support encode pods.
- Updated gRPC client protocol usage (including health proto support) for new routing needs.
- Updated related config, metrics, and router manager wiring for EPD mode.

<details>
<summary>Checklist</summary>

</details>


grpc_client/src/health.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// gRPC health client (grpc.health.v1.Health).
+#[derive(Clone)]
+pub struct HealthClient {
+    client: proto::health_client::HealthClient<Channel>,
+}
+
+impl HealthClient {
+    pub async fn connect(endpoint: &str) -> Result<Self, Box<dyn std::error::Error + Send + Sync>> {
+        let http_endpoint = if let Some(addr) = endpoint.strip_prefix("grpc://") {
```
CatherineSue, line 25: What is this change for?

model_gateway/src/routers/grpc/utils.rs
```diff
@@ -414,6 +414,202 @@ pub(crate) fn filter_chat_request_by_tool_choice(
...
+                urls.push(url.clone());
+            }
+        }
+        _ => {}
+    }
+}
+
+pub(crate) fn build_multimodal_inputs_from_generate(
+    request: &GenerateRequest,
+) -> Option<smg_grpc_client::sglang_proto::MultimodalInputs> {
```
CatherineSue, line 532: This doesn't work with our current mm pipeline. The function should not be put here. And having parameter solely relates to sglang_proto also brings up concerns for compatibility.

CatherineSue (commented):
Why do we remove sglang_encoder.proto here? The overall design seems to be diverged from the previous PR. Is it possible to be clear about your current design?

Threading a `epd_mode` through so many files is too invasive IMO.

chenzongyao200127:
> Why do we remove sglang_encoder.proto here? The overall design seems to be diverged from the previous PR. Is it possible to be clear about your current design?
> 
> Threading a `epd_mode` through so many files is too invasive IMO.

Thank you for taking the time to review this and for your valuable comments.

"Regarding the removal of sglang_encoder.proto": Apologies for the confusion—that was an accidental git mistake. I will fix it in the next update.

Regarding the design differences:
"Could you please clarify which specific previous PR you are referring to? Are you mentioning this [PR](https://github.com/sgl-project/sglang/pull/17550/changes/8257f4d76e9bd39fdf10e01c0df2e3b8f042c630..2b9d76c6af407ee8a19ee5359ea07718f8437611)? "
I am happy to create a comprehensive design issue to elaborate on the current EPD implementation in sglang, as well as the required modifications for the router under the current EPD mode. I can prepare this within 2 days.

Regarding epd_mode being too invasive:
I agree with your concern that propagating `epd_mode` through so many files is too invasive. Since EPD is still somewhat experimental, it perhaps shouldn't be treated on par with PD at this stage. I will address the background context, the necessary router modifications, and how we can achieve this with a more minimal change in the design issue mentioned above.

I will ping you as soon as the design issue is ready. Thanks again!

CatherineSue:
> "Regarding the removal of sglang_encoder.proto": Apologies for the confusion—that was an accidental git mistake. I will fix it in the next update.

Ahh I see. I think we may need to update the `__init__.py` in `proto/python` as well if we intend to remove the proto.

> "Could you please clarify which specific previous PR you are referring to? Are you mentioning this PR?"

I had the impression that we will be using `sglang_encoder.proto` to talk to another sglang gRPC server. I was mainly referring to the previous proto PR. Sorry about the confusion.

Regarding the design, is it possible to point me the relevant EPD code in sglang side? I read about the changes in `sglang/entrypoints/grpc_server.py` already. I might have a better idea about what we are trying to send here if I get more context. And be able to help you think of a design to move forward. I think some code links should be enough, no need for a thorough design doc unless we have too many questions, so as to save some of your time.

PR #586, opened 2026-03-03
feat(lora): add LoRA adapter lifecycle management

ConnorLi96:
Load local LoRA adapters into SGLang/vLLM on first use, cache in memory, and rewrite inference requests to use the engine-side adapter name.

- core/lora/uri.rs: classify lora_path; error on remote schemes (s3://, etc.)
- core/lora/engine_client.rs: backend-aware load/unload HTTP calls (SGLang + vLLM)
- core/lora/middleware.rs: DashMap cache + per-URI Mutex for concurrent load dedup
- config/types.rs: LoraConfig + LoraBackend
- main.rs: --enable-lora flag; backend auto-derived from --backend
- router.rs: call lora.resolve() in send_typed_request; UnsupportedUri → 400, Engine → 500
- app_context.rs: wire LoraMiddleware into AppContext builder

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added per-worker LoRA adapter state management with automatic caching and lifecycle control.
  * Added load monitor interval configuration option for workers.

* **Documentation**
  * Updated `lora_path` field documentation to clarify it accepts adapter serving names or filesystem paths.
  * Deprecated `lora_id` field; migrate to `lora_path` instead.


slin1237 (changes requested):
Hey, thanks for taking this on — LoRA lifecycle management is an important feature and I can see you put a lot of thought into the architecture. I have some feedback on the overall design direction that I think will simplify the implementation significantly.

### 1. Don't change the request spec (`chat.rs`, `completion.rs`, `generate.rs`)

Changing `lora_path` from `Option<String>` to `Option<StorageSpec>` is a breaking API change for existing clients. Even with the custom deserializer accepting both forms, the serialization output changes — anything downstream consuming the raw JSON will break.

The request types are part of our wire protocol and should stay stable. Keep `lora_path: Option<String>` as-is in all three request types.

### 2. LoRA belongs on the worker spec, not the request path

The core design question is: **when** do adapters get loaded?

The current PR treats this as a request-time concern — every inference request with `lora_path` hits a DashMap lookup, potentially triggers a load call to the engine, and rewrites the JSON. This is a middleware/caching pattern that seems simple but actually hides a much harder problem: if you're dynamically loading adapters on inference requests, you need a proper GPU memory scheduler (eviction policies, memory budgets, concurrency coordination across DP ranks, etc.). That's a large scope and should be a separate effort.

**For v1, the simpler and correct approach is:**
- Add a `loras: Vec<LoraSpec>` field to `WorkerSpec`
- During the worker startup workflow, iterate the list and call the engine's load API for each adapter
- On worker shutdown, unload them
- Inference requests just pass `lora_path` through as a string — the engine already knows the adapter by name

This keeps the hot path clean and avoids the scheduling complexity entirely.

### 3. `LoraConfig` / `LoraBackend` is redundant

The worker already knows what engine it's running via `WorkerSpec.runtime_type` (`Sglang`, `Vllm`, `TrtLlm`, `External`). There's no need for a separate `LoraBackend` enum that duplicates this information. When you need to call the load/unload endpoint, just match on `runtime_type` to pick the right URL path.

### 4. No need for a separate engine client

`LoraEngineClient` adds a new struct with its own HTTP client. The load/unload calls are just HTTP POSTs to well-known endpoints — this can be a few lines in the worker startup workflow step using the existing HTTP client. No need for a dedicated abstraction.

### 5. Storage + auth design

The `auth: Option<serde_json::Value>` approach is a good instinct (keeping it flexible), but different storage vendors need different auth shapes:
- **S3**: AWS credentials or IAM role
- **OCI Object Storage**: user principal, instance principal, resource principal
- **GCS**: service account or workload identity

Consider making this a tagged enum so we get type safety and clear documentation of what each vendor needs:

```rust
#[serde(tag = "type")]
enum StorageAuth {
    Aws { region: String, ... },
    Oci { auth_type: OciAuthType, ... },
    Gcs { ... },
}
```

This doesn't have to be fully fleshed out in v1, but the shape should be extensible.

### 6. What about remote storage?

The current PR rejects all remote schemes (`s3://`, `gs://`, etc.) with a 400. That's fine for v1, but the worker workflow approach makes remote support straightforward to add later: check if the path is remote → download to a local temp dir using vendor-appropriate auth → then call the engine's load API with the local path.

### Suggested approach for v1

1. **Don't touch** `chat.rs`, `completion.rs`, `generate.rs`
2. **Add `loras: Vec<LoraSpec>` to `WorkerSpec`** — each `LoraSpec` has `id: String`, `path: String`, and optional `auth`
3. **Add a workflow step** in the worker startup flow that loads each adapter via the engine's HTTP API (derive endpoint from `runtime_type`)
4. **Add a shutdown step** that unloads adapters
5. **Remove** `LoraConfig`, `LoraBackend`, `LoraEngineClient`, `WorkerLoraState`, `StorageSpec` from protocols, and the `prepare_request` override

This cuts the PR from ~800 lines to probably ~150-200 and gives us a solid foundation to build on (dynamic scheduling, remote download, etc.) in follow-up PRs.

Happy to pair on the redesign if that would be helpful!

slin1237 (approved):

slin1237 (approved):

CatherineSue:
Removed `stale` label. Added by bot accidentally.

PR #627, opened 2026-03-04
refactor(oracle): use existing CONVERSATIONS table in Flyway test schema

key4ng:
### Problem
The Flyway test migrations create a separate CONVERSATIONS_V2 table, but in production the existing CONVERSATIONS table (with `CONVERSATION_ID` as the primary key) is already available and should be reused directly.

### Solution
Move the CONVERSATIONS table DDL into V1 alongside RESPONSES to reflect the actual production baseline, and update the schema-config to remap SMG's logical `id` column to the physical `CONVERSATION_ID` column.

## Changes
- `scripts/oracle_flyway/schema-config.yaml`: Point conversations at `CONVERSATIONS` table instead of `CONVERSATIONS_V2`, add `id: CONVERSATION_ID` column mapping
- `scripts/oracle_flyway/sql/V1__Create_responses_and_conversations_table.sql`: Renamed from `V1__Create_responses_table.sql`, added full CONVERSATIONS table DDL matching the production schema
- `scripts/oracle_flyway/sql/V2__Create_conversation_items_and_alter_responses.sql`: Renamed from `V2__Create_v2_conversations_and_alter_responses.sql`, removed CONVERSATIONS_V2 DDL (now in V1)

## Summary by CodeRabbit

* **Database Updates**
  * Consolidated conversation schema and established baseline tables for conversations and responses
  * Added EXPIRES_AT column with a default timestamp for automatic expiration
  * Added SAFETY_IDENTIFIER to response records for tracking
  * Removed legacy V2 conversation definition, refreshed table layout, indexes, and constraints for reliability and query performance
  * Introduced conversation item tables and linking structures

* **Chores**
  * CI workflow updated to trigger on database migration path changes

### DIFF
--- a/.github/workflows/pr-test-rust.yml
+++ b/.github/workflows/pr-test-rust.yml
@@ -321,7 +321,6 @@ jobs:
               - 'e2e_test/router/test_pd_mmlu.py'
               - 'scripts/ci_install_vllm.sh'
               - 'scripts/ci_install_trtllm.sh'
-              - 'scripts/ci_install_sglang.sh'
             agentic:
               - 'mcp/**'
               - 'data_connector/**'


.github/workflows/pr-test-rust.yml
```diff
@@ -321,12 +321,14 @@ jobs:
               - 'e2e_test/router/test_pd_mmlu.py'
               - 'scripts/ci_install_vllm.sh'
               - 'scripts/ci_install_trtllm.sh'
+              - 'scripts/ci_install_sglang.sh'
```
CatherineSue: We don't need this, it is in common already.

PR #648, opened 2026-03-05
fix(grpc): increase SGLang keep-alive timeout to prevent connection drops during long generation

vschandramourya:
### Problem

Long-running SGLang generation requests (30K+ output tokens, ~3-16 minutes) are silently killed mid-stream by the SMG's HTTP/2 keep-alive mechanism, then enter an expensive retry loop where each attempt regenerates all tokens before failing identically.

The SGLang proto has no application-level heartbeat — `GenerateResponse` only carries `chunk | complete | error`, so liveness depends entirely on HTTP/2 PING/PONG at the transport layer.

When the connection drops, `collect_stream_responses` (`grpc/utils.rs:681`) returns `worker_stream_failed` (500), which propagates up as `"Stage ResponseProcessing failed"` (`pipeline.rs:307`). Since 500 is retryable, `RetryExecutor` re-enters the pipeline, taking up another 3-16 minutes of GPU time per attempt.

### Solution

Relax the SGLang gRPC channel keep-alive settings:

| Setting | Before | After | Rationale |
|---------|--------|-------|-----------|
| `http2_keep_alive_interval` | 30s | **60s** | Matches existing `tcp_keepalive` (60s); reduces PING frequency to stay within server-side enforcement limits |
| `keep_alive_timeout` | 10s | **120s** | Gives SGLang's event loop adequate time to respond during heavy batch processing; still well under the 30-min `request_timeout_secs` |

vLLM and TRT-LLM clients are unchanged as this targets only the SGLang path where the issue is observed.

## Changes

- `grpc_client/src/sglang_scheduler.rs`: `http2_keep_alive_interval` 30s → 60s, `keep_alive_timeout` 10s → 120s

## Test Plan

- Deploy with a model that generates to max tokens (~32K) and verify requests complete without `ResponseProcessing` failures
- Confirm `worker_stream_failed` errors are eliminated for long generation requests
- Verify short-lived requests are unaffected

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Performance**
  * Adjusted gRPC client keep-alive settings for improved connection management.

### DIFF
--- a/grpc_client/src/sglang_scheduler.rs
+++ b/grpc_client/src/sglang_scheduler.rs
@@ -148,8 +148,8 @@ impl SglangSchedulerClient {
         };
 
         let channel = Channel::from_shared(http_endpoint)?
-            .http2_keep_alive_interval(Duration::from_secs(30))
-            .keep_alive_timeout(Duration::from_secs(10))
+            .http2_keep_alive_interval(Duration::from_secs(60))
+            .keep_alive_timeout(Duration::from_secs(120))
             .keep_alive_while_idle(true)
             .tcp_keepalive(Some(Duration::from_secs(60)))
             .tcp_nodelay(true)


slin1237 (changes requested):
## Review: This fix addresses a symptom, not the root cause

I investigated the full gRPC keepalive interaction between the Rust client and the SGLang Python server. The proposed timeout bump will likely mask failures rather than fix them, and could make things worse.

### The actual root cause

The SGLang gRPC server (`sglang/srt/entrypoints/grpc_server.py:934`) configures **zero keepalive options**:

```python
server = grpc.aio.server(
    futures.ThreadPoolExecutor(max_workers=10),
    options=[
        ("grpc.max_send_message_length", 1024 * 1024 * 256),
        ("grpc.max_receive_message_length", 1024 * 1024 * 256),
    ],
)
```

This means the server uses **gRPC C-core defaults**, critically:
- `grpc.http2.min_recv_ping_interval_without_data_ms` = **300,000ms (5 minutes)**
- `grpc.http2.max_ping_strikes` = **2** (GOAWAY after 2 violations)

Meanwhile the client sends HTTP/2 PINGs every **30s**. During long prefill or any gap between DATA frames, the server sees pings arriving 10x more frequently than its minimum interval allows. After 2 violations (~60s), the server sends **GOAWAY and kills the connection**.

This explains:
1. Why it only hits long-running requests (more time in prefill / between chunks = more "data-free" ping cycles)
2. Why it's "silent" (GOAWAY is transport-level, not application-level)
3. Why vLLM/TRT-LLM are unaffected — their server (`grpc_servicer/vllm/server.py`) has different traffic patterns or the issue hasn't been observed yet (they share the same client keepalive settings)

### Why this PR's fix is wrong

| Problem | Detail |
|---------|--------|
| **Doesn't fix the root cause** | Changing interval from 30s→60s still violates the server's 300s minimum. Connections will still be killed, just slightly later. |
| **120s keep_alive_timeout hides dead connections** | If a connection genuinely dies, you wait **2 full minutes** before detecting it. The current 10s is aggressive but reasonable. |
| **Makes retry storms worse** | The PR description says retries burn 3-16 min of GPU time per attempt. A 120s detection delay means even more wasted GPU time before failover. |
| **Only changes one of three identical clients** | All three clients (SGLang, vLLM, TRT-LLM) have the exact same `30s/10s` config. If this were the right fix, all three should change. |

### The correct fix

**Option A — Fix the server side (preferred, upstream SGLang PR):**
```python
server = grpc.aio.server(
    futures.ThreadPoolExecutor(max_workers=10),
    options=[
        ("grpc.max_send_message_length", 1024 * 1024 * 256),
        ("grpc.max_receive_message_length", 1024 * 1024 * 256),
        # Allow client keepalive pings
        ("grpc.http2.min_recv_ping_interval_without_data_ms", 10000),
        ("grpc.keepalive_permit_without_calls", True),
    ],
)
```

This tells the server "pings every 10s+ are fine", which is compatible with the client's 30s interval. No client-side changes needed.

**Option B — If we can't change the server, fix the client properly:**
The client should either:
1. Stop sending HTTP/2 PINGs when there's no data flowing (not directly configurable in tonic)
2. Or increase `http2_keep_alive_interval` to **>300s** to stay within the server's default enforcement — but this defeats the purpose of keepalive

CatherineSue:
Do you have the actual logs of SGLang and smg when this issue happens? Because if there are constant data chunks flow, the settings here should not be an issue. The gRPC channel is already busy with streaming chunks.

The PING is per connection, not per request. This situation can only happen under the following cases:

```
1. Connection is idle — no active RPCs, or no DATA frames flowing                                                                                                                     
2. Client sends PING (every 30s because of keep_alive_while_idle(true))
3. Server checks: "last PING was 30s ago, my minimum is 300s, that's too frequent"
4. After a few violations, server sends GOAWAY (ENHANCE_YOUR_CALM) and kills the connection
```

This means under this situation, the whole connection is idle. There is literally no chunk. Unless you are using `stream=False`, it is hard for me to believe the engine would send no data chunks even with preemption and kv cache swap, and those are used for the purpose of serving some of the requests active. Not all of them idle.

Another thing is smg should be sending health checks to the server every 60s. It shares the same connection. Every 60s there is a full RPC trip. 

That means the "without data" window is at most ~60s between health checks. The default `min_recv_ping_interval_without_data` is 300s (5 min). The client PINGs every 30s, so between two health checks there's only one PING "without data" — the server sees PINGs 60s apart at worst (data resets the counter), which is well under the 300s enforcement threshold.

I agree the above server fix is programatic for a real issue. I am not sure if it would fix the issue you describe. We need more information.

vschandramourya:
@CatherineSue - showing the reproduced issue with GLM-5-FP4, running GPQA (32 workers, temp=1, max_new_tokens=30000, stream=False). Full logs attached.

Catherine, your point about DATA frames resetting the ping counter is correct for streaming, but this workload is non-streaming. With stream=False, the SGLang gRPC server sends exactly one GenerateComplete message at the end of generation. For the entire generation window (which can be minutes at 30k tokens), there are zero DATA frames on the wire. The only thing the transport sees is PINGs into what looks like a dead connection.

The logs confirm this. There's a clean connection drop where a health check and all 10 active requests are cancelled by client simultaneously, that's the gRPC channel dying, not individual request failures. After that, 55 seconds of silence, then the 500s start degrading... as retried requests hit the broken response state. 100+ total ResponseProcessing failures.

@slin1237 , just checked your GOAWAY analysis, which is correct for this non-streaming case. The server's default min_recv_ping_interval_without_data_ms (300s) sees the client's 30s PINGs as violations during these long data-less generation windows. I guess the server fix (adding min_recv_ping_interval_without_data_ms: 10000 to grpc.aio.server) is the right long-term answer. The client-side timeout bump in this PR is a short-term mitigation while that land

Three questions for you both:

- Is this strictly a non-streaming problem? If streaming requests always have DATA frames flowing, does that mean this issue can never surface with stream=True and the current keepalive settings are fine for that path?
- Should we treat this PR as a temporary fix in local and land the proper server-side min_recv_ping_interval_without_data_ms change upstream in SGLang instead?
- On the ResponseProcessing 500s, part of this comes from the reasoning parser's BufferOverflow being treated as fatal. Could we downgrade that to non-fatal and return a warning to the client with a partial/best-effort response instead of rejecting the entire request?

I agree this PR is a rough patch to the existing issue, not a root cause fix. Glad you both are pointing out new angles, the server-side keepalive config and the non-streaming DATA frame gap weren't on my radar before this. Learning more! Thank you.

## Relevant

>  Logs
> 
> 1) Engine idle before the burst
> 
> [2026-03-05 23:35:45] Receive health check request: HEALTH_CHECK_1772753745.855462
> [2026-03-05 23:35:46 TP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0, input throughput (token/s): 1.07, cuda graph: False
> 
> 2) gRPC channel death
> 
> [2026-03-05 23:36:45] Receive health check request: HEALTH_CHECK_1772753805.8569539
> [2026-03-05 23:36:46 TP0] Decode batch, #running-req: 10, #token: 15040, token usage: 0.01, accept len: 2.96, accept rate: 0.74, cuda graph: True, gen throughput (token/s): 1049.57, #queue-req: 0
> [2026-03-05 23:36:46] Request HEALTH_CHECK_1772753805.8569539 cancelled by client
> 
> [2026-03-05 23:36:47 TP0] Decode batch, #running-req: 10, #token: 16256, token usage: 0.01, accept len: 2.86, accept rate: 0.71, cuda graph: True, gen throughput (token/s): 1016.69, #queue-req: 0
> [2026-03-05 23:36:49 TP0] Decode batch, #running-req: 10, #token: 18432, token usage: 0.01, accept len: 2.92, accept rate: 0.73, cuda graph: True, gen throughput (token/s): 1034.95, #queue-req: 0
> 
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-6b6b-7561-953c-fafd5c363a18 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3b3d-7ad0-bba1-84863670c807 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3ba0-7cc1-9065-c6f168d379b6 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3c04-7f03-95a7-83a048ca0357 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3c69-7611-b3bf-1d50864c69cf cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3cce-7d00-a4c4-bd112db084a4 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3d32-7020-bd15-ec752a87ba48 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3d97-7593-80b2-9b2b69522906 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3dfd-7af3-9fc1-30c3849d9389 cancelled by client
> [2026-03-05 23:36:50] Request chatcmpl-019cc05c-3e68-7d40-87c8-b2ce905fec2a cancelled by client
> 
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3c04-7f03-95a7-83a048ca0357
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3c69-7611-b3bf-1d50864c69cf
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3ba0-7cc1-9065-c6f168d379b6
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3cce-7d00-a4c4-bd112db084a4
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3d32-7020-bd15-ec752a87ba48
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3d97-7593-80b2-9b2b69522906
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3dfd-7af3-9fc1-30c3849d9389
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-6b6b-7561-953c-fafd5c363a18
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3e68-7d40-87c8-b2ce905fec2a
> [2026-03-05 23:36:50] Receive abort request: chatcmpl-019cc05c-3b3d-7ad0-bba1-84863670c807
> 
> 
> 3) ~55s silence 
> 
> [2026-03-05 23:37:45] Receive health check request: HEALTH_CHECK_1772753865.8576007
> [2026-03-05 23:37:46 TP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0, input throughput (token/s): 44.45, cuda graph: False
> 
> [2026-03-05 23:38:03] Request chatcmpl-019cc04e-f3cc-77e0-9995-e563c7c0aaca cancelled by client
> [2026-03-05 23:38:03] Request chatcmpl-019cc04e-f3cb-7822-8791-469452635142 cancelled by client
> 
> 
> 4) 500 cascade
> 
> [2026-03-05 23:42:09 TP0] Decode batch, #running-req: 13, #token: 167744, token usage: 0.12, accept len: 2.73, accept rate: 0.68, cuda graph: True, gen throughput (token/s): 1112.99, #queue-req: 0
> [2026-03-05 23:42:11 TP0] Decode batch, #running-req: 13, #token: 169280, token usage: 0.12, accept len: 2.77, accept rate: 0.69, cuda graph: True, gen throughput (token/s): 1127.67, #queue-req: 0
> 
> [2026-03-05 23:42:12] ERROR http_request{method=POST uri=/v1/chat/completions version=HTTP/1.1 module="smg" request_id="chatcmpl-D6noRa7fU1KvI6OlKwgtgy5b"}: smg::routers::grpc::pipeline: model_gateway/src/routers/grpc/pipeline.rs:307: Stage ResponseProcessing failed with status 500 Internal Server Error
> 
> [2026-03-05 23:42:12] Receive generation request: chatcmpl-019cc061-5c7e-7043-8556-608a52b962dc
> [2026-03-05 23:42:12 TP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 384, token usage: 0.11, #running-req: 12, #queue-req: 0, input throughput (token/s): 9.83, cuda graph: False
> [2026-03-05 23:42:12 TP0] Decode batch, #running-req: 13, #token: 151104, token usage: 0.11, accept len: 2.83, accept rate: 0.71, cuda graph: True, gen throughput (token/s): 1032.58, #queue-req: 0

CatherineSue:
@vschandramourya Thanks for the logs. It will be better to view if it is more formatted.

Feel free to call me Chang, as I have introduced myself in the first meet up meeting.

The non-streaming case (stream=False) is a valid scenario — with no DATA frames during generation, the PING enforcement could trigger. But looking at the logs more carefully, the timing still doesn't quite add up:
                                                                                                                                                                                    
The timeline:

```
- 23:35:45 — health check (DATA frames flow, should reset PING strikes)                                                                                                               
- 23:36:15 — PING #1 (30s later). First PING after data → "with data", no strike
- 23:36:45 — PING #2 (30s later). "Without data" → strike 1
- 23:36:45 — health check also arrives (DATA frames, should reset strikes)
- 23:36:46 — health check cancelled, all 10 requests cancelled
```

The connection died after only 1 strike. GOAWAY needs **2 strikes**. So either:

1. There were accumulated strikes from before the burst (the engine was idle, keep_alive_while_idle(true) was sending PINGs into a truly empty connection)
2. The PING and health check raced at 23:36:45 and the PING arrived first, triggering strike 2 from a previous cycle
3. It's not GOAWAY at all — it's something else

I don't think this PR works for a mitigation. The setting is non-standrad for most of the gRPC service. The correct fix should be able to land soon as well. 

Would have better info if we have smg logs as well

CatherineSue:
> On the ResponseProcessing 500s, part of this comes from the reasoning parser's BufferOverflow being treated as fatal. Could we downgrade that to non-fatal and return a warning to the client with a partial/best-effort response instead of rejecting the entire request?

This is a separate issue. I am confused with the overall situation right now. How do we combine the ReasoningParser buffer overflow with GOAWAY? Please open a separate issue for this. Please include detailed original logs of smg and the inference backend with no trimming.

vschandramourya:
Sure, that makes sense Chang... this PR isnt the right mitigation, so Ill close it and defer to the server-side min_recv_ping_interval_without_data_ms fix once that lands properly.

On the BufferOverflow point, to avoid confusion... I think that should be tracked separately, I was conflating the two issues. I’ll open a dedicated issue for that with the full unedited logs from both SMG and the inference backend.

PR #653, opened 2026-03-06
fix(reasoning-parser): increase buffer limit to 4MB and make parse errors non-fatal

CatherineSue:
### Problem

The reasoning parser's buffer limit was set to 64KB, which is too small for models that produce long reasoning outputs (e.g., DeepSeek-R1 with 30K+ tokens). Additionally, parse errors were fatal — a buffer overflow would kill the entire request instead of gracefully falling back to raw text.

### Solution

- Increase `max_buffer_size` from 64KB to 4MB across all reasoning parsers. This is a per-request allocation, so 4MB is safe even under high concurrency.
- Make reasoning parse errors non-fatal: log a warning and fall back to returning raw text instead of returning an error to the caller.

## Changes

- Updated `max_buffer_size` from `65536` to `4 * 1024 * 1024` in all parser configs (base, DeepSeek-R1, Qwen3, Kimi, Cohere, GLM45, MiniMax, NanoV3, Step3) and the `ParserConfig` default.
- Changed `processor.rs` to `warn!` on reasoning parse errors and skip parsing, instead of returning `Err`.

## Test Plan

- Existing unit tests pass with the updated buffer size.
- Models producing >64KB of reasoning content (e.g., DeepSeek-R1 with extended thinking) will no longer hit buffer overflow errors.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reasoning parsing now logs a warning and skips failed parses instead of aborting, improving resilience.

* **Improvements**
  * Default parser buffer capacity increased from 64KB to 4MB across reasoning parsers to better handle larger reasoning outputs.

### DIFF
--- a/reasoning_parser/src/factory.rs
+++ b/reasoning_parser/src/factory.rs
@@ -11,7 +11,7 @@ use crate::{
         BaseReasoningParser, CohereCmdParser, DeepSeekR1Parser, Glm45Parser, KimiParser,
         MiniMaxParser, NanoV3Parser, Qwen3Parser, QwenThinkingParser, Step3Parser,
     },
-    traits::{ParserConfig, ReasoningParser},
+    traits::{ParserConfig, ReasoningParser, DEFAULT_MAX_BUFFER_SIZE},
 };
 
 /// Type alias for pooled parser instances.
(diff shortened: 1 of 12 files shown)


CatherineSue:
@vschandramourya

vschandramourya:
Thank you Chang for taking this up and making the reasoning parser failure non-fatal! Definitely helps with long responses.

PR #664, opened 2026-03-07
feat(grpc): add /v1/completions endpoint support with streaming

vschandramourya:
### Problem

The gRPC router returned `501 Not Implemented` for `/v1/completions`, so OpenAI-compatible raw completion requests were rejected instead of flowing through the same staged request-processing architecture used by the rest of SMG.

### Solution

Implement `/v1/completions` as a native gRPC pipeline request type.

`CompletionRequest` now flows through the pipeline as `RequestType::Completion` from HTTP entry through preparation, worker selection, client acquisition, backend request building, execution, and response processing. The branch adds backend-specific completion request builders, a completion preparation stage, a completion request-building stage, and completion response processing for both streaming and non-streaming responses.

The streaming path now emits `CompletionStreamResponse` chunks directly from typed proto responses instead of serializing generate SSE and reparsing it afterward. This keeps stop-decoder behavior aligned with non-streaming completions and removes the fragile SSE rewrite layer.

### Changes

- Add `RequestType::Completion` and `FinalResponse::Completion` to the gRPC pipeline context
- Add `CompletionPreparationStage` for tokenizer resolution, prompt tokenization, and stop-decoder setup
- Add `CompletionRequestBuildingStage` and `GrpcClient::build_completion_request()` for direct `CompletionRequest -> proto::GenerateRequest` backend translation
- Add native completion execution via `RequestPipeline::execute_completion()` using the shared pipeline loop
- Add native completion streaming that emits `CompletionStreamResponse` chunks directly from proto streams
- Align completion stop handling between streaming and non-streaming paths
- Preserve `system_fingerprint` and improve finish-reason/logprob handling for non-streaming completions
- Add explicit gRPC validation/rejection for unsupported completion cases that previously behaved incorrectly:
  - prompt arrays for batched `/v1/completions`
  - streaming completion logprobs
  - Harmony-model completion requests
- Fix TRT-LLM completion parity by keeping gateway-owned echo handling and forwarding guided decoding constraints
- Add completion request validation and focused regression tests

## Test Plan

- [x] `cargo test -p smg completion --quiet`
- [x] `cargo test -p smg --quiet`
- [x] `cargo test -p openai-protocol completion --quiet`
- [x] `cargo fmt --all`
- [x] checked edited files for diagnostics in the IDE linter

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/model_gateway/src/routers/grpc/router.rs
+++ b/model_gateway/src/routers/grpc/router.rs
@@ -1,35 +1,22 @@
-use std::{
-    collections::{HashMap, HashSet},
-    sync::Arc,
-};
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use axum::{
     http::HeaderMap,
     response::{IntoResponse, Response},
 };
-use bytes::Bytes;
-use futures_util::StreamExt;
 use openai_protocol::{
     chat::ChatCompletionRequest,
     classify::ClassifyRequest,
-    common::{StringOrArray, Usage},
-    completion::{
-        CompletionChoice, CompletionRequest, CompletionResponse, CompletionStreamChoice,
-        CompletionStreamResponse,
-    },
+    completion::CompletionRequest,
     embedding::EmbeddingRequest,
-    generate::{GenerateFinishReason, GenerateRequest},
+    generate::GenerateRequest,
     responses::{ResponsesGetParams, ResponsesRequest},
-    sampling_params::SamplingParams,
 };
-use serde_json::Value;
-use tokio::sync::mpsc;
-use tracing::{debug, error};
+use tracing::debug;
 
 use super::{
     common::responses::{
-        build_sse_response,
         handlers::{cancel_response_impl, get_response_impl},
         utils::validate_worker_availability,
         ResponsesContext,
@@ -383,52 +370,6 @@ impl GrpcRouter {
             .await
     }
 
-    /// Build a GenerateRequest from a CompletionRequest, returning the resolved prompt text.
-    fn build_generate_from_completion(
-        body: &CompletionRequest,
-        stream: bool,
-    ) -> (GenerateRequest, String) {
-        let prompt_text = match &body.prompt {
-            StringOrArray::String(s) => s.clone(),
-            StringOrArray::Array(arr) => arr.join(""),
-        };
-
-        let request = GenerateRequest {
-            text: Some(prompt_text.clone()),
-            model: Some(body.model.clone()),
-            sampling_params: Some(SamplingParams {
-                temperature: body.temperature,
-                max_new_tokens: body.max_tokens,
-                top_p: body.top_p,
-                top_k: body.top_k,
-                frequency_penalty: body.frequency_penalty,
-                presence_penalty: body.presence_penalty,
-                repetition_penalty: body.repetition_penalty,
-                stop: body.stop.clone(),
-                ignore_eos: Some(body.ignore_eos),
-                n: body.n,
-                min_p: body.min_p,
-                min_new_tokens: body.min_tokens,
-                regex: body.regex.clone(),
-                ebnf: body.ebnf.clone(),
-                json_schema: body.json_schema.clone(),
-                stop_token_ids: body.stop_token_ids.clone(),
-                no_stop_trim: Some(body.no_stop_trim),
-                skip_special_tokens: Some(body.skip_special_tokens),
-                sampling_seed: body.sampling_seed.or_else(|| body.seed.map(|s| s as u64)),
-            }),
-            stream,
-            return_logprob: body.logprobs.map(|_| true),
-            top_logprobs_num: body.logprobs.map(|l| l as i32),
-            lora_path: body.lora_path.clone(),
-            session_params: body.session_params.clone(),
-            return_hidden_states: body.return_hidden_states,
-            ..Default::default()
-        };
-
-        (request, prompt_text)
-    }
-
     /// Main route_completion implementation
     async fn route_completion_impl(
         &self,
@@ -442,32 +383,11 @@ impl GrpcRouter {
             body.stream
         );
 
-        if body.stream {
-            self.route_completion_streaming(headers, body, model_id)
-                .await
-        } else {
-            self.route_completion_non_streaming(headers, body, model_id)
-                .await
-        }
-    }
-
-    /// Non-streaming /v1/completions: run through generate pipeline, wrap as CompletionResponse.
-    async fn route_completion_non_streaming(
-        &self,
-        headers: Option<&HeaderMap>,
-        body: &CompletionRequest,
-        model_id: Option<&str>,
-    ) -> Response {
-        let (gen_request, prompt_text) = Self::build_generate_from_completion(body, false);
-
-        let request = Arc::new(gen_request);
+        let request = Arc::new(body.clone());
         let headers_cloned = headers.cloned();
         let model_id_cloned = model_id.map(|s| s.to_string());
         let components = self.shared_components.clone();
         let pipeline = &self.pipeline;
-        let model = body.model.clone();
-        let echo = body.echo;
-        let suffix = body.suffix.clone();
 
         RetryExecutor::execute_response_with_retry(
             &self.retry_config,
@@ -476,25 +396,10 @@ impl GrpcRouter {
                 let headers = headers_cloned.clone();
                 let model_id = model_id_cloned.clone();
                 let components = Arc::clone(&components);
-                let model = model.clone();
-                let prompt_text = prompt_text.clone();
-                let suffix = suffix.clone();
                 async move {
-                    let gen_responses = match pipeline
-                        .execute_generate_typed(request, headers, model_id, components)
+                    pipeline
+                        .execute_completion(request, headers, model_id, components)
                         .await
-                    {
-                        Ok(responses) => responses,
-                        Err(response) => return response,
-                    };
-
-                    build_completion_response(
-                        gen_responses,
-                        &model,
-                        &prompt_text,
-                        echo,
-                        suffix.as_deref(),
-                    )
                 }
             },
             |res, _attempt| is_retryable_status(res.status()),
@@ -513,72 +418,6 @@ impl GrpcRouter {
             },
         )
         .await
-    }
-
-    /// Streaming /v1/completions: run through generate pipeline, transform SSE from
-    /// SGLang generate format to OpenAI CompletionStreamResponse format.
-    async fn route_completion_streaming(
-        &self,
-        headers: Option<&HeaderMap>,
-        body: &CompletionRequest,
-        model_id: Option<&str>,
-    ) -> Response {
-        let (gen_request, prompt_text) = Self::build_generate_from_completion(body, true);
-
-        let request = Arc::new(gen_request);
-        let headers_cloned = headers.cloned();
-        let model_id_cloned = model_id.map(|s| s.to_string());
-        let components = self.shared_components.clone();
-        let pipeline = &self.pipeline;
-
-        let response = RetryExecutor::execute_response_with_retry(
-            &self.retry_config,
-            |_attempt| {
-                let request = Arc::clone(&request);
-                let headers = headers_cloned.clone();
-                let model_id = model_id_cloned.clone();
-                let components = Arc::clone(&components);
-                async move {
-                    pipeline
-                        .execute_generate(request, headers, model_id, components)
-                        .await
-                }
-            },
-            |res, _attempt| is_retryable_status(res.status()),
-            |delay, attempt| {
-                Metrics::record_worker_retry(
-                    metrics_labels::WORKER_REGULAR,
-                    metrics_labels::ENDPOINT_COMPLETIONS,
-                );
-                Metrics::record_worker_retry_backoff(attempt, delay);
-            },
-            || {
-                Metrics::record_worker_retries_exhausted(
-                    metrics_labels::WORKER_REGULAR,
-                    metrics_labels::ENDPOINT_COMPLETIONS,
-                );
-            },
-        )
-        .await;
-
-        if !response.status().is_success() {
-            return response;
-        }
-
-        let created = std::time::SystemTime::now()
-            .duration_since(std::time::UNIX_EPOCH)
-            .unwrap_or_default()
-            .as_secs();
-
-        let echo_prompt = if body.echo { Some(prompt_text) } else { None };
-
-        transform_generate_sse_to_completion_sse(
-            response,
-            body.model.clone(),
-            created,
-            echo_prompt,
-            body.suffix.clone(),
-        )
     }
 }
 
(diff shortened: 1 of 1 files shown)


model_gateway/src/routers/grpc/pipeline.rs
```diff
@@ -664,6 +664,72 @@ impl RequestPipeline {
         }
     }
 
+    /// Execute the generate pipeline and return typed results
+    ///
+    /// Used by route_completion to get typed Vec<GenerateResponse> without
+    /// serializing to JSON and re-parsing. Follows the same pattern as
+    /// execute_chat_for_responses.
+    pub async fn execute_generate_typed(
```
slin1237: we shouldnt do this
write actual completion pipeline

vschandramourya:
@CatherineSue @slin1237 Thanks for the feedback. Updated the PR description, implementation structure to fit existing infra and the unit tests. Please let me know if any changes needed.

model_gateway/src/routers/grpc/common/stages/dispatch_metadata.rs
```diff
@@ -32,14 +32,11 @@ impl PipelineStage for DispatchMetadataStage {
         let request_id = proto_request.request_id().to_string();
         let model = match &ctx.input.request_type {
             RequestType::Chat(req) => req.model.clone(),
-            RequestType::Generate(_req) => {
-                // Generate requests don't have a model field
```
CatherineSue, line 36: It is not a good idea to remove important comments like this.

model_gateway/src/routers/grpc/regular/stages/completion/preparation.rs
```diff
@@ -0,0 +1,211 @@
...
+        self.generate_stage.execute(ctx).await
+    }
+
+    fn name(&self) -> &'static str {
+        "CompletionPreparation"
+    }
+}
+
+/// Convert a `CompletionRequest` into a `GenerateRequest`, mapping all applicable fields.
+pub(crate) fn build_generate_from_completion(body: &CompletionRequest) -> GenerateRequest {
```
CatherineSue, line 55: This is not the right way. It is the same workaround as the previous approach. Bypassing the actual flow from router to grpc_client, then to each worker backend is not the way to implement pipeline. CompletionRequest should be converted using RPC `Generate` in the request_building step. Not here.

And the reason we have preparation step is for the tokenizer to encode everything, so worker selection could assign the best worker for the request based on its load and the current policy.

model_gateway/src/routers/grpc/regular/stages/completion/preparation.rs
```diff
@@ -0,0 +1,211 @@
+//! Completion preparation stage: Convert CompletionRequest to GenerateRequest,
```
CatherineSue, line 1: Please try to use `git cp` or just cp the `chat/preparation.rs` first then update upon it. The file should not differ much.

model_gateway/src/routers/grpc/regular/stages/completion/preparation.rs
```diff
@@ -0,0 +1,211 @@
...
+        return_hidden_states: body.return_hidden_states,
+        ..Default::default()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn sample_completion_request() -> CompletionRequest {
```
CatherineSue, line 99: All the tests are unncessary.

model_gateway/src/routers/grpc/regular/stages/completion/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Completion endpoint pipeline stages
+//!
+//! Completions are processed by converting to a GenerateRequest, running
+//! through the generate preparation pipeline, then wrapping the generate
+//! response as a CompletionResponse or CompletionStreamResponse.
```
CatherineSue, line 5: This module doc describes the wrong architecture. Completions should NOT be "processed by converting to a GenerateRequest". They should flow natively as `RequestType::Completion` through every pipeline stage, just like `ChatCompletionRequest` flows as `RequestType::Chat`. The `chat/` stages don't convert to Generate — neither should completions.

model_gateway/src/routers/grpc/regular/stages/completion/preparation.rs
```diff
@@ -0,0 +1,211 @@
...
+
+#[async_trait]
+impl PipelineStage for CompletionPreparationStage {
+    async fn execute(&self, ctx: &mut RequestContext) -> Result<Option<Response>, Response> {
+        let completion_req = ctx.completion_request_arc();
+        let gen_request = Arc::new(build_generate_from_completion(&completion_req));
+
+        // Preserve original for response formatting, swap type so middle stages see Generate
+        ctx.state.original_completion_request = Some(completion_req);
+        ctx.input.request_type = RequestType::Generate(gen_request);
```
CatherineSue, line 44: **Blocker:** This is the core issue. Mutating `ctx.input.request_type` from `Completion` to `Generate` mid-pipeline breaks the type-dispatch contract that the entire pipeline is built on. No other request type does this — `Chat` stays `Chat`, `Generate` stays `Generate`, `Embedding` stays `Embedding`. This type swap forces every downstream stage to either (a) see it as a generate request when it isn't, or (b) check a side-channel field to detect the conversion.

The correct approach: write a preparation stage that does what `GeneratePreparationStage` does — resolve tokenizer, `tokenizer.encode(prompt_text, false)`, create `stop_decoder`, populate `PreparationOutput.token_ids` — without swapping the type. ~30 lines.

Look at `GeneratePreparationStage::prepare_generate()` — it's 40 lines of straightforward tokenize-and-prepare logic. A `CompletionPreparationStage` that does the same thing (extract prompt text from `CompletionRequest`, tokenize, create stop decoder) would be similarly simple and wouldn't require any of the type conversion machinery.

model_gateway/src/routers/grpc/regular/stages/completion/preparation.rs
CatherineSue, line 55: **Blocker:** This function should not exist. You're converting `CompletionRequest` → `GenerateRequest` (OpenAI-type to OpenAI-type), which is the wrong abstraction boundary. The pipeline converts OpenAI types to *proto* types at the request_building stage via backend-specific client adapters.

Look at how chat does it: `ChatCompletionRequest` is never converted to `GenerateRequest`. It goes directly from `ChatCompletionRequest` → `proto::GenerateRequest` via `builder_client.build_chat_request()` in `chat/request_building.rs:84`. Completions should follow the same pattern: `CompletionRequest` → `proto::GenerateRequest` via a new `builder_client.build_completion_request()` in a `completion/request_building.rs`.

This is exactly how vLLM, SGLang, and TensorRT-LLM handle it internally — each has a separate serving class (`OpenAIServingCompletion`) that converts `CompletionRequest` directly to the internal generate format, not by bouncing through another OpenAI request type.

model_gateway/src/routers/grpc/context.rs
```diff
@@ -92,6 +94,10 @@ pub(crate) struct ProcessingState {
 
     // Stage 6: Response processing state
     pub response: ResponseState,
+
+    /// Original completion request, preserved when CompletionPreparationStage
+    /// converts the request type to Generate for pipeline compatibility.
+    pub original_completion_request: Option<Arc<CompletionRequest>>,
```
CatherineSue, line 100: **Blocker:** This field only exists because the preparation stage converts `Completion` → `Generate`. If `RequestType::Completion` survived the pipeline (like every other request type does), you'd just call `ctx.completion_request()` in the response processing stage — exactly like chat calls `ctx.chat_request()`. No extra field needed. Keep the type intact throughout.

model_gateway/src/routers/grpc/regular/stages/response_processing.rs
```diff
@@ -48,9 +54,16 @@ impl ResponseProcessingStage {
 #[async_trait]
 impl PipelineStage for ResponseProcessingStage {
     async fn execute(&self, ctx: &mut RequestContext) -> Result<Option<Response>, Response> {
+        // Completion requests are converted to Generate in the preparation stage,
+        // so we check original_completion_request to route to the correct handler.
+        if ctx.state.original_completion_request.is_some() {
+            return self.completion_stage.execute(ctx).await;
```
CatherineSue, line 60: **Blocker:** This `if` check runs *before* the `match &ctx.input.request_type` dispatch because the type was already swapped to `Generate` in the preparation stage — the match arm at line 63 would route it to `GenerateResponseProcessingStage` otherwise.

This is a fragile ordering dependency. Every other request type dispatches cleanly on `request_type`. Completion is the only one that needs a pre-match escape hatch via a side-channel field. This is a direct consequence of the type laundering in the preparation stage.

model_gateway/src/routers/grpc/regular/stages/request_building.rs
```diff
@@ -41,14 +41,14 @@ impl PipelineStage for RequestBuildingStage {
             RequestType::Generate(_) => self.generate_stage.execute(ctx).await,
             RequestType::Embedding(_) => self.embedding_stage.execute(ctx).await,
             RequestType::Classify(_) => self.embedding_stage.execute(ctx).await,
-            RequestType::Responses(_request) => {
+            RequestType::Completion(_) | RequestType::Responses(_) => {
```
CatherineSue, line 44: `Completion` explicitly errors in request_building. The only reason this doesn't crash at runtime is because the preparation stage already swapped the type to `Generate`. If `RequestType::Completion` survived the pipeline intact (as it should), you'd need a `CompletionRequestBuildingStage` that calls `builder_client.build_completion_request(...)` — analogous to `ChatRequestBuildingStage` calling `builder_client.build_chat_request(...)` at `chat/request_building.rs:84`.

The fact that this error arm is needed shows that the type-swap approach doesn't integrate cleanly with the pipeline's stage dispatch model.

model_gateway/src/routers/grpc/regular/stages/completion/response_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+            .ok_or_else(|| {
+                error!(
+                    function = "CompletionResponseProcessingStage::execute",
+                    "original_completion_request not set — was CompletionPreparationStage skipped?"
+                );
+                error::internal_error(
+                    "missing_completion_request",
+                    "Original completion request not preserved by preparation stage",
+                )
+            })?;
```
CatherineSue, line 90: This reads the original request from `ctx.state.original_completion_request` which was stored during preparation before the type swap. The error message at line 84 even hints at the fragility: `"was CompletionPreparationStage skipped?"`. If `RequestType::Completion` survived the pipeline, this would just be `ctx.completion_request_arc()` — no `Option`, no error path, no question about stage ordering.

model_gateway/src/routers/grpc/regular/stages/completion/response_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+
+            let transformed = transform_generate_sse_to_completion_sse(
+                response,
+                completion_req.model.clone(),
+                created,
+                echo_prompt,
+                completion_req.suffix.clone(),
+            );
+
+            return Ok(Some(transformed));
```
CatherineSue, line 157: **Blocker:** You correctly call `process_streaming_generate()` at line 125 — which properly decodes proto responses, handles token decoding via `StopSequenceDecoder`, and produces a well-formed SSE response. But then at line 149 you pipe the *fully-formed* SSE Response into `transform_generate_sse_to_completion_sse()`, which decomposes it back into raw bytes and re-parses the JSON.

This is: proto → decoded tokens → SSE bytes → re-parse bytes → re-serialize as different SSE.

The correct approach is to transform the stream *before* it gets serialized to SSE, not after. Either write a completion-specific streaming processor that emits `CompletionStreamResponse` chunks directly from proto responses, or transform the decoded token chunks before SSE serialization.

model_gateway/src/routers/grpc/regular/stages/completion/response_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+        system_fingerprint: None,
+    }
+}
+
+/// Transform an SSE response from SGLang generate format to OpenAI CompletionStreamResponse.
+///
+/// SGLang emits accumulated text per chunk; this function computes deltas per index so
+/// the output matches the OpenAI `/v1/completions` streaming contract.
+///
+fn transform_generate_sse_to_completion_sse(
```
CatherineSue, line 283: **Blocker:** This entire ~180-line function is a hand-rolled SSE parser that manually:
1. Decomposes a fully-formed Response back into raw bytes (line 290)
2. Parses SSE framing by searching for `\n\n` boundaries (line 315)
3. Strips `data: ` prefixes (line 323)
4. Deserializes to untyped `serde_json::Value` (line 332)
5. Computes text deltas via byte-length tracking (line 359-361)
6. Re-serializes to new SSE

The pipeline already has `StreamingProcessor` that handles all of this correctly using typed proto responses and proper token decoding via `StopSequenceDecoder`. This is the same raw SSE parsing approach from the original PR, moved into the pipeline structure. It should be replaced with proper use of the existing `StreamingProcessor` infrastructure.

model_gateway/src/routers/grpc/regular/stages/completion/response_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+                }
+
+                let prev_len = prev_lens.entry(index).or_insert(0);
+                let mut delta = String::new();
+                if echo_prompt.is_some() && echo_sent_for.insert(index) {
+                    delta.push_str(echo_prompt.as_deref().unwrap_or(""));
+                }
+                if accumulated_text.len() > *prev_len {
+                    delta.push_str(&accumulated_text[*prev_len..]);
+                }
```
CatherineSue, line 361: **Bug:** `accumulated_text.len()` returns **byte** length. `&accumulated_text[*prev_len..]` slices at a **byte** offset. If `prev_len` lands in the middle of a multi-byte UTF-8 character (CJK text, emoji, accented characters), this panics at runtime with `byte index N is not a char boundary`. This is a direct consequence of re-parsing serialized SSE text instead of working with the typed proto responses that the `StreamingProcessor` already provides.

model_gateway/src/routers/grpc/common/stages/dispatch_metadata.rs
```diff
@@ -32,14 +32,11 @@ impl PipelineStage for DispatchMetadataStage {
...
-                // Use model_id from input or UNKNOWN_MODEL_ID
-                ctx.input
-                    .model_id
-                    .clone()
-                    .unwrap_or_else(|| UNKNOWN_MODEL_ID.to_string())
-            }
+            RequestType::Generate(_) | RequestType::Completion(_) => ctx
```
CatherineSue, line 35: `CompletionRequest` has a required `.model` field (per the OpenAI spec). Grouping it with `Generate` (which legitimately lacks a model field) means it falls back to `ctx.input.model_id` instead of reading `req.model` directly. This should be:

```rust
RequestType::Completion(req) => req.model.clone(),
```

Same pattern as `RequestType::Chat(req) => req.model.clone()` on line 34.

protocols/src/generate.rs
```diff
@@ -173,6 +173,48 @@ impl Normalizable for GenerateRequest {
     // Use default no-op implementation - no normalization needed for GenerateRequest
 }
 
+impl Default for GenerateRequest {
```
CatherineSue, line 176: This manual `Default` impl only exists to support `build_generate_from_completion()` using `..Default::default()`. If you stop converting `CompletionRequest` → `GenerateRequest`, this is unnecessary. It's also a maintenance hazard — every time a field is added to `GenerateRequest`, someone has to remember to update this manual impl (vs `#[derive(Default)]` which would be automatic and correct).

model_gateway/src/routers/grpc/pipeline.rs
```diff
@@ -664,6 +666,100 @@ impl RequestPipeline {
         }
     }
 
+    /// Execute the complete pipeline for a completion request
+    ///
+    /// The completion preparation stage converts `CompletionRequest` into a `GenerateRequest`
+    /// and stores the original for response formatting. The response processing stage wraps
```
CatherineSue, line 672: nit: This is a line-for-line copy of `execute_generate` (line 365) with `ENDPOINT_GENERATE` → `ENDPOINT_COMPLETIONS` and `FinalResponse::Generate` → `FinalResponse::Completion`. ~94 lines of pure boilerplate. Consider factoring the pipeline execution loop into a generic helper parameterized by endpoint label and response variant.

model_gateway/src/routers/grpc/harmony/stages/request_building.rs
```diff
@@ -82,14 +82,14 @@ impl PipelineStage for HarmonyRequestBuildingStage {
                     "Embedding requests are not supported with Harmony models".to_string(),
                 ));
             }
-            RequestType::Classify(_) => {
+            RequestType::Classify(_) | RequestType::Completion(_) => {
```
CatherineSue, line 85: nit: Each unsupported type should have its own descriptive error code. Changing `harmony_classify_not_supported` to the generic `harmony_not_supported` loses specificity in error tracking. Keep them separate:

```rust
RequestType::Classify(_) => { ... "harmony_classify_not_supported" ... }
RequestType::Completion(_) => { ... "harmony_completion_not_supported" ... }
```

model_gateway/src/routers/grpc/regular/stages/completion/response_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+            }
+            text.push_str(&gen.text);
+            if let Some(sfx) = suffix {
+                text.push_str(sfx);
+            }
+
+            CompletionChoice {
+                text,
+                index: i as u32,
+                logprobs: None,
```
CatherineSue, line 249: `logprobs: None` — but the preparation stage passes `return_logprob: true` and `top_logprobs_num` to the backend when `CompletionRequest.logprobs` is set. The backend computes logprobs, sends them back in `GenerateResponse.meta_info`, and they're silently discarded here.

Either extract logprobs from `gen.meta_info.output_token_logprobs` and return them, or reject completion requests with `logprobs != None` upfront. Current behavior silently ignores a user-requested feature.

model_gateway/src/routers/grpc/regular/stages/completion/response_processing.rs
```diff
@@ -0,0 +1,505 @@
...
+            completion_req.echo,
+            completion_req.suffix.as_deref(),
+        );
+
+        ctx.state.response.final_response = Some(FinalResponse::Completion(completion_response));
+        Ok(None)
+    }
+}
+
+fn resolve_prompt_text(req: &CompletionRequest) -> String {
```
CatherineSue, line 201: This is the same `match &req.prompt { String(s) => s.clone(), Array(arr) => arr.join("") }` logic that exists in `preparation.rs:56-59`. Duplicated prompt resolution — should be a single shared utility, or better yet, cache the resolved prompt text in `ProcessingState` during preparation so response processing doesn't have to re-derive it.

model_gateway/src/routers/grpc/context.rs
```diff
@@ -341,6 +365,31 @@ impl RequestContext {
         }
     }
 
+    /// Get completion request (panics if not completion)
+    #[expect(
+        dead_code,
+        clippy::panic,
+        reason = "typed accessor: caller guarantees variant via RequestType construction"
+    )]
+    pub fn completion_request(&self) -> &CompletionRequest {
```
CatherineSue, line 374: `#[expect(dead_code)]` — this accessor is dead code precisely because the preparation stage swaps `RequestType::Completion` to `RequestType::Generate`. No downstream stage ever sees `Completion`, so nothing calls this. If the type survived the pipeline (as it should), this accessor would be actively used by the response processing stage instead of reading from `ctx.state.original_completion_request`.

vschandramourya:
Closing this PR, since its addressed in others PR starting from #840

PR #672, opened 2026-03-07
fix(scripts): handle empty arrays in check_release_versions.sh

slin1237:
## Summary

Fixes `make check-versions` crashing with `NEEDS_PY_BUMP[@]: unbound variable` when Python package arrays are empty.

## What changed

- `scripts/check_release_versions.sh`: use `${arr[@]+"${arr[@]}"}` pattern for safe expansion of potentially-empty arrays in for-loops, and cache array lengths into local variables before conditionals.

## Why

Bash 3.x (macOS default) treats `${empty_array[@]}` as an unbound variable under `set -u` even after `=()` initialization. When `NEEDS_PY_BUMP` had no entries appended (e.g. the Python package already had its version bumped), the script crashed at line 388.

## How

- `${arr[@]+"${arr[@]}"}` — the `+` parameter expansion only expands if the variable is set and non-empty, avoiding the nounset error.
- Pre-cached `n_bump`, `n_ws`, `n_py` variables replace inline `${#arr[@]}` in guard conditionals to avoid the same issue.

## Test plan

- [x] Run `make check-versions` on macOS (bash 3.x) with no Python packages needing bumps — no crash
- [x] Run `make check-versions` with crates needing bumps — proposed fixes display correctly

## Summary by CodeRabbit

* **Chores**
  * Improved internal script reliability with enhanced array handling and null-safety checks to reduce potential edge-case errors.

### DIFF
--- a/scripts/check_release_versions.sh
+++ b/scripts/check_release_versions.sh
@@ -385,7 +385,10 @@ echo -e "${RED}${BOLD}$issues issue(s) found.${NC}"
 # ---------------------------------------------------------------------------
 # Phase 2: Offer to fix
 # ---------------------------------------------------------------------------
-total_fixes=$(( ${#NEEDS_BUMP[@]} + ${#NEEDS_WS_SYNC[@]} + ${#NEEDS_PY_BUMP[@]} ))
+n_bump=${#NEEDS_BUMP[@]}
+n_ws=${#NEEDS_WS_SYNC[@]}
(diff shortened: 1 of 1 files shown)


docs/plans/2026-03-05-workspace-reorganization-design.md
```diff
@@ -0,0 +1,85 @@
+# Workspace Reorganization Design
```
CatherineSue: This file is added by accident, it seems?

PR #679, opened 2026-03-09
feat(interactions): Register Gemini Router

XinyueZhang369:
### Problem

The model gateway currently supports OpenAI and Anthropic as external provider backends, but lacks support for Google's Gemini API. Users who want to route requests through Gemini-compatible endpoints cannot do so.

### Solution

Add the routing infrastructure and registration plumbing for Gemini as a new routing mode in the model gateway. This PR wires up the config, CLI, job queue, router factory, server endpoint, and router manager so that the
   Gemini router is fully registered and routable. The router implementation stubs (worker selection, request building, upstream execution, response processing) and integration tests will follow in a subsequent PR. Refer to https://github.com/lightseekorg/smg/pull/417 for more details of the gemini router structure 

## Changes

- Gemini routing mode: Add `RoutingMode::Gemini` variant to config types, validation, CLI `--backend gemini` flag, and job queue worker initialization
- Job queue refactor: Consolidate duplicate OpenAI/Anthropic worker submission into shared `submit_external_worker_jobs` helper, add Gemini to the combined match arm
 - Router registration: Register `GeminiRouter` in the router factory (both single-router and IGW multi-router modes), add `HTTP_GEMINI` router ID, include in `create_all_routers`
- Server endpoint: Add `POST /v1/interactions` route with `ValidatedJson<InteractionsRequest>` extractor
- RouterTrait: Add default `route_interactions` method (returns 501) so existing routers are unaffected
- Router manager: Implement `route_interactions` dispatch and provider-aware routing in `get_router_for_model` using `ProviderType`
- GeminiRouter: Construct from `AppContext` with `SharedComponents` (client, worker_registry, mcp_orchestrator, request_timeout), implement `RouterTrait::route_interactions`
- Protocol: Implement `Normalizable` for `InteractionsRequest` for `ValidatedJson` extractor compatibility

## Test Plan
- Existing tests pass — no behavioral changes to OpenAI/Anthropic paths
- `cargo build` succeeds with new Gemini config variants
- Integration tests for the Gemini router (mock server, round-trip, error cases, circuit breaker) will be added in the follow-up PR
- run ` cargo run -p smg --bin smg -- --backend gemini --worker-urls "https://generativelanguage.googleapis.com" --port 8080` to start smg for gemini backend. The router starts successfully and return 501 not implemented for /v1/interactions request

Screenshot 2026-03-09 at 9 13 41 AM
Screenshot 2026-03-09 at 4 16 07 PM

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Gemini backend added with configurable worker URLs, routing, and HTTP-only backend handling
  * New /v1/interactions endpoint and Interactions API routing support
  * Gemini model discovery via /v1beta/models and provider-specific auth (x-goog-api-key)
  * Worker selection, payload transformation, upstream request/response handling, and metadata patching for Gemini flows

* **Tests**
  * Comprehensive end-to-end tests and a Mock Gemini server for interactions testing

### DIFF
--- a/crates/protocols/src/interactions.rs
+++ b/crates/protocols/src/interactions.rs
@@ -11,6 +11,7 @@ use validator::{Validate, ValidationError};
 use super::{
     common::{default_model, default_true, Function, GenerationRequest},
     sampling_params::validate_top_p_value,
+    validated::Normalizable,
 };
 
 // ============================================================================
@@ -91,6 +92,10 @@ impl Default for InteractionsRequest {
             store: true,
         }
     }
+}
+
+impl Normalizable for InteractionsRequest {
+    // Use default no-op implementation
 }
 
 impl GenerationRequest for InteractionsRequest {
--- a/model_gateway/src/main.rs
+++ b/model_gateway/src/main.rs
@@ -63,6 +63,8 @@ pub enum Backend {
     Openai,
     #[value(name = "anthropic")]
     Anthropic,
+    #[value(name = "gemini")]
+    Gemini,
 }
 
 impl std::fmt::Display for Backend {
@@ -73,6 +75,7 @@ impl std::fmt::Display for Backend {
             Backend::Trtllm => "trtllm",
             Backend::Openai => "openai",
             Backend::Anthropic => "anthropic",
+            Backend::Gemini => "gemini",
         };
         write!(f, "{s}")
     }
@@ -955,6 +958,10 @@ impl CliArgs {
             RoutingMode::Anthropic {
                 worker_urls: self.worker_urls.clone(),
             }
+        } else if matches!(self.backend, Some(Backend::Gemini)) {
+            RoutingMode::Gemini {
+                worker_urls: self.worker_urls.clone(),
+            }
         } else if self.pd_disaggregation {
             RoutingMode::PrefillDecode {
                 prefill_urls,
@@ -1017,6 +1024,9 @@ impl CliArgs {
                 all_urls.extend(worker_urls.clone());
             }
             RoutingMode::Anthropic { worker_urls } => {
+                all_urls.extend(worker_urls.clone());
+            }
+            RoutingMode::Gemini { worker_urls } => {
                 all_urls.extend(worker_urls.clone());
             }
         }
--- a/model_gateway/src/routers/gemini/router.rs
+++ b/model_gateway/src/routers/gemini/router.rs
@@ -3,21 +3,18 @@
 use std::{
     any::Any,
     sync::{atomic::AtomicBool, Arc},
+    time::Duration,
 };
 
 use async_trait::async_trait;
-use axum::{
-    http::{HeaderMap, StatusCode},
-    response::{IntoResponse, Response},
-};
-use openai_protocol::{chat::ChatCompletionRequest, interactions::InteractionsRequest};
-use smg_mcp::McpOrchestrator;
+use axum::{http::HeaderMap, response::Response};
+use openai_protocol::interactions::InteractionsRequest;
 
 use super::{
     context::{RequestContext, SharedComponents},
     driver,
 };
-use crate::{core::WorkerRegistry, routers::RouterTrait};
+use crate::{app_context::AppContext, routers::RouterTrait};
 
 pub struct GeminiRouter {
     shared_components: Arc<SharedComponents>,
@@ -32,47 +29,26 @@ impl std::fmt::Debug for GeminiRouter {
 }
 
 impl GeminiRouter {
-    /// Create a new `GeminiRouter`.
-    pub fn new(
-        worker_registry: Arc<WorkerRegistry>,
-        mcp_orchestrator: Arc<McpOrchestrator>,
-        client: reqwest::Client,
-    ) -> Self {
+    /// Create a new `GeminiRouter` from the application context.
+    pub fn new(ctx: Arc<AppContext>) -> Result<Self, String> {
+        let mcp_orchestrator = ctx
+            .mcp_orchestrator
+            .get()
+            .ok_or_else(|| "Gemini router requires MCP orchestrator".to_string())?
+            .clone();
+
+        let request_timeout = Duration::from_secs(ctx.router_config.request_timeout_secs);
+
         let shared_components = Arc::new(SharedComponents {
-            client,
-            worker_registry,
+            client: ctx.client.clone(),
+            worker_registry: ctx.worker_registry.clone(),
             mcp_orchestrator,
+            request_timeout,
         });
-        Self {
+        Ok(Self {
             shared_components,
             healthy: AtomicBool::new(true),
-        }
-    }
-
-    /// Main handler for `POST /v1/interactions`.
-    ///
-    /// Builds a `RequestContext` and runs the driver state machine.
-    ///
-    /// For **non-streaming** requests the driver returns the HTTP response directly.
-    ///
-    /// For **streaming** requests the terminal streaming step (`StreamRequest` or
-    /// `StreamRequestWithTool`) creates an SSE channel internally, spawns the
-    /// streaming work in a background task, and returns the SSE `Response` — the
-    /// same pattern as `process_streaming_response` in the gRPC router.
-    pub async fn route_interactions(
-        &self,
-        headers: Option<HeaderMap>,
-        body: InteractionsRequest,
-        model_id: Option<String>,
-    ) -> Response {
-        let mut ctx = RequestContext::new(
-            Arc::new(body),
-            headers,
-            model_id,
-            self.shared_components.clone(),
-        );
-
-        driver::execute(&mut ctx).await
+        })
     }
 }
 
@@ -90,12 +66,19 @@ impl RouterTrait for GeminiRouter {
         "gemini"
     }
 
-    async fn route_chat(
+    async fn route_interactions(
         &self,
-        _headers: Option<&HeaderMap>,
-        _body: &ChatCompletionRequest,
-        _model_id: Option<&str>,
+        headers: Option<&HeaderMap>,
+        body: &InteractionsRequest,
+        model_id: Option<&str>,
     ) -> Response {
-        (StatusCode::NOT_IMPLEMENTED, "Not implemented").into_response()
+        let mut ctx = RequestContext::new(
+            Arc::new(body.clone()),
+            headers.cloned(),
+            model_id.map(String::from),
+            self.shared_components.clone(),
+        );
+
+        driver::execute(&mut ctx).await
     }
 }
--- a/model_gateway/src/routers/router_manager.rs
+++ b/model_gateway/src/routers/router_manager.rs
@@ -21,6 +21,7 @@ use openai_protocol::{
     completion::CompletionRequest,
     embedding::EmbeddingRequest,
     generate::GenerateRequest,
+    interactions::InteractionsRequest,
     messages::CreateMessageRequest,
     rerank::RerankRequest,
     responses::{ResponsesGetParams, ResponsesRequest},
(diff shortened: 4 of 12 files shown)


model_gateway/src/core/steps/worker/external/discover_models.rs
```diff
@@ -53,6 +53,23 @@ pub struct ModelInfo {
     pub owned_by: Option<String>,
 }
 
+/// Gemini /v1/models response format.
+/// Returns `{models: [{name: "models/gemini-2.5-flash", ...}]}`.
+#[derive(Debug, Clone, Deserialize)]
+pub struct GeminiModelsResponse {
+    pub models: Vec<GeminiModelInfo>,
```
CatherineSue: Question: do we need `GeminiModelsResponse` and `GeminiModelInfo` for Phase 1? Anthropic doesn't need a separate struct because its `/v1/models` returns an OpenAI-compatible `{data: [{id: ...}]}` format — so the existing `ModelsResponse` works for both OpenAI and Anthropic.

Gemini's format (`{models: [{name: ...}]}`) is indeed different, so I understand the separate structs. But for Phase 1, could we use **wildcard model selection** instead (no admin key, accept any model, forward the user's `x-goog-api-key`)? That way model discovery can be a focused follow-up PR.

If we do keep this, the `if provider == Gemini` branching inside the shared `fetch_models()` will grow with each new provider — worth thinking about a cleaner abstraction when we add more.

model_gateway/src/routers/gemini/steps/worker_selection.rs
```diff
@@ -1,46 +1,222 @@
...
+        .collect();
+
+    futures::future::join_all(futures).await;
+}
+
+/// Fetch models list from a worker and update its model list.
+///
+/// Uses `/v1beta/models` for Gemini (googleapis.com) workers,
+/// `/v1/models` for all others.
+async fn refresh_worker_models(
```
CatherineSue: This `refresh_worker_models()` reimplements the same Gemini-vs-OpenAI model format parsing that exists in `discover_models.rs::fetch_models()`. Having two copies could drift over time.

For Phase 1, could we use wildcard model selection instead? If explicit model refresh is needed later, we could reuse the existing `discover_models` infrastructure rather than maintaining a second implementation.

model_gateway/src/main.rs
```diff
@@ -1019,9 +1026,12 @@ impl CliArgs {
             RoutingMode::Anthropic { worker_urls } => {
                 all_urls.extend(worker_urls.clone());
             }
+            RoutingMode::Gemini { worker_urls } => {
+                all_urls.extend(worker_urls.clone());
+            }
         }
         let connection_mode = match &mode {
-            RoutingMode::OpenAI { .. } => ConnectionMode::Http,
+            RoutingMode::OpenAI { .. } | RoutingMode::Gemini { .. } => ConnectionMode::Http,
```
CatherineSue: Pre-existing issue (not introduced by this PR): `RoutingMode::Anthropic` is missing from this match arm. It falls through to `determine_connection_mode()` which happens to return `Http` because Anthropic URLs don't start with `grpc://` — so it works, but accidentally rather than intentionally.

Since this PR is touching this line, could you add Anthropic here too?
```rust
RoutingMode::OpenAI { .. } | RoutingMode::Anthropic { .. } | RoutingMode::Gemini { .. } => ConnectionMode::Http,
```

The whole `connection_mode` detection logic is a bit odd and could use a cleanup pass later, but this fix would be good to include now.

model_gateway/tests/common/mock_gemini_server.rs
```diff
@@ -0,0 +1,208 @@
+//! Mock Gemini API server for testing the Interactions router.
```
CatherineSue: Suggestion: this mock server might be better suited in the e2e test infrastructure. Unit tests with mocks can miss format mismatches with the real API — e2e tests against an actual Gemini endpoint would give stronger confidence.

crates/protocols/src/interactions.rs
```diff
@@ -93,6 +94,8 @@ impl Default for InteractionsRequest {
     }
 }
 
+impl Normalizable for InteractionsRequest {}
```
CatherineSue: Nit: add a comment like the other no-op impls (`CreateMessageRequest`, `RerankRequest`):
```rust
impl Normalizable for InteractionsRequest {
    // Use default no-op implementation
}
```

model_gateway/src/core/steps/worker/external/discover_models.rs
```diff
@@ -53,6 +53,23 @@ pub struct ModelInfo {
...
+}
+
+/// Individual model information from Gemini /v1/models.
+#[derive(Debug, Clone, Deserialize)]
+#[allow(dead_code)]
+pub struct GeminiModelInfo {
```
CatherineSue: Should use `#[expect(dead_code, reason = "...")]` per project conventions, not `#[allow(dead_code)]`.

model_gateway/src/routers/gemini/router.rs
```diff
@@ -90,12 +91,17 @@ impl RouterTrait for GeminiRouter {
...
     ) -> Response {
-        (StatusCode::NOT_IMPLEMENTED, "Not implemented").into_response()
+        self.route_interactions(
+            headers.cloned(),
+            body.clone(),
+            model_id.map(String::from),
+        )
+        .await
     }
 }
```
CatherineSue, line 84: Question: why are there two `route_interactions` methods?

The inherent method (line 68) takes owned types (`Option<HeaderMap>`, `InteractionsRequest`, `Option<String>`) while this trait method takes references — which forces `.cloned()` / `.clone()` / `.map(String::from)` on every call.

Could the inherent method take references instead to avoid the extra allocations? Or could the trait be designed to pass ownership so there's only one method?

model_gateway/src/core/steps/worker/external/discover_models.rs
CatherineSue, line 37: It is a bit awkward to see model responses defined here. This should be better moved to protocols crate.

CatherineSue: cc @slin1237 @key4ng

model_gateway/src/main.rs
CatherineSue: cc @slin1237 @key4ng

XinyueZhang369:
> Thanks for the work on this! Some feedback below.
> 
> **Size**: At 1380+ additions / 25 files, this is above our 1k line guideline. Consider splitting model discovery (GeminiModelsResponse, worker_selection refresh) into a separate PR — perhaps focus on `/v1/models` support first, then the interactions router steps.
> 
> **Reference**: Please add a reference to PR #417 in the description since this builds directly on that scaffolding.
> 
> **Lint**: Looks like CI lint is currently failing — please fix.
> 
> See inline comments for specifics.

Thanks for reviewing! 
I removed the the model discovery but still having more than 1000+ changes, so I now only keep the changes needed to register the gemini router with router manager. I'll add tests in followup router step implement PR, but I still did manual e2e test that I attached in the PR description, does this sound good?

model_gateway/src/core/steps/worker/external/discover_models.rs
XinyueZhang369: Removed from the current PR, will note this for follow up model discovery PR

model_gateway/src/routers/gemini/router.rs
XinyueZhang369: Ah my bad, the inherent method is a placeholder used in previous basic structure PR as don't wanna touch other file outside gemini dir, but after adding route_interactions in trait, this is no longer needed, I removed it. And use the reference in trait definition mainly follows the existing pattern we have for other APIs

model_gateway/src/core/steps/worker/external/discover_models.rs
XinyueZhang369: Removed from the current PR, will note this for follow up model discovery PR

model_gateway/src/main.rs
XinyueZhang369: Saw you merged the PR https://github.com/lightseekorg/smg/pull/680 to clean this, thanks, I removed this line

model_gateway/src/core/steps/worker/external/discover_models.rs
XinyueZhang369: Yep okay, as the current PR is restricted to registering gemini router only, I'll create second PR to implement gemini steps with wildcard model selection, then third PR to add model discovery

model_gateway/src/routers/gemini/router.rs
```diff
@@ -90,12 +91,17 @@ impl RouterTrait for GeminiRouter {
...
+        model_id: Option<&str>,
     ) -> Response {
-        (StatusCode::NOT_IMPLEMENTED, "Not implemented").into_response()
+        self.route_interactions(
+            headers.cloned(),
+            body.clone(),
+            model_id.map(String::from),
+        )
+        .await
     }
```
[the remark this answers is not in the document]
XinyueZhang369: The InteractionsRequest in RequestContext cannot be reference, as the streaming steps (Phase 2) will spawn background tasks.

model_gateway/src/routers/router_manager.rs
```diff
@@ -225,8 +228,13 @@ impl RouterManager {
                 let is_external = matches!(w.metadata().spec.runtime_type, RuntimeType::External);
 
                 if is_external {
-                    // External workers should be routed via OpenAI-compatible router
-                    return (4, &router_ids::HTTP_OPENAI);
+                    // Route external workers to the correct provider-specific router
+                    let router_id = match w.provider_for_model(model_id) {
+                        Some(ProviderType::Gemini) => &router_ids::HTTP_GEMINI,
+                        Some(ProviderType::Anthropic) => &router_ids::HTTP_ANTHROPIC,
+                        _ => &router_ids::HTTP_OPENAI,
```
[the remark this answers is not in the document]
XinyueZhang369: I think if the user wants to call /v1/chat/completions with Gemini models, they start SMG with --backend openai pointing at the Gemini-compatible endpoint. The OpenAI router handles chat completions fine. Other reviewers please correct me if I'm wrong

model_gateway/src/server.rs
```diff
@@ -580,6 +594,7 @@ pub fn build_app(
         .route("/v1/responses", post(v1_responses))
         .route("/v1/embeddings", post(v1_embeddings))
         .route("/v1/messages", post(v1_messages))
+        .route("/v1/interactions", post(v1_interactions))
```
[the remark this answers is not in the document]
XinyueZhang369: Will implement this later, will keep the current scope for this PR now

model_gateway/tests/common/mock_gemini_server.rs
XinyueZhang369: Yep, will add e2e with actual gemini endpoint in follow up PRs

model_gateway/src/server.rs
```diff
@@ -237,6 +239,18 @@ async fn v1_responses(
...
 
+async fn v1_interactions(
+    State(state): State<Arc<AppState>>,
+    headers: http::HeaderMap,
+    ValidatedJson(body): ValidatedJson<InteractionsRequest>,
+) -> Response {
+    let model_id = body.model.as_deref().or(body.agent.as_deref());
+    state
+        .router
+        .route_interactions(Some(&headers), &body, model_id)
```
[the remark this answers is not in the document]
XinyueZhang369: It should return 400 only when both model and agent is empty, it's issue with test plan, but as we move tests in follow up PR, no action for this now

model_gateway/src/routers/router_manager.rs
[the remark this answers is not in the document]
CatherineSue: The self-hosted engines don't qualifie as `is_external`. The scenario in codex is unlikely to be realistic

PR #687, opened 2026-03-09
feat(realtime-api): realtime webrtc handler

### LINKED ISSUE #249: [Realtime Api] gateway-realtime-webrtc-handler
# Description:
Add WebRTC relay support to SMG's OpenAI Realtime API proxy. SMG now terminates both WebRTC peer connections and bridges all traffic — data channel messages and audio RTP packets — putting SMG in the media data path. This enables future interception of realtime events (MCP tool injection, function call handling) identical to the existing WebSocket proxy.

# File Path:
`model_gateway/src/routers/openai/realtime/webrtc_handler.rs` (part 2)

# Prerequisites:
gateway-realtime-session

# Endpoint:
`POST /v1/realtime/calls`
`POST /v1/realtime/calls/{call_id}/hangup`
`GET /v1/realtime`

# Flow:
### Architecture

```mermaid
graph LR
    Client -->|WebRTC ICE-lite| SMG[SMG Gateway]
    SMG -->|WebRTC Full ICE| OpenAI
```

### Signaling:

```mermaid
sequenceDiagram
    participant C as Client
    participant S as SMG Gateway
    participant O as OpenAI

    C->>S: POST /v1/realtime/calls (client SDP offer)
    Note over S: Create upstream Rtc (offerer)
    S->>O: POST /v1/realtime/calls (SMG SDP offer + auth)
    O-->>S: 201 Created (upstream SDP answer)
    Note over S: Accept upstream answer
    Note over S: Create client Rtc (ICE-lite answerer)
    S-->>C: 201 Created (SMG SDP answer)
```
### Media Relay:

```mermaid
sequenceDiagram
    participant C as Client
    participant S as SMG Gateway
    participant O as OpenAI

    Note over S: tokio::select! loop polls both UDP sockets

    C->>S: UDP (STUN/DTLS/RTP)
    Note over S: client_rtc processes input
    S->>O: ChannelData / RTP via upstream_rtc

    O->>S: UDP (STUN/DTLS/RTP)
    Note over S: upstream_rtc processes input
    S->>C: ChannelData / RTP via client_rtc
```
### Teardown:

```mermaid
sequenceDiagram
    participant C as Client
    participant S as SMG Gateway
    participant O as OpenAI

    C->>S: POST /v1/realtime/calls/{id}/hangup
    Note over S: cancel_token.cancel()
    Note over S: bridge loop exits, both peers disconnect
    Note over S: registry.remove_call()
 
    S-->>C: 200 OK
```

# Request:
**Multipart path** (Content-Type: multipart/form-data):

```
POST /v1/realtime/calls
Authorization: Bearer <api-key>
Content-Type: multipart/form-data; boundary=...

--boundary
Content-Disposition: form-data; name="sdp"
Content-Type: application/sdp

v=0
o=- 0 0 IN IP4 0.0.0.0
...
--boundary
Content-Disposition: form-data; name="session"
Content-Type: application/json

{"model": "gpt-4o-realtime-preview-2024-12-17", "modalities": ["audio", "text"]}
--boundary--
```

**Direct SDP path** (Content-Type: application/sdp):

```
POST /v1/realtime/calls?model=gpt-4o-realtime-preview-2024-12-17
Authorization: Bearer <api-key>
Content-Type: application/sdp

v=0
o=- 0 0 IN IP4 0.0.0.0
...
```
**Hangup:**

```
POST /v1/realtime/calls/{call_id}/hangup
Authorization: Bearer <api-key>
```
# Response:

**Create call** — 201 Created:

```
Content-Type: application/sdp

v=0
o=- 0 0 IN IP4 <smg-candidate-ip>
s=-
t=0 0
a=group:BUNDLE 0 1
a=ice-lite
m=audio <port> UDP/TLS/RTP/SAVPF 111
a=candidate:1 1 udp 2130706431 <smg-ip> <port> typ host
...
m=application <port> UDP/DTLS/SCTP webrtc-datachannel
...
```

# Verification:
```bash
curl -X POST http://localhost:8080/v1/realtime/sessions \
  -H "Authorization: Bearer $API_KEY" \
  -d '{"model":"gpt-4o-realtime-preview"}'
```

pallasathena92:
## Description
Fix https://github.com/lightseekorg/smg/issues/249
### Problem
SMG had no support for WebRTC-based realtime sessions. The OpenAI Realtime API supports WebRTC as an alternative transport to WebSocket, where clients exchange SDP offers/answers via HTTP POST and then communicate over data channels and RTP audio. Without WebRTC support, clients that prefer WebRTC (e.g., browser-based apps using the /v1/realtime/calls endpoint) could not use SMG as a gateway.

### Solution

Add a full WebRTC relay implementation where SMG terminates the client's peer connection, establishes its own peer connection to upstream, and bridges data-channel messages plus audio RTP packets between the two. The relay uses str0m (a pure-Rust WebRTC library) for DTLS/ICE/RTP handling and routes requests through the standard RouterTrait → RouterManager → OpenAiRouter pipeline, getting lazy model refresh, routing policy, and metrics for free.

### WebRTC relay core
- **`webrtc_bridge.rs`** (new) — `WebRtcBridge` struct: bidirectional UDP relay between client and upstream peer connections via `str0m`. Handles ICE candidate gathering, DTLS handshake, data channel message forwarding, and RTP audio packet relay. `bridge.run()` returns `bool` for accurate `worker.record_outcome()` reporting (upstream dying while client connected = failure).
- **`webrtc.rs`** (new) — SDP signaling handlers for `POST /v1/realtime/calls`. Supports `application/sdp` (direct) and `multipart/form-data` (with session config). Includes `validate_sdp()` for early 400 on invalid/empty SDP. `BridgeSetupError::UpstreamHttp` preserves original status code and `Content-Type`. Upstream rejections log at `warn!`; internal failures at `error!`. Sanitized SDP logging (candidate count, fingerprint algo, codec names — no raw SDP).

### Router integration
- **`router.rs`** — `OpenAiRouter::route_realtime_webrtc()` implementation using `select_worker_for_model()` for lazy model refresh + routing policy
- **`router_manager.rs`** — Dispatch method for `route_realtime_webrtc()`
- **`mod.rs`** — `route_realtime_webrtc()` added to `RouterTrait` with default stub

### Configuration plumbing
- **`main.rs`** — `--webrtc-bind-addr` and `--webrtc-stun-server` CLI args (STUN default wired via `default_value` so `--help` shows it)
- **`server.rs`** — Passes webrtc params from `ServerConfig` into `AppContextBuilder`
- **`app_context.rs`** — `webrtc_bind_addr` and `webrtc_stun_server` fields on `AppContext` and `AppContextBuilder`
- **`router_args.py`** — Python `RouterArgs` mirrors both params for `python -m smg.launch_router`
- **`lib.rs`** (bindings) — PyO3 `Router` struct accepts and forwards both params to `ServerConfig`

### E2E tests
- **`test_realtime_webrtc.py`** (new) — 6 test classes:
  - `TestRealtimeWebRtcSignaling` — SDP offer/answer (direct + multipart), error handling (missing model, missing auth, unsupported content type, invalid SDP, empty body, wrong HTTP method), concurrent calls
  - `TestRealtimeWebRtcAnswerFormat` — Required SDP fields, ICE credentials, DTLS fingerprint, audio media, content type, multipart vs direct consistency
  - `TestRealtimeWebRtcDataChannel` — session.created, session.update, text response round-trip, conversation.item.created, invalid event → error, response.done schema validation (via `aiortc`)
  - `TestRealtimeWebRtcErrorForwarding` — Invalid API key → 401, invalid model → 4xx (not masked as 502)
  - `TestRealtimeWebRtcBridgeCleanup` — Reconnect after disconnect, multiple sequential sessions
  - `TestRealtimeWebRtcStun` — Signaling and data channel with explicit STUN server
- `httpx.AsyncClient` for SDP POST in async context (avoids blocking aiortc event loop)
- `@pytest.mark.flaky(reruns=2)` on data channel / bridge / STUN classes for cloud latency tolerance
- Session config uses correct API structure (`audio.output.voice` not `voice`)

### CI / infra
- **`Cargo.toml`** — Added `str0m` dependency
- **`ci_install_e2e_deps.sh`** / **`pyproject.toml`** — Added `aiortc`
- **`metrics.rs`** — Added `CONNECTION_WEBRTC` label
- **`pr-test-rust.yml`** — Updated CI workflow

## Notes:
For hangup call "post /v1/realtime/calls/{call_id}/hangup". 
I tried neither `application/sdp` nor `multipart/form-data`. Openai only returns a session_id from OpenAI's /v1/realtime/calls endpoint. I tried to use session_id to handup the call, however session_id is an incorrect identifier. 
```
"{
    "error": {
        "message": "No session found for the provided call_id",
        "type": "invalid_request_error",
        "code": "call_id_not_found",
        "param": ""
    }
}"
```
also tried raw http request with application/json. I got error " OpenAI's /v1/realtime/calls endpoint only accepts application/sdp and returns raw SDP"

For current solution in this code is SMG will generate call_id. And do hangup by just cancel the bridge.
If we want to respone the call_id to client, we can add a header such like X-CALL-ID to client and will handle it in following pr.
**I remove hangup call handler in this pr**

## Test Plan

- [x] Ephemeral direct WebRTC flow
- [x] Unified server-side flow
- [ ] E2e test passed

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * POST /v1/realtime/calls: WebRTC signaling (multipart and raw SDP), SDP answers, data-channel support, ICE/STUN handling, and configurable UDP bind/STUN options via CLI/config.
  * Bidirectional WebRTC relay with lifecycle, load management, and realtime metrics for WebRTC connections.

* **Bug Fixes / Improvements**
  * Improved realtime session cleanup and outcome reporting; more robust proxy outcome handling.

* **Tests**
  * Comprehensive end-to-end WebRTC realtime test suite added; test tooling/deps updated.

### DIFF
--- a/e2e_test/realtime/test_realtime_webrtc.py
+++ b/e2e_test/realtime/test_realtime_webrtc.py
@@ -1,0 +1,846 @@
+"""E2E tests for the Realtime WebRTC relay (/v1/realtime/calls).
+
+Tests the gateway's ability to relay WebRTC sessions via SDP signaling:
+- SDP offer/answer exchange (direct and multipart content types)
+- SDP answer format validation (ICE, DTLS, audio media)
+- Error handling (missing model, invalid SDP, missing auth, wrong method)
+- Concurrency (parallel SDP offers)
+- Hangup call lifecycle
+- Data channel events (session lifecycle, text response, error handling)
+
+Prerequisites:
+- OPENAI_API_KEY environment variable set
+
+Usage:
+    pytest e2e_test/realtime/test_realtime_webrtc.py -v
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import os
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+import httpx
+import pytest
+import requests
+
+logger = logging.getLogger(__name__)
+
+OPENAI_API_KEY = os.environ.get("OPENAI_API_KEY", "")
+REALTIME_MODEL = "gpt-4o-realtime-preview-2024-12-17"
+
+# Minimal SDP offer — enough for the signaling endpoint to accept and return
+# an answer.  We only test the HTTP signaling layer, not actual media flow.
+MINIMAL_SDP_OFFER = (
+    "v=0\r\n"
+    "o=- 0 0 IN IP4 127.0.0.1\r\n"
+    "s=-\r\n"
+    "t=0 0\r\n"
+    "a=group:BUNDLE 0 1\r\n"
+    "a=extmap-allow-mixed\r\n"
+    "a=msid-semantic: WMS\r\n"
+    "m=audio 9 UDP/TLS/RTP/SAVPF 111\r\n"
+    "c=IN IP4 0.0.0.0\r\n"
+    "a=rtcp:9 IN IP4 0.0.0.0\r\n"
+    "a=ice-ufrag:test\r\n"
+    "a=ice-pwd:testpasswordtestpassword\r\n"
+    "a=fingerprint:sha-256 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
+    ":00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00\r\n"
+    "a=setup:actpass\r\n"
+    "a=mid:0\r\n"
+    "a=sendrecv\r\n"
+    "a=rtpmap:111 opus/48000/2\r\n"
+    "m=application 9 UDP/DTLS/SCTP webrtc-datachannel\r\n"
+    "c=IN IP4 0.0.0.0\r\n"
+    "a=ice-ufrag:test\r\n"
+    "a=ice-pwd:testpasswordtestpassword\r\n"
+    "a=fingerprint:sha-256 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
+    ":00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00\r\n"
+    "a=setup:actpass\r\n"
+    "a=mid:1\r\n"
+    "a=sctp-port:5000\r\n"
+    "a=max-message-size:262144\r\n"
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _post_sdp(calls_url, auth_headers, *, model=None, sdp=MINIMAL_SDP_OFFER, timeout=30):
+    """POST an SDP offer to the calls endpoint and return the response."""
+    url = f"{calls_url}?model={model}" if model else calls_url
+    return requests.post(
+        url,
+        data=sdp,
+        headers={**auth_headers, "Content-Type": "application/sdp"},
+        timeout=timeout,
+    )
+
+
+def _post_multipart(calls_url, auth_headers, *, session_config, sdp=MINIMAL_SDP_OFFER, timeout=30):
+    """POST a multipart SDP offer with session config and return the response."""
+    files = {
+        "sdp": ("offer.sdp", sdp, "application/sdp"),
+        "session": ("session.json", json.dumps(session_config), "application/json"),
+    }
+    return requests.post(
+        calls_url,
+        files=files,
+        headers=auth_headers,
+        timeout=timeout,
+    )
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def gateway():
+    """Launch a cloud gateway in OpenAI mode for realtime WebRTC tests."""
+    from infra import launch_cloud_gateway
+
+    if not OPENAI_API_KEY:
+        pytest.skip("OPENAI_API_KEY not set")
+
+    gw = launch_cloud_gateway("openai", history_backend="memory")
+    yield gw
+    gw.shutdown()
+
+
+@pytest.fixture(scope="module")
+def dc_gateway():
+    """Dedicated gateway for data-channel tests.
+
+    Signaling tests leave zombie bridges (fake SDP, no real ICE peer) that
+    consume sockets and poll-loop CPU. A fresh gateway avoids resource
+    starvation that causes DTLS handshake timeouts in CI.
+    """
+    from infra import launch_cloud_gateway
+
+    if not OPENAI_API_KEY:
+        pytest.skip("OPENAI_API_KEY not set")
+
+    gw = launch_cloud_gateway("openai", history_backend="memory")
+    yield gw
+    gw.shutdown()
+
+
+@pytest.fixture()
+def calls_url(gateway):
+    """Build the realtime calls URL."""
+    return f"http://{gateway.host}:{gateway.port}/v1/realtime/calls"
+
+
+@pytest.fixture()
+def auth_headers():
+    """Build the Authorization headers."""
+    return {"Authorization": f"Bearer {OPENAI_API_KEY}"}
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
+@pytest.mark.vendor("openai")
+@pytest.mark.gpu(0)
+class TestRealtimeWebRtcSignaling:
+    """E2E tests for the WebRTC SDP signaling layer."""
+
+    def test_sdp_offer_answer_direct(self, calls_url, auth_headers):
+        """POST application/sdp offer should return an SDP answer with 201."""
+        resp = _post_sdp(calls_url, auth_headers, model=REALTIME_MODEL)
+        assert resp.status_code == 201, (
+            f"Expected 201 CREATED, got {resp.status_code}: {resp.text[:500]}"
+        )
+        content_type = resp.headers.get("Content-Type", "")
+        assert "application/sdp" in content_type, (
+            f"Expected application/sdp Content-Type, got: {content_type}"
+        )
+        # Answer SDP must start with "v=0"
+        answer_sdp = resp.text
+        assert answer_sdp.startswith("v=0"), f"Answer SDP doesn't look valid: {answer_sdp[:200]}"
+        logger.info("Direct SDP: got %d-byte answer SDP", len(answer_sdp))
+
+    def test_sdp_offer_answer_multipart(self, calls_url, auth_headers):
+        """POST multipart/form-data with sdp + session fields should return SDP answer."""
+        resp = _post_multipart(
+            calls_url,
+            auth_headers,
+            session_config={
+                "model": REALTIME_MODEL,
+                "type": "realtime",
+                "audio": {"output": {"voice": "alloy"}},
+            },
+        )
+        assert resp.status_code == 201, (
+            f"Expected 201 CREATED, got {resp.status_code}: {resp.text[:500]}"
+        )
+        answer_sdp = resp.text
+        assert answer_sdp.startswith("v=0"), f"Answer SDP doesn't look valid: {answer_sdp[:200]}"
+        logger.info("Multipart SDP: got %d-byte answer SDP", len(answer_sdp))
+
+    def test_missing_model_returns_error(self, calls_url, auth_headers):
+        """POST without model should return 400."""
+        resp = _post_sdp(calls_url, auth_headers, timeout=10)
+        assert resp.status_code == 400, f"Expected 400 for missing model, got {resp.status_code}"
+
+    def test_missing_auth_returns_unauthorized(self, calls_url):
+        """POST without Authorization header should return 401."""
+        resp = _post_sdp(calls_url, {}, model=REALTIME_MODEL, timeout=10)
+        assert resp.status_code == 401, f"Expected 401 for missing auth, got {resp.status_code}"
+
+    def test_unsupported_content_type_returns_error(self, calls_url, auth_headers):
+        """POST with unsupported Content-Type should return 400."""
+        url = f"{calls_url}?model={REALTIME_MODEL}"
+        resp = requests.post(
+            url,
+            data='{"sdp": "fake"}',
+            headers={**auth_headers, "Content-Type": "application/json"},
+            timeout=10,
+        )
+        assert resp.status_code == 400, (
+            f"Expected 400 for unsupported content type, got {resp.status_code}"
+        )
+
+    def test_invalid_sdp_returns_error(self, calls_url, auth_headers):
+        """POST with invalid SDP body should return 400."""
+        resp = _post_sdp(
+            calls_url,
+            auth_headers,
+            model=REALTIME_MODEL,
+            sdp="this is not valid SDP",
+            timeout=10,
+        )
+        assert resp.status_code == 400, f"Expected 400 for invalid SDP, got {resp.status_code}"
+
+    def test_multipart_missing_sdp_returns_error(self, calls_url, auth_headers):
+        """POST multipart without sdp field should return 400."""
+        # Not using _post_multipart since we intentionally omit the sdp field
+        session_config = json.dumps({"model": REALTIME_MODEL, "type": "realtime"})
+        resp = requests.post(
+            calls_url,
+            files={
+                "session": ("session.json", session_config, "application/json"),
+            },
+            headers=auth_headers,
+            timeout=10,
+        )
+        assert resp.status_code == 400, (
+            f"Expected 400 for missing sdp field, got {resp.status_code}"
+        )
+
+    def test_multipart_missing_model_returns_error(self, calls_url, auth_headers):
+        """POST multipart with session but no model should return 400."""
+        resp = _post_multipart(
+            calls_url,
+            auth_headers,
+            session_config={"audio": {"output": {"voice": "alloy"}}},
+            timeout=10,
+        )
+        assert resp.status_code == 400, (
+            f"Expected 400 for missing model in session, got {resp.status_code}"
+        )
+
+    def test_empty_body_returns_error(self, calls_url, auth_headers):
+        """POST with empty SDP body should return an error."""
+        resp = _post_sdp(calls_url, auth_headers, model=REALTIME_MODEL, sdp="", timeout=10)
+        assert resp.status_code == 400, f"Expected 400 for empty SDP body, got {resp.status_code}"
+
+    def test_get_method_not_allowed(self, calls_url, auth_headers):
+        """GET /v1/realtime/calls should return 405 Method Not Allowed."""
+        url = f"{calls_url}?model={REALTIME_MODEL}"
+        resp = requests.get(url, headers=auth_headers, timeout=10)
+        assert resp.status_code == 405, f"Expected 405 for GET method, got {resp.status_code}"
+
+    def test_put_method_not_allowed(self, calls_url, auth_headers):
+        """PUT /v1/realtime/calls should return 405 Method Not Allowed."""
+        url = f"{calls_url}?model={REALTIME_MODEL}"
+        resp = requests.put(
+            url,
+            data=MINIMAL_SDP_OFFER,
+            headers={**auth_headers, "Content-Type": "application/sdp"},
+            timeout=10,
+        )
+        assert resp.status_code == 405, f"Expected 405 for PUT method, got {resp.status_code}"
+
+    def test_multipart_extra_fields_accepted(self, calls_url, auth_headers):
+        """POST multipart with extra unknown fields should be silently ignored by SMG."""
+        session_config = json.dumps({"model": REALTIME_MODEL, "type": "realtime"})
+        resp = requests.post(
+            calls_url,
+            files={
+                "sdp": ("offer.sdp", MINIMAL_SDP_OFFER, "application/sdp"),
+                "session": ("session.json", session_config, "application/json"),
+                "extra": ("extra.txt", "some extra data", "text/plain"),
+            },
+            headers=auth_headers,
+            timeout=30,
+        )
+        assert resp.status_code == 201, (
+            f"Expected 201 CREATED with extra fields, got {resp.status_code}: {resp.text[:500]}"
+        )
+
+    def test_concurrent_calls(self, calls_url, auth_headers):
+        """Multiple concurrent SDP offers should all succeed."""
+        num_calls = 3
+
+        def _make_call():
+            return _post_sdp(calls_url, auth_headers, model=REALTIME_MODEL)
+
+        with ThreadPoolExecutor(max_workers=num_calls) as pool:
+            futures = [pool.submit(_make_call) for _ in range(num_calls)]
+            results = [f.result() for f in as_completed(futures)]
+
+        for i, resp in enumerate(results):
+            assert resp.status_code == 201, (
+                f"Concurrent call {i} failed: {resp.status_code} {resp.text[:200]}"
+            )
+            assert resp.text.startswith("v=0"), (
+                f"Concurrent call {i} SDP invalid: {resp.text[:200]}"
+            )
+        logger.info("All %d concurrent calls succeeded", num_calls)
+
+
+@pytest.mark.e2e
+@pytest.mark.vendor("openai")
+@pytest.mark.gpu(0)
+class TestRealtimeWebRtcAnswerFormat:
+    """Validate the SDP answer format returned by the signaling endpoint."""
+
+    def test_answer_sdp_format(self, calls_url, auth_headers):
+        """SDP answer must have correct Content-Type, required SDP fields, ICE, DTLS, and audio."""
+        resp = _post_sdp(calls_url, auth_headers, model=REALTIME_MODEL)
+        assert resp.status_code == 201
+
+        # Content-Type
+        ct = resp.headers.get("Content-Type", "")
+        assert "application/sdp" in ct, f"Expected application/sdp, got: {ct}"
+
+        # Required SDP fields
+        answer = resp.text
+        assert "v=0" in answer, "Missing v= line in SDP answer"
+        assert "\no=" in answer or answer.startswith("v=0\r\no="), "Missing o= line"
+        assert "\ns=" in answer or "\r\ns=" in answer, "Missing s= line"
+        assert "\nm=" in answer or "\r\nm=" in answer, "Missing m= (media) line"
+
+        # ICE credentials
+        assert "a=ice-ufrag:" in answer, "Missing ice-ufrag in SDP answer"
+        assert "a=ice-pwd:" in answer, "Missing ice-pwd in SDP answer"
+
+        # DTLS fingerprint
+        assert "a=fingerprint:" in answer, "Missing DTLS fingerprint in SDP answer"
+
+        # Audio media
+        assert "m=audio" in answer, "Missing audio media line in SDP answer"
+        logger.info("SDP answer format validated (%d bytes)", len(answer))
+
+    def test_multipart_answer_matches_direct(self, calls_url, auth_headers):
+        """Multipart and direct SDP flows should return structurally similar answers."""
+        direct_resp = _post_sdp(calls_url, auth_headers, model=REALTIME_MODEL)
+        multi_resp = _post_multipart(
+            calls_url,
+            auth_headers,
+            session_config={"model": REALTIME_MODEL, "type": "realtime"},
+        )
+        assert direct_resp.status_code == 201
+        assert multi_resp.status_code == 201
+
+        # Both must be valid SDP with audio
+        for label, text in [("direct", direct_resp.text), ("multipart", multi_resp.text)]:
+            assert text.startswith("v=0"), f"{label} SDP doesn't start with v=0"
+            assert "m=audio" in text, f"{label} SDP missing audio media"
+            assert "a=ice-ufrag:" in text, f"{label} SDP missing ICE credentials"
+        logger.info("Direct and multipart answers are structurally consistent")
+
+
+# ---------------------------------------------------------------------------
+# Data channel event tests (requires aiortc)
+# ---------------------------------------------------------------------------
+
+RECV_TIMEOUT = 60  # seconds — DTLS handshake can be slow in CI
+
+
+def _parse_event(raw: str | bytes) -> dict | None:
+    """Parse a JSON event from a data channel message."""
+    try:
+        text = raw.decode("utf-8") if isinstance(raw, bytes) else raw
+        return json.loads(text)
+    except (json.JSONDecodeError, UnicodeDecodeError):
+        logger.warning("Non-JSON data channel message: %s", str(raw)[:200])
+        return None
+
+
+async def _recv_dc_event(
+    dc, *, event_type: str | None = None, timeout: float = RECV_TIMEOUT
+) -> dict:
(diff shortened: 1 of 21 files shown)


model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+
+use super::{rest::select_worker, webrtc_bridge::WebRtcBridge};
+use crate::{
+    core::worker::WorkerLoadGuard,
+    routers::{error, header_utils::extract_auth_header},
+    server::AppState,
+};
+
+/// Default STUN server for ICE server-reflexive candidate gathering.
+const DEFAULT_STUN_SERVER: &str = "stun.l.google.com:19302";
```
CatherineSue: **Suggestion**: This should probably be configurable (CLI flag or config file). Enterprise deployments may not allow traffic to external STUN servers, or may need to use their own TURN/STUN infrastructure.

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+    let content_type = headers
+        .get(CONTENT_TYPE)
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("")
+        .to_string();
+
+    if content_type.starts_with("multipart/form-data") {
+        create_call_multipart(state, headers, body, &content_type).await
+    } else if content_type.starts_with("application/sdp") {
+        let model = params.model.as_deref().unwrap_or("");
```
CatherineSue: **Nit**: `params.model.as_deref().unwrap_or("")` followed by `if model.is_empty()` on line 168 (in `create_call_sdp`) — could simplify by matching on `params.model` directly and returning the error for `None`, similar to how `ws.rs` handles it.

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+) -> Response {
+    let Some(worker) = select_worker(state, model) else {
+        error!(model, "No available worker for realtime model");
+        return StatusCode::SERVICE_UNAVAILABLE.into_response();
+    };
+
+    let Some(auth) = extract_auth_header(Some(headers), worker.api_key()) else {
+        return StatusCode::UNAUTHORIZED.into_response();
+    };
+    let auth_str = auth.to_str().unwrap_or("").to_string();
```
CatherineSue: **Nit**: `auth.to_str().unwrap_or("")` silently falls back to an empty string for non-UTF8 auth headers. Compare with `ws.rs:64-74` which returns a 400 error in this case. Would be good to keep behavior consistent across the realtime handlers.

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    let _guard = WorkerLoadGuard::new(worker.clone(), Some(headers));
```
CatherineSue: **Issue**: This `WorkerLoadGuard` is dropped at the end of `setup_and_spawn_bridge()`, but the actual WebRTC call (which could last minutes or hours) continues in the spawned bridge task. The load balancer won't see the active call, so subsequent calls could pile onto the same worker.

Consider moving the guard into the bridge struct or the spawned task so it lives for the call's duration.

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+        session_config,
+        call_id.clone(),
+        &state.context.client,
+        bind_addr,
+        stun_server,
+    )
+    .await
+    {
+        Ok(result) => {
+            worker.record_outcome(true);
```
CatherineSue: **Issue**: `record_outcome(true)` is called when bridge *setup* succeeds, but the actual call hasn't started yet. If the bridge crashes during `run()`, no failure outcome is ever recorded. The circuit breaker will see 100% success regardless of actual call reliability.

Consider recording the outcome when the bridge task completes, similar to how the ws.rs change in this PR now records outcome after the proxy finishes.

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+        .expect("static response builder")
+}
+
+/// `POST /v1/realtime/calls/{call_id}/hangup` — Terminate a WebRTC call.
+///
+/// Cancels the bridge relay task, which disconnects both the client-facing
+/// and upstream peer connections via ICE teardown.  No upstream HTTP
+/// forwarding — OpenAI does not expose a call ID in the direct WebRTC
+/// flow, so the upstream connection is terminated by the ICE disconnect.
+pub async fn hangup_call(
```
CatherineSue: **Suggestion**: The hangup endpoint only checks that the caller passed the route-level auth middleware (any valid API key). It doesn't verify the caller owns the call. Any authenticated user could hang up another user's call if they know the `call_id`. UUIDv7 is time-ordered, which makes it somewhat guessable.

Might be worth storing the caller's identity in the registry entry and checking it here.

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+    /// Audio mid on the *upstream* peer — used to look up a TX stream for
+    /// forwarding client audio.
+    upstream_audio_mid: Option<Mid>,
+    /// Audio mid on the *client* peer — used to look up a TX stream for
+    /// forwarding upstream audio.
+    client_audio_mid: Option<Mid>,
+
+    /// Upstream data-channel messages received before the client channel opens.
+    /// Flushed once `client_channel` becomes `Some`.
+    pending_to_client: Vec<(bool, Vec<u8>)>,
```
CatherineSue, line 129: **Issue**: This buffer grows without bound if the upstream sends data before the client data channel opens (which is the expected race condition). A slow client connection paired with a fast upstream could cause unbounded memory growth.

Consider adding a cap (e.g., 1000 messages) and dropping oldest messages or erroring out if exceeded.

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+    /// hangup cancellation is observed by the relay loop.
+    pub fn set_cancel_token(&mut self, token: CancellationToken) {
+        self.cancel_token = token;
+    }
+
+    /// Run the bidirectional relay until cancelled or disconnected.
+    pub async fn run(mut self, registry: Arc<RealtimeRegistry>) {
+        registry.set_call_state(&self.call_id, ConnectionState::Connected);
+
+        let mut buf_client = vec![0u8; 4096];
```
CatherineSue, line 320: **Nit**: Worth documenting why 4096 was chosen. Standard MTU is ~1500 bytes, but DTLS handshake messages can be larger. Is 4096 sufficient for all expected packet types?

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+        let mut buf_upstream = vec![0u8; 4096];
+
+        loop {
+            // Fixed 50ms poll interval — str0m's poll_output drains events so
+            // we can't peek without consuming; a short fixed interval is fine
+            // for realtime audio latency.
+            let timeout = Duration::from_millis(50);
```
CatherineSue, line 330: **Suggestion**: 50ms fixed poll interval means up to 50ms added latency when both sockets are quiet. For real-time audio with 20ms Opus frames this is probably fine, but worth a comment explaining the tradeoff. Also — could `Rtc::poll_output()` drive a more precise next-timeout instead of a fixed interval?

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+                        len = t.contents.len(),
+                        "{} Rtc → transmit", Self::side_label(side)
+                    );
+                    let _ = self.socket(side).send_to(&t.contents, t.destination).await;
+                }
+                Ok(Output::Event(event)) => match side {
+                    Side::Client => self.handle_client_event(event),
+                    Side::Upstream => self.handle_upstream_event(event),
+                },
+                Ok(Output::Timeout(_)) | Err(_) => break,
```
CatherineSue: **Nit**: The `Err(_)` case breaks silently. Might be worth a `debug!` log for the error — could help diagnose issues in production.

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+}
+
+/// Find the first audio Mid in an Rtc instance (after SDP negotiation).
+///
+/// str0m's `Rtc` only exposes `media(mid)` for individual lookup — there is no
+/// public iterator over all media sections.  SDP mid values are string
+/// representations of small integers ("0", "1", …).  OpenAI realtime sessions
+/// typically have ≤3 media lines (audio + data channel), so probing 0..16
+/// provides generous headroom while keeping the search bounded.
+fn find_audio_mid(rtc: &Rtc) -> Option<Mid> {
```
CatherineSue, line 759: **Suggestion**: Probing string values `"0"` through `"15"` as Mid values is a bit fragile. If str0m changes Mid representation or the upstream uses non-numeric mid values, this would silently return `None` (no audio forwarding, no error). Is there a more robust way to iterate media sections from `Rtc`?

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// STUN binding — minimal client for server-reflexive candidate gathering
+// ---------------------------------------------------------------------------
+
+/// Perform a STUN Binding Request (RFC 5389) on `socket` to discover its
+/// server-reflexive (public) address.  Returns `None` on timeout or parse
+/// failure — callers should proceed with host-only candidates in that case.
+async fn stun_gather_srflx(socket: &UdpSocket, stun_server: SocketAddr) -> Option<SocketAddr> {
```
CatherineSue, line 775: **Issue**: This is a hand-rolled STUN Binding Request client. A few concerns:
- Only handles IPv4 (line 710: `resp[off + 1] == 0x01`). IPv6 environments silently get no server-reflexive candidate.
- No retry for lost STUN packets (single shot, 3s timeout).
- No message integrity/fingerprint validation.

Does str0m provide built-in ICE gathering that could be used instead? If hand-rolling is necessary, this would benefit from thorough unit tests (at minimum: successful IPv4 parse, IPv6 handling, malformed response, transaction ID mismatch).

**Also — no tests for any of the new code** (1040 lines). The STUN parser and bridge state machine in particular would benefit from test coverage.

model_gateway/src/server.rs
```diff
@@ -813,7 +821,15 @@ pub async fn startup(config: ServerConfig) -> Result<(), Box<dyn std::error::Err
     );
 
     let app_context = Arc::new(
-        AppContext::from_config(config.router_config.clone(), config.request_timeout_secs).await?,
+        Box::pin(AppContextBuilder::from_config(
```
CatherineSue, line 908: **Issue**: This changes `startup()` from using `AppContext::from_config()` to manually calling `AppContextBuilder::from_config()...build()`. But `AppContext::from_config()` (app_context.rs:105-118) still exists and doesn't set `webrtc_bind_addr`. This creates two divergent construction paths.

Consider either:
- Adding `webrtc_bind_addr` as a parameter to `AppContext::from_config()` so there's one construction path, or
- Adding a `with_webrtc_bind_addr()` step in `AppContextBuilder::from_config()` so the builder chain handles it

model_gateway/src/routers/openai/realtime/ws.rs
```diff
@@ -101,10 +101,14 @@ pub async fn ws_handler(
...
-        }
+            Ok(()) => true,
+            Err(e) => {
+                error!(session_id, error = %e, "Realtime WebSocket proxy error");
+                false
+            }
+        };
 
-        // Cleanup: remove session on disconnect
+        worker.record_outcome(success);
```
CatherineSue, line 137: Nice fix — recording the outcome after the proxy completes is the right pattern. The WebRTC handler should follow this same approach (record after bridge `run()` completes, not at setup time).

CatherineSue (commented):
Thanks for working on this!

---

### Broader Context: Realtime API Architecture

Before diving into the WebRTC-specific feedback, I want to flag something that applies to the realtime subsystem as a whole (starting from PR #637). The realtime handlers (`ws.rs`, `rest.rs`, and now `webrtc.rs`) have diverged from SMG's standard `server → RouterTrait → handler` pattern. Every other endpoint (`/v1/chat/completions`, `/v1/embeddings`, etc.) follows:

```
server.rs thin handler → state.router.route_*() → OpenAIRouter (RouterTrait impl)
```

The realtime handlers instead access `state.context.worker_registry` directly, which means they currently lack:

1. **Lazy model refresh** — `OpenAIRouter::select_worker_for_model()` calls `refresh_external_models()` when no worker is found initially. The realtime `select_worker()` just returns `None` → 503.
2. **Routing policy** — The configured policy (round-robin, session-affinity, etc.) is ignored; realtime always uses least-load.
3. **Metrics** — No `Metrics::record_router_request()` / `record_router_error()` calls, so realtime traffic is invisible to observability.
4. **WASM middleware** — `protected_routes` have `wasm_middleware` but `realtime_routes` do not.
5. **Provider transforms** — No `ProviderRegistry` integration.
6. **Retry logic** — No `RetryExecutor`.
7. **Request validation is completely bypassed** — The REST handlers (`create_client_secret`, `create_session`, `create_transcription_session`) all accept `Json<Value>` (raw JSON) instead of the typed protocol structs. Compare:

```rust                                                   
// Regular handler (server.rs:184) — typed + validated
ValidatedJson(body): ValidatedJson<ChatCompletionRequest>

// Realtime handler (rest.rs:32) — raw JSON, no validation                                                                                                                                                                                                                                                                                                                          
Json(body): Json<Value>
```        
                                                                                          
The protocols crate defines typed structs for these endpoints (`RealtimeSessionCreateRequest`, `RealtimeTranscriptionSessionCreateRequest`, `RealtimeResponseCreateParams` with `#[derive(Validate)]`), but none of them are used in the handlers. Model extraction is done manually via `body.pointer("/session/model")` and `body.get("model")` instead of accessing typed fields.

This isn't a blocker for this PR specifically, but it's worth noting that the WebRTC handler inherits all of these gaps by building on top of `select_worker()`. We should plan to fix the overall structure first before we add more functionality to the realtime API.

model_gateway/src/routers/openai/realtime/ws.rs
pallasathena92: will do

pallasathena92:
> Thanks for working on this!
> 
> ### Broader Context: Realtime API Architecture
> Before diving into the WebRTC-specific feedback, I want to flag something that applies to the realtime subsystem as a whole (starting from PR #637). The realtime handlers (`ws.rs`, `rest.rs`, and now `webrtc.rs`) have diverged from SMG's standard `server → RouterTrait → handler` pattern. Every other endpoint (`/v1/chat/completions`, `/v1/embeddings`, etc.) follows:
> 
> ```
> server.rs thin handler → state.router.route_*() → OpenAIRouter (RouterTrait impl)
> ```
> 
> The realtime handlers instead access `state.context.worker_registry` directly, which means they currently lack:
> 
> 1. **Lazy model refresh** — `OpenAIRouter::select_worker_for_model()` calls `refresh_external_models()` when no worker is found initially. The realtime `select_worker()` just returns `None` → 503.
> 2. **Routing policy** — The configured policy (round-robin, session-affinity, etc.) is ignored; realtime always uses least-load.
> 3. **Metrics** — No `Metrics::record_router_request()` / `record_router_error()` calls, so realtime traffic is invisible to observability.
> 4. **WASM middleware** — `protected_routes` have `wasm_middleware` but `realtime_routes` do not.
> 5. **Provider transforms** — No `ProviderRegistry` integration.
> 6. **Retry logic** — No `RetryExecutor`.
> 
> This isn't a blocker for this PR specifically, but it's worth noting that the WebRTC handler inherits all of these gaps by building on top of `select_worker()`. We should plan to fix the overall structure first before we add more functionality to the realtime API.

Thanks the suggestion. 
At the beginning of this realtime api work, i checked some vendor have realtime api for now, they have different api format with openai ai, that is why I add it in a seperate handlers rather than SMG's standard `server → RouterTrait → handler` pattern.
I acutal have plan to refactor it as your suggestion later.

pallasathena92:
discussed with @CatherineSue . we want to refactor openai/router.rs. will refactor this pr later

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+
+        // XOR-MAPPED-ADDRESS = 0x0020
+        if attr_type == 0x0020 && attr_len >= 8 && resp[off + 1] == 0x01 {
+            let port = u16::from_be_bytes([resp[off + 2], resp[off + 3]]) ^ 0x2112;
+            let ip = std::net::Ipv4Addr::new(
+                resp[off + 4] ^ 0x21,
+                resp[off + 5] ^ 0x12,
+                resp[off + 6] ^ 0xA4,
+                resp[off + 7] ^ 0x42,
+            );
```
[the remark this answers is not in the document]
pallasathena92: the magic cookie constant already exists 2 lines above at line 732 (0x2112_A442). The XOR bytes here are just its individual octets per RFC 5389 §15.2. Adding another constant for the decomposed bytes adds indirection without much clarity — anyone reading this code already needs to understand the STUN spec.
Add a short comment instead.

model_gateway/src/routers/openai/realtime/webrtc.rs
pallasathena92: add cli flag --webrtc-stun-server. if not set, it falls back to stun.l.google.com:19302 (the default)

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,315 @@
...
+    let auth_str = auth.to_str().unwrap_or("").to_string();
+
+    let _guard = WorkerLoadGuard::new(worker.clone(), Some(headers));
+
+    let upstream_url = format!(
+        "{}/v1/realtime/calls?model={model}",
+        worker.url().trim_end_matches('/')
+    );
+
+    let call_id = uuid::Uuid::now_v7().to_string();
```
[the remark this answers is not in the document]
pallasathena92: The call_id is used in logging for now. I have moved all hangup call related code to another pr. 
For hangup call needs more discussion. The problem is openai didn't response any call identifier.

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
pallasathena92: added comment

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
pallasathena92: added in https://github.com/lightseekorg/smg/pull/687/changes#diff-f83feb385ba95dd8590608bd916fbd392048726f6c222acd70bf778954dcf9bfR522

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+        // A loopback address can't reach external servers, so the upstream
+        // socket falls back to `0.0.0.0` in that case.
+        let upstream_bind = if bind_addr.is_loopback() {
+            IpAddr::V4(std::net::Ipv4Addr::UNSPECIFIED)
+        } else {
+            bind_addr
+        };
+
+        let client_candidate_ip = resolve_candidate_ip(bind_addr)?;
+        let upstream_candidate_ip = resolve_candidate_ip(upstream_bind)?;
```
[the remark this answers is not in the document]
pallasathena92: That's a real limitation but it's pre-existing behavior — not introduced by this PR. Fixing it properly (IPv6 support, candidate override flag) is a meaningful change that deserves its own PR with proper testing.

model_gateway/src/routers/openai/realtime/ws.rs
pallasathena92: added

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
pallasathena92: added test in e2e test

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
pallasathena92: str0m's poll_output() does return timing hints, but using a fixed 50ms simplifies the loop and avoids tight-spinning. For 20ms Opus frames, the sockets are rarely quiet long enough for the 50ms to matter — there's almost always a packet waiting. It's a reasonable tradeoff.
add comment

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+                pkt.seq_no,
+                pkt.header.timestamp,
+                pkt.timestamp,
+                pkt.header.marker,
+                pkt.header.ext_vals.clone(),
+                true,
+                pkt.payload.clone(),
+            );
+        }
+    }
```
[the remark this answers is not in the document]
pallasathena92: str0m's write_rtp takes owned values so zero-copy would require upstream API changes. We can revisit if profiling shows allocation pressure.

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
pallasathena92: str0m doesn't expose a public iterator over media sections. The probe approach is the only option with the current API. The comment on find_audio_mid already documents this limitation ([webrtc_bridge.rs:707-713](vscode-webview://06ud6usisdmpkrn61g0t5414hh7t2o5mgeikb6i92mo4geffem0n/model_gateway/src/routers/openai/realtime/webrtc_bridge.rs#L707-L713)):


/// str0m's `Rtc` only exposes `media(mid)` for individual lookup — there is no
/// public iterator over all media sections.  SDP mid values are string
/// representations of small integers ("0", "1", …).  OpenAI realtime sessions
/// typically have ≤3 media lines (audio + data channel), so probing 0..16
/// provides generous headroom while keeping the search bounded.
OpenAI's realtime API consistently uses numeric mid values, and 0..16 is generous. If str0m adds a public media iterator in the future, we can switch to it.

model_gateway/src/routers/openai/realtime/webrtc_bridge.rs
```diff
@@ -0,0 +1,725 @@
...
+///
+/// When `addr` is unspecified (`0.0.0.0` / `::`), performs a non-sending UDP
+/// "connect" to a public address to let the OS routing table pick the default
+/// outbound interface.  No traffic is sent.
+fn resolve_candidate_ip(addr: IpAddr) -> anyhow::Result<IpAddr> {
+    if !addr.is_unspecified() {
+        return Ok(addr);
+    }
+    let sock = std::net::UdpSocket::bind("0.0.0.0:0")?;
+    sock.connect("8.8.8.8:80")?;
```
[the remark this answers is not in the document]
pallasathena92: Same pre-existing issue as the IPv4/IPv6 in another concern earlier — both stem from resolve_candidate_ip() assuming public IPv4 reachability. Fixing it properly (IPv6 support, candidate override flag) is a meaningful change that deserves its own PR with proper testing.

model_gateway/src/routers/openai/router.rs
```diff
@@ -1127,6 +1132,55 @@ impl crate::routers::RouterTrait for OpenAIRouter {
...
+            metrics_labels::CONNECTION_WEBRTC,
+            &parsed.model,
+            metrics_labels::ENDPOINT_REALTIME,
+            "false",
+        );
+
+        let auth_header = extract_auth_header(Some(&parts.headers), None);
+        let worker = self
+            .select_worker_for_model(&parsed.model, auth_header.as_ref())
+            .await;
```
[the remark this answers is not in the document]
pallasathena92: select_worker_for_model is the same function used by every route in OpenAiRouter (chat, completions, embeddings, realtime REST, realtime WS, and realtime WebRTC). All 8 call sites use the same path (lines 598, 836, 1054, 1075, 1095, 1122, 1163).

So this isn't a WebRTC-specific bug — it's a general OpenAiRouter design choice. Changing it for WebRTC alone would create inconsistency. If you want policy-based routing, it should be a separate PR that changes find_best_worker_for_model for all endpoints.

e2e_test/realtime/test_realtime_webrtc.py
```diff
@@ -0,0 +1,839 @@
...
+        from infra import launch_cloud_gateway
+
+        if not OPENAI_API_KEY:
+            pytest.skip("OPENAI_API_KEY not set")
+
+        gw = launch_cloud_gateway(
+            "openai",
+            history_backend="memory",
+            extra_args=["--webrtc-stun-server", "stun.l.google.com:19302"],
+        )
```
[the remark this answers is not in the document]
pallasathena92: i have stun.l.google.com:19302 at the CLI level, this test is exercising the override path — it explicitly passes the flag and the gateway receives it. To truly test "override works differently from default" you'd need a second working STUN server, which adds infra complexity for little value.

The only safe change would be to use a different real public STUN server (like stun1.l.google.com:19302 or stun.cloudflare.com:3478), but that's fragile and not meaningfully better.

e2e_test/realtime/test_realtime_webrtc.py
```diff
@@ -0,0 +1,839 @@
...
+    """
+    from aiortc import RTCPeerConnection, RTCSessionDescription
+
+    pc = RTCPeerConnection()
+
+    # Create a data channel (matches the "oai-events" channel SMG expects)
+    dc = pc.createDataChannel("oai-events")
+
+    # Add audio transceiver to match expected SDP structure
+    pc.addTransceiver("audio", direction="sendrecv")
```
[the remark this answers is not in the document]
pallasathena92: will add more test case in next one

model_gateway/src/routers/openai/realtime/webrtc.rs
```diff
@@ -0,0 +1,436 @@
...
+        &auth_str,
+        worker,
+        client,
+        bind_addr,
+        stun_server,
+        realtime_registry,
+        label,
+    )
+    .await
+}
```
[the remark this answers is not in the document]
pallasathena92: The record_router_request is already called in router.rs:1152 before handle_realtime_webrtc, and record_router_error is already called for the no-workers case. The missing piece is just success/duration recording after bridge setup — it's a metrics improvement

I wil put it in a follow-up pr

model_gateway/src/server.rs
pallasathena92: i add webrtc_bind_addr and webrtc_stun_server as a parameter into AppContext::from_config()

PR #693, opened 2026-03-10
Add logging request level statistics from engine

scottjlee:
! WIP - Not ready for review !

## Description
WIP, add logging request-level statistics from engine. Create abstraction for general engines, and implement concrete SGLang version.

## Test Plan

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/crates/grpc_client/proto/sglang_scheduler.proto
+++ b/crates/grpc_client/proto/sglang_scheduler.proto
@@ -275,7 +275,6 @@ message RequestStats {
   optional uint32 prompt_tokens = 8;
   optional uint32 completion_tokens = 9;
   optional uint32 cached_tokens = 10;
-  optional string error_message = 11;
 }
 
 message GetRequestStatsResponse {
--- a/model_gateway/src/observability/events.rs
+++ b/model_gateway/src/observability/events.rs
@@ -79,7 +79,6 @@ impl Event for RequestReceivedEvent {
 #[derive(Debug, Clone, Default, PartialEq)]
 pub struct UnifiedRequestStats {
     pub engine: &'static str,
-    pub error_message: Option<String>,
     pub request_received_timestamp_s: Option<f64>,
     pub first_token_generated_timestamp_s: Option<f64>,
     pub request_finished_timestamp_s: Option<f64>,
@@ -98,16 +97,12 @@ impl UnifiedRequestStats {
         request_id: &str,
         model: &str,
         router_backend: &str,
-        http_status_code: Option<u16>,
-        error_message: Option<&str>,
     ) {
         if let Some(ref s) = stats {
             RequestStatsEvent {
                 request_id,
                 model,
                 router_backend,
-                http_status_code,
-                error_message,
                 stats: s,
             }
             .emit();
@@ -117,11 +112,11 @@ impl UnifiedRequestStats {
     /// Merge another stats sample into this one (for multi-sample aggregation).
     /// Uses min for start timestamps, max for end timestamps, sum for
     /// completion_tokens, and first seen value for the remaining fields.
-    /// Skip merging if engines differ.
     pub(crate) fn merge(&mut self, other: &Self) {
-        if self.engine != other.engine {
-            return;
-        }
+        debug_assert_eq!(
+            self.engine, other.engine,
+            "merging stats from different engines is not supported"
+        );
         self.request_received_timestamp_s = opt_min(
             self.request_received_timestamp_s,
             other.request_received_timestamp_s,
@@ -178,21 +173,16 @@ struct RequestStatsEvent<'a> {
     request_id: &'a str,
     model: &'a str,
     router_backend: &'a str,
-    http_status_code: Option<u16>,
-    error_message: Option<&'a str>,
     stats: &'a UnifiedRequestStats,
 }
 
 macro_rules! emit_request_stats {
     ($log_macro:ident, $event:expr, $($prefix:tt)*) => {{
-        let error_message = $event.error_message.or($event.stats.error_message.as_deref());
         $log_macro!(
             $($prefix)*
             request_id = %($event.request_id),
             model = %($event.model),
             router_backend = %($event.router_backend),
-            http_status_code = $event.http_status_code,
-            error_message = error_message,
             engine = %($event.stats.engine),
             request_received_timestamp_s = $event.stats.request_received_timestamp_s,
             first_token_generated_timestamp_s = $event.stats.first_token_generated_timestamp_s,
@@ -256,7 +246,6 @@ mod tests {
             cached_tokens: Some(cached),
             cache_hit_rate: Some(hit_rate),
             spec_decoding_acceptance_rate: Some(spec_rate),
-            ..Default::default()
         }
     }
 
@@ -264,29 +253,17 @@ mod tests {
         use super::*;
 
         #[test]
-        fn different_engines_is_noop() {
+        #[should_panic(expected = "merging stats from different engines is not supported")]
+        fn different_engines_panics_in_debug() {
             let mut a = UnifiedRequestStats {
                 engine: "sglang",
-                request_received_timestamp_s: Some(1.0),
-                prompt_tokens: Some(10),
-                completion_tokens: Some(20),
                 ..Default::default()
             };
-            let original = a.clone();
             let b = UnifiedRequestStats {
                 engine: "vllm",
-                request_received_timestamp_s: Some(0.5),
-                prompt_tokens: Some(5),
-                completion_tokens: Some(30),
                 ..Default::default()
             };
             a.merge(&b);
-            assert_eq!(
-                a.request_received_timestamp_s,
-                original.request_received_timestamp_s
-            );
-            assert_eq!(a.prompt_tokens, original.prompt_tokens);
-            assert_eq!(a.completion_tokens, original.completion_tokens);
         }
 
         #[test]
@@ -344,6 +321,8 @@ mod tests {
         use super::*;
         use crate::observability::metrics::metrics_labels;
 
+        // Test utility to count the number of emitted events.
+
         struct EventCounter(Arc<AtomicUsize>);
 
         impl<S: tracing::Subscriber> tracing_subscriber::Layer<S> for EventCounter {
@@ -371,8 +350,6 @@ mod tests {
                     "req-id",
                     "test-model",
                     metrics_labels::BACKEND_REGULAR,
-                    Some(200),
-                    None,
                 );
             });
             assert_eq!(n, 0);
@@ -384,7 +361,6 @@ mod tests {
                 engine: "sglang",
                 prompt_tokens: Some(10),
                 completion_tokens: Some(20),
-                error_message: Some("engine error".into()),
                 ..Default::default()
             };
             let n = count_events(|| {
@@ -393,8 +369,6 @@ mod tests {
                     "req-id",
                     "test-model",
                     metrics_labels::BACKEND_REGULAR,
-                    Some(500),
-                    Some("caller error"),
                 );
             });
             assert_eq!(n, 1);
@@ -414,8 +388,6 @@ mod tests {
                         &format!("req-id-{i}"),
                         "test-model",
                         metrics_labels::BACKEND_REGULAR,
-                        Some(200),
-                        None,
                     );
                 }
             });
--- a/model_gateway/src/routers/grpc/common/response_collection.rs
+++ b/model_gateway/src/routers/grpc/common/response_collection.rs
@@ -55,8 +55,6 @@ pub(crate) async fn collect_responses(
             // Mark both streams as completed now that both succeeded
             prefill.mark_completed();
             decode_stream.mark_completed();
-
-            // Emit stats from the decode stream (preferred over prefill)
             decode_stream.spawn_stats_emission();
 
             // Merge prefill input_logprobs if requested
--- a/model_gateway/src/routers/grpc/proto_wrapper.rs
+++ b/model_gateway/src/routers/grpc/proto_wrapper.rs
@@ -866,7 +866,6 @@ impl ProtoGenerateComplete {
 fn unified_stats_from_sglang(stats: sglang::RequestStats) -> UnifiedRequestStats {
     UnifiedRequestStats {
         engine: "sglang",
-        error_message: stats.error_message,
         request_received_timestamp_s: stats.request_received_timestamp_s,
         first_token_generated_timestamp_s: stats.first_token_generated_timestamp_s,
         request_finished_timestamp_s: stats.request_finished_timestamp_s,
@@ -893,14 +892,6 @@ impl ProtoGenerateError {
         match self {
             Self::Sglang(e) => &e.message,
             Self::Trtllm(e) => &e.message,
-        }
-    }
-
-    /// Get HTTP status code when provided by backend.
-    pub fn http_status_code(&self) -> Option<u16> {
-        match self {
-            Self::Sglang(e) => e.http_status_code.parse::<u16>().ok(),
-            Self::Trtllm(_) => None,
         }
     }
 }
@@ -977,8 +968,6 @@ impl ProtoStream {
                         &request_id,
                         &model,
                         router_backend,
-                        Some(200),
-                        None,
                     );
                 });
             }
@@ -1179,13 +1168,11 @@ mod tests {
             prompt_tokens: Some(100),
             completion_tokens: Some(200),
             cached_tokens: Some(50),
-            error_message: Some("some error".to_string()),
             ..Default::default()
         };
         let unified = unified_stats_from_sglang(stats);
 
         assert_eq!(unified.engine, "sglang");
-        assert_eq!(unified.error_message.as_deref(), Some("some error"));
         assert_eq!(unified.request_received_timestamp_s, Some(1.0));
         assert_eq!(unified.first_token_generated_timestamp_s, Some(2.0));
         assert_eq!(unified.request_finished_timestamp_s, Some(3.0));
@@ -1205,6 +1192,5 @@ mod tests {
         assert_eq!(partial.prompt_tokens, Some(42u64));
         assert_eq!(partial.completion_tokens, None);
         assert_eq!(partial.first_token_generated_timestamp_s, None);
-        assert_eq!(partial.error_message, None);
     }
 }


crates/grpc_client/proto/sglang_scheduler.proto
```diff
@@ -30,6 +30,9 @@ service SglangScheduler {
   // Get comprehensive load metrics
   rpc GetLoads(GetLoadsRequest) returns (GetLoadsResponse);
 
+  // Get request-level statistics for a completed request
+  rpc GetRequestStats(GetRequestStatsRequest) returns (GetRequestStatsResponse);
```
CatherineSue, line 34: This RPC is defined but never called anywhere in the gateway code. If the plan is to use it later, it shouldn't be in this PR. If it's the intended mechanism for stats collection, then the in-band `RequestStats` variant in the streaming response (line 201) is redundant — you'd be collecting stats two different ways.

Pick one approach: either poll via this RPC after completion, or receive stats in-band. Having both is confusing and creates maintenance burden.

crates/grpc_client/proto/sglang_scheduler.proto
```diff
@@ -195,6 +198,7 @@ message GenerateResponse {
     GenerateStreamChunk chunk = 2;
     GenerateComplete complete = 3;
     GenerateError error = 4;
+    RequestStats request_stats = 5;
```
CatherineSue: Adding `RequestStats` as a variant in the `GenerateResponse` oneof is a fundamental design issue.

The oneof currently has a clean semantic: `chunk | complete | error`. These are *response lifecycle states*. `RequestStats` is metadata *about* the response, not a response itself. Mixing these concerns in the same union means every consumer of the stream now needs to handle a variant that has nothing to do with the generation lifecycle.

The `Complete` message already carries `prompt_tokens`, `completion_tokens`, and `cached_tokens`. For the additional fields (timestamps, cache hit rate, etc.), consider either:
- Extending `GenerateComplete` with optional stats fields, or
- Using the `GetRequestStats` RPC you defined at line 34 (which is currently unused)

crates/grpc_client/proto/sglang_scheduler.proto
```diff
@@ -257,6 +261,28 @@ message GenerateError {
   string details = 3;
 }
 
+message GetRequestStatsRequest {
+  string request_id = 1;
+}
+
+message RequestStats {
```
CatherineSue, line 267: The field numbering mixes `optional` and non-optional inconsistently. `prompt_tokens` (field 8), `completion_tokens` (field 9), and `cached_tokens` (field 10) are non-optional `uint32`, meaning they default to 0 on the wire. But semantically, "0 tokens" and "unknown" are different. If this message is meant to represent stats that may or may not be available from a given engine, all numeric fields should be `optional`.

Also, `response_sent_timestamp_s` (field 5) is defined here but never populated anywhere in the gateway code.

crates/grpc_client/proto/sglang_scheduler.proto
```diff
@@ -372,6 +398,7 @@ message AbortRequest {
 message AbortResponse {
   bool success = 1;
   string message = 2;
+  RequestStats request_stats = 3;
```
CatherineSue: This is the core of the problem with this PR's design: **abort is a lifecycle operation, not a stats-collection mechanism**.

An abort says "stop processing this request." The response should confirm whether the abort succeeded — which it already does with `success` and `message`. Piggybacking stats onto the abort response couples two orthogonal concerns:

1. It forces abort callers to understand and handle stats even when they don't care
2. It means stats are only available if you abort — what about requests that complete naturally?
3. It changes the abort contract from "fire and forget" to "fire and parse stats"

This then cascades into the gateway code where `abort_request()` changes from `Result<(), tonic::Status>` to `Result<proto::AbortResponse, Box<dyn Error>>`, which is a breaking API change across all three backends.

crates/grpc_client/src/sglang_scheduler.rs
```diff
@@ -66,6 +66,17 @@ impl AbortOnDropStream {
         self.aborted.store(true, Ordering::Release);
         debug!("Request {} marked as completed", self.request_id);
     }
+
+    /// Manually abort the request and return the backend abort response.
+    pub async fn abort(
+        &mut self,
+        reason: String,
+    ) -> Result<proto::AbortResponse, Box<dyn std::error::Error + Send + Sync>> {
```
CatherineSue: `Box<dyn std::error::Error + Send + Sync>` as the error type is a step backwards from the existing `tonic::Status`. The original `abort_request` returned `Result<(), tonic::Status>` — callers knew exactly what errors to expect. Now they get an opaque box that could contain anything.

If you need a richer error type, define a proper error enum. Don't erase the type information.

crates/grpc_client/src/sglang_scheduler.rs
```diff
@@ -254,7 +265,7 @@ impl SglangSchedulerClient {
             response.get_ref().success,
             response.get_ref().message
         );
-        Ok(())
+        Ok(response.into_inner())
```
CatherineSue: `response.into_inner()` leaks tonic's `Response<T>` unwrapping into the public API. The previous code correctly used `get_ref()` for inspection and returned `Ok(())`. Now the caller receives a raw proto message — which means proto types leak through the gRPC client abstraction boundary.

The gRPC client module should present a *domain-level* API, not forward raw protos.

crates/grpc_client/src/trtllm_service.rs
```diff
@@ -68,6 +68,18 @@ impl AbortOnDropStream {
         self.aborted.store(true, Ordering::Release);
         debug!("Request {} marked as completed", self.request_id);
     }
+
+    /// Manually abort the request and return the backend abort response.
+    pub async fn abort(
+        &mut self,
+        _reason: String,
```
CatherineSue: `_reason: String` — the underscore prefix tells the reader this parameter is intentionally unused. But it's not unused *by design*; it's unused because TRT-LLM's abort RPC doesn't take a reason parameter.

This is a sign that the `abort()` signature is being forced to match across backends when the backends don't actually share the same contract. A trait with a proper abstraction would handle this better than copy-pasting the same method with suppressed warnings.

CatherineSue (commented):
Hey Scott, appreciate the effort on this — the idea of collecting request-level statistics from engines is genuinely useful and I can see the thought that went into supporting multiple backends. I couldn't help but take a look out of interest, and I noticed quite a few issues with the current design that I think are worth addressing before this goes further, even as a WIP.

I've left detailed line-by-line comments below, but here's the high-level summary:

### Architectural concerns

1. **Stats collection is deeply entangled with abort/lifecycle** — these are orthogonal concerns that shouldn't be coupled
2. **`enable_request_statistics: bool` is threaded through 15+ function signatures** — this creates a massive API surface change for what should be an observability-only concern. The boolean-parameter-threading pattern is a well-known anti-pattern; consider a middleware/interceptor approach instead
3. **The proto changes couple stats into the streaming response union *and* the abort response** — stats are metadata, not a response variant. This breaks the single-responsibility of those messages
4. **~430 lines added to `proto_wrapper.rs`** (already a large file) with three near-identical mapper implementations and four overlapping collection functions

### Code quality concerns

1. Multiple instances of `match { Ok(x) => x, Err(e) => return Err(e) }` which is literally the `?` operator
2. `send_sse_or_abort` takes **8 mutable reference parameters** — a clear sign the function is doing too many things
3. `Box<dyn Error>` used as error type throughout instead of proper error enums
4. `format_optional_f64` allocates `String` to represent `None` as the literal string `"None"` — use tracing's native `Option` support
5. Copy-paste `abort()` methods across three gRPC client backends instead of a shared trait
6. Two identical `CollectedResponses`/`CollectedStreamResponses` structs in different modules
7. No tests for any of the new functionality
8. Unrelated changes mixed in (MCP label constant, format string changes, `finalized_analysis` removal)

### Suggested alternative approach

Rather than threading a boolean through every function, consider:
- A **stats-collecting stream wrapper** that transparently intercepts `ProtoStream` and collects stats — zero changes to existing function signatures
- Keep stats out of the proto response oneof — use the dedicated `GetRequestStats` RPC you already defined, or collect from the `Complete` messages you already receive
- Keep abort as a pure lifecycle operation — don't overload it with stats collection

crates/grpc_client/src/trtllm_service.rs
```diff
@@ -68,6 +68,18 @@ impl AbortOnDropStream {
...
+    ) -> Result<proto::AbortResponse, Box<dyn std::error::Error + Send + Sync>> {
+        self.aborted.store(true, Ordering::Release);
+        self.client
+            .abort_request(self.request_id.clone())
+            .await
+            .map_err(|e| -> Box<dyn std::error::Error + Send + Sync> { Box::new(e) })
```
CatherineSue: `.map_err(|e| -> Box<dyn std::error::Error + Send + Sync> { Box::new(e) })` — this manual boxing is a code smell. The sglang and vllm versions don't need it because their error types happen to auto-convert. This asymmetry across the three backends is a sign that the error type should be unified properly, not papered over with boxing.

model_gateway/src/config/types.rs
```diff
@@ -33,6 +33,8 @@ pub struct RouterConfig {
     pub log_dir: Option<String>,
     pub log_level: Option<String>,
     pub request_id_headers: Option<Vec<String>>,
+    #[serde(default)]
```
CatherineSue: Missing `#[serde(skip_serializing_if = "std::ops::Not::not")]` (or the equivalent for bool). Without it, serialized configs will always include `enable_request_statistics: false`, which is noisy and breaks the convention used by other optional fields in this struct.

Also, the field is placed between `request_id_headers` and `max_concurrent_requests` which are both request-handling fields. Consider grouping it with observability-related config, or at least adding a section comment.

model_gateway/src/observability/events.rs
```diff
@@ -75,6 +75,119 @@ impl Event for RequestReceivedEvent {
     }
 }
 
+/// Backend-specific field names mapped into the unified request stats schema.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct RequestStatsFieldMapping {
```
CatherineSue: `RequestStatsFieldMapping` stores string descriptions of where each field came from (e.g., `"cached_tokens/prompt_tokens"`, `"perf_metrics.arrival_time"`). But these strings are:
1. Never used for anything other than being stored in `UnifiedRequestStats`
2. Never logged, displayed, or queried
3. Never used for actual field mapping — the mapping is done by the `EngineRequestStatsMapper` trait methods

This is dead-weight metadata. If the intent is documentation, use code comments. If the intent is runtime introspection, this should be an enum, not arbitrary strings.

model_gateway/src/observability/events.rs
```diff
@@ -75,6 +75,119 @@ impl Event for RequestReceivedEvent {
...
+                completion_tokens = self.stats.completion_tokens,
+                cached_tokens = self.stats.cached_tokens,
+                "request_stats"
+            );
+        }
+    }
+}
+
+#[inline]
+fn format_optional_f64(value: Option<f64>) -> String {
```
CatherineSue:
```rust
fn format_optional_f64(value: Option<f64>) -> String {
    value.map(|v| v.to_string()).unwrap_or_else(|| "None".to_string())
}
```

This allocates a `String` on every call just to represent `None` as the literal string `"None"`. This function is called 5 times per event emission, on a potentially hot path.

`tracing` already supports `Option<T>` natively via the `valuable` feature or by using `tracing::field::Empty`. Alternatively, use `tracing::field::display(value)` with a wrapper that implements `Display` for `Option<f64>` without allocation.

model_gateway/src/observability/events.rs
```diff
@@ -75,6 +75,119 @@ impl Event for RequestReceivedEvent {
...
+        let cache_hit_rate = format_optional_f64(self.stats.cache_hit_rate);
+        let spec_decoding_acceptance_rate =
+            format_optional_f64(self.stats.spec_decoding_acceptance_rate);
+        let http_status_code = format_optional_u16(self.http_status_code);
+        let error_message = self
+            .error_message
+            .or(self.stats.error_message.as_deref())
+            .unwrap_or("None");
+
+        if is_otel_enabled() {
```
CatherineSue, line 204: The `emit()` implementation duplicates the entire `event!()` / `debug!()` macro call — 15+ fields copy-pasted between the `if is_otel_enabled()` and `else` branches. The only difference is `event!(Level::INFO, ...)` vs `debug!(...)`. This is a maintenance hazard: any new field must be added in both places.

Consider a macro or helper that takes the log level as a parameter.

model_gateway/src/routers/grpc/proto_wrapper.rs
```diff
@@ -484,9 +489,16 @@ pub enum ProtoResponseVariant {
     Chunk(ProtoGenerateStreamChunk),
     Complete(ProtoGenerateComplete),
     Error(ProtoGenerateError),
+    RequestStats(ProtoRequestStats),
     None,
 }
 
+/// Unified request-stats payload from stream responses.
+#[derive(Clone)]
+pub enum ProtoRequestStats {
```
CatherineSue:
```rust
pub enum ProtoRequestStats {
    Sglang(sglang::RequestStats),
}
```

An enum with a single variant. vLLM and TRT-LLM don't produce in-band `RequestStats` messages, so this enum will likely remain single-variant. This is premature abstraction — a type alias or newtype would be simpler and wouldn't require `match` arms that can never be reached.

model_gateway/src/routers/grpc/proto_wrapper.rs
```diff
@@ -861,6 +873,435 @@ impl ProtoGenerateComplete {
...
+    first_token_generated_timestamp_s: Option<f64>,
+    request_finished_timestamp_s: Option<f64>,
+    cache_hit_rate: Option<f64>,
+    spec_decoding_acceptance_rate: Option<f64>,
+    prompt_tokens: u64,
+    completion_tokens: u64,
+    cached_tokens: u64,
+}
+
+trait EngineRequestStatsMapper {
```
CatherineSue: The `EngineRequestStatsMapper` trait and its three implementations (`SglangRequestStatsMapper`, `VllmRequestStatsMapper`, `TrtllmRequestStatsMapper`) span ~140 lines but are nearly identical. The only differences:

- SGLang: matches `ProtoGenerateComplete::Sglang`, no timestamps
- vLLM: matches `ProtoGenerateComplete::Vllm`, no timestamps
- TRT-LLM: matches `ProtoGenerateComplete::Trtllm`, has perf_metrics timestamps

The SGLang and vLLM implementations are *literally identical* except for the enum variant they match. This is a textbook case where you should extract the common logic and only parameterize the differences. A simple `impl ProtoGenerateComplete { fn to_stats_sample(&self) -> Option<EngineRequestStatsSample> }` method would eliminate all three mapper structs.

model_gateway/src/routers/grpc/proto_wrapper.rs
```diff
@@ -861,6 +873,435 @@ impl ProtoGenerateComplete {
...
+                cached_tokens,
+            })
+        } else {
+            None
+        }
+    }
+}
+
+/// Build unified request-level stats from engine-specific complete responses.
+pub fn collect_unified_request_stats(
```
CatherineSue: Four public collection functions with overlapping responsibilities:
- `collect_unified_request_stats` — from completes
- `collect_stream_request_stats` — from stream stats
- `collect_request_stats` — tries stream stats first, falls back to completes
- `collect_sglang_request_stats` — SGLang-specific

This is a confusing API surface. Callers have to know which one to use and when. The logic in `collect_request_stats` ("prefer stream stats, fallback to completes") should be the *only* public function. The rest should be private implementation details.

Also, `aggregate_request_stats` at line 1164 duplicates most of the logic in `collect_sglang_request_stats` — same accumulator pattern, same min/max timestamp logic, same averaging. This is ~80 lines of near-duplicate code.

model_gateway/src/routers/grpc/harmony/processor.rs
```diff
@@ -45,8 +54,19 @@ impl HarmonyResponseProcessor {
...
+            completes: all_responses,
+            request_stats,
+        } = match response_collection::collect_responses(
+            execution_result,
+            request_logprobs,
+            self.enable_request_statistics,
+        )
+        .await
+        {
+            Ok(collected) => collected,
```
CatherineSue:
```rust
match response_collection::collect_responses(...).await {
    Ok(collected) => collected,
    Err(err) => return Err(err),
}
```

This pattern appears twice in this file (lines 67-68 and 233-234) and also in `regular/processor.rs`. It is **literally the `?` operator**. Replace with:

```rust
let collected = response_collection::collect_responses(...).await?;
```

The match-and-return pattern is not only verbose — it's misleading, suggesting there's some transformation happening on the error path when there isn't.

model_gateway/src/routers/grpc/regular/streaming.rs
```diff
@@ -1303,4 +1669,42 @@ impl StreamingProcessor {
         }
         buffer.extend_from_slice(b"\n\n");
     }
+
+    async fn send_sse_or_abort(
```
CatherineSue:
```rust
async fn send_sse_or_abort(
    tx: &UnboundedSender<Result<Bytes, io::Error>>,
    payload: Bytes,
    client_disconnected: &mut bool,
    abort_sent: &mut bool,
    stream: &mut ProtoStream,
    send_error_message: &'static str,
    client_disconnect_error_message: &mut Option<&'static str>,
    stream_request_stats: Option<&mut Vec<ProtoRequestStats>>,
) -> Result<(), String>
```

**8 parameters**, 5 of which are mutable references. This function is doing at least three things:
1. Sending an SSE payload
2. Managing client disconnect state
3. Aborting the backend stream and collecting stats

This should be decomposed. Consider a `ClientConnection` struct that encapsulates `tx`, `client_disconnected`, `abort_sent`, and `client_disconnect_error_message`. Then this becomes `connection.send_or_abort(payload, stream, stats)`.

The function is called **13 times** in this file alone, each time with the same 8-argument boilerplate. That's 13 call sites that will all break if the signature changes.

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -1011,7 +1091,10 @@ impl HarmonyStreamingProcessor {
         if let Some(tool_calls) = accumulated_tool_calls {
             if !tool_calls.is_empty() {
                 let analysis_content = if has_analysis {
-                    finalized_analysis
+                    // Get analysis from finalized parser output by calling finalize again
```
CatherineSue:
```rust
// Get analysis from finalized parser output by calling finalize again
// This is safe because finalize can be called multiple times
let output = parser.finalize(finish_reason.clone());
output.analysis
```

Calling `finalize()` twice because the original `finalized_analysis` variable was removed is a hack. "This is safe because finalize can be called multiple times" — is this actually guaranteed by the parser contract? If it is, it should be documented on the trait. If it isn't, this is a latent bug.

The original code stored `finalized_analysis` for exactly this reason. Why was it removed?

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -694,7 +774,7 @@ impl HarmonyStreamingProcessor {
 
                                 let label = session
                                     .map(|s| s.resolve_tool_server_label(tool_name))
-                                    .unwrap_or_else(|| DEFAULT_SERVER_LABEL.to_string());
+                                    .unwrap_or_else(|| "mcp".to_string());
```
CatherineSue: `.unwrap_or_else(|| "mcp".to_string())` — This replaces `DEFAULT_SERVER_LABEL.to_string()` with a hardcoded string literal. This appears in 3 places in this file (lines 777, 939, 1072).

This is an unrelated change that:
1. Removes the use of a named constant, making future changes error-prone
2. Hardcodes `"mcp"` without explanation of why the constant was wrong
3. Shouldn't be in a PR about request statistics

Unrelated changes make PRs harder to review and should be in a separate commit at minimum.

model_gateway/src/routers/grpc/utils/chat_utils.rs
```diff
@@ -586,6 +596,82 @@ pub(crate) fn parse_json_schema_response(
...
+/// Used by both regular and PD routers for non-streaming requests.
+///
+/// # Arguments
+/// * `stream` - The gRPC response stream to consume
+/// * `worker_name` - Name for logging (e.g., "Prefill", "Decode", "Worker")
+///
+/// # Returns
+/// * `Ok(CollectedStreamResponses)` - Collected complete responses and unified request stats
+/// * `Err(Response)` - Error response if the stream fails or returns an error
+pub(crate) struct CollectedStreamResponses {
```
CatherineSue: `CollectedStreamResponses` — this struct has the exact same fields as `CollectedResponses` in `response_collection.rs` (line 22): a `Vec<ProtoGenerateComplete>` and an `Option<UnifiedRequestStats>`.

Two identical structs in different modules for the same purpose is a clear DRY violation. Use one type.

model_gateway/src/routers/grpc/common/response_collection.rs
```diff
@@ -29,10 +38,12 @@ use crate::routers::{
 pub(crate) async fn collect_responses(
     execution_result: ExecutionResult,
     merge_logprobs: bool,
-) -> Result<Vec<ProtoGenerateComplete>, Response> {
-    let all_responses = match execution_result {
+    enable_request_statistics: bool,
```
CatherineSue: `enable_request_statistics: bool` — this parameter is now threaded through:
- `collect_responses` → `collect_stream_responses` (this file)
- `ResponseProcessor::new`, `HarmonyResponseProcessor::new`, `HarmonyStreamingProcessor::new`
- `StreamingProcessor::new`, `GenerateStreamContext`
- `RequestPipeline::new_regular`, `::new_harmony_single`, `::new_harmony_dual`, `::new_pd`
- `HarmonyResponseProcessingStage::new`
- `GrpcRouter::build`, `GrpcPDRouter::build`

That's 15+ function signatures changed for a single boolean. This is the "boolean parameter threading" anti-pattern. The feature should be implemented at the boundary (e.g., a wrapper around `ProtoStream` that transparently collects stats) rather than plumbed through every layer of the stack.

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -504,11 +584,11 @@ impl HarmonyStreamingProcessor {
         // Phase 1: Drain prefill stream, collecting cached_tokens from Complete messages
         let mut prefill_cached_tokens_by_index: HashMap<u32, u32> = HashMap::new();
         while let Some(result) = prefill_stream.next().await {
-            let response = result.map_err(|e| format!("Prefill stream error: {}", e.message()))?;
+            let response = result.map_err(|e| format!("Prefill stream error: {e}"))?;
             if let ProtoResponseVariant::Complete(complete_wrapper) = response.into_response() {
                 prefill_cached_tokens_by_index
                     .insert(complete_wrapper.index(), complete_wrapper.cached_tokens());
-            }
+        }
```
CatherineSue: **Indentation bug.** The closing `}` here has been dedented one level:

```rust
            if let ProtoResponseVariant::Complete(complete_wrapper) = response.into_response() {
                prefill_cached_tokens_by_index
                    .insert(complete_wrapper.index(), complete_wrapper.cached_tokens());
        }  // <-- this should be at 12-space indent, not 8
        }
```

This makes it look like the `if let` body closes at the `while` level. The code still compiles because both braces are present, but the formatting is broken and misleading. Did `cargo fmt` actually run on this?

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -546,8 +626,8 @@ impl HarmonyStreamingProcessor {
             HashMap::new();
 
         // Metadata from Complete message; seed cached_tokens from prefill phase (dual-stream)
-        let mut finish_reason: String;
-        let mut finalized_analysis: Option<String> = None;
+        let mut finish_reason = String::from("stop");
```
CatherineSue: `let mut finish_reason = String::from("stop");` — pre-initializing `finish_reason` to `"stop"` masks bugs. The original code left it uninitialized (declared later or via `let finish_reason: String;`), which means the compiler would catch any code path that uses it before the Complete message sets it.

By defaulting to `"stop"`, if the Complete message is never received (e.g., stream ends early), you silently report a successful stop instead of surfacing the error. This defeats Rust's ability to catch missing-initialization bugs at compile time.

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -793,12 +874,10 @@ impl HarmonyStreamingProcessor {
...
 
                     // Finalize parser and get complete output
-                    // Responses API: no user-specified stop sequences
                     let final_output = parser.finalize(finish_reason.clone());
 
-                    // Store finalized output for later use
-                    finalized_analysis = final_output.analysis;
-                    accumulated_tool_calls = final_output.commentary;
+                    // Store finalized tool calls and reasoning token count
+                    accumulated_tool_calls.clone_from(&final_output.commentary);
```
CatherineSue: `accumulated_tool_calls.clone_from(&final_output.commentary);` — why `clone_from` instead of `= final_output.commentary`? The `clone_from` method is typically used to reuse allocations when overwriting an existing value, but `accumulated_tool_calls` was `None` before this (it's initialized as `Option<Vec<...>>`). There's no allocation to reuse.

Also, the original code was `accumulated_tool_calls = final_output.commentary;` (a move). This PR changes it to a clone. Why? `final_output` isn't used after this point — the move was correct and zero-cost.

model_gateway/src/routers/grpc/regular/streaming.rs
```diff
@@ -409,11 +484,21 @@ impl StreamingProcessor {
...
+                            &mut grpc_stream,
+                            "Failed to send content chunk",
+                            &mut client_disconnect_error_message,
+                            Some(&mut stream_request_stats),
+                        )
+                        .await?;
                     }
                 }
                 ProtoResponseVariant::Complete(complete) => {
+                    completed_responses.push(complete.clone());
```
CatherineSue: `completed_responses.push(complete.clone());` — you're cloning the *entire* `ProtoGenerateComplete` proto message (which contains output text, logprobs, etc.) on the hot path, purely for stats collection.

This clone happens at every `Complete` event. For requests with `n > 1`, you clone multiple times. The stats you extract from it are just a few integers (prompt_tokens, completion_tokens, cached_tokens). Extract the stats eagerly into a lightweight struct instead of cloning the full proto.

Same issue in `harmony/streaming.rs:303`.

model_gateway/src/routers/grpc/regular/streaming.rs
```diff
@@ -1035,6 +1358,49 @@ impl StreamingProcessor {
         });
     }
 
+    async fn emit_generate_request_stats(
```
CatherineSue: `async fn emit_generate_request_stats(...)` — this function is marked `async` but its body is entirely synchronous. It calls `collect_request_stats()` (sync) and `emit()` (sync). No `.await` appears in the function body.

An unnecessary `async` forces all callers to `.await` it and prevents calling from synchronous contexts. Remove the `async`.

model_gateway/src/routers/grpc/regular/streaming.rs
```diff
@@ -1035,6 +1358,49 @@ impl StreamingProcessor {
...
+                &ctx.model,
+                ctx.backend_type,
+                http_status_code,
+                error_message,
+                &request_stats,
+            );
+        }
+    }
+
+    fn emit_request_stats_event(
```
CatherineSue: `emit_request_stats_event` is a separate function that does nothing but construct a `RequestStatsEvent` and call `.emit()`. This is 14 lines of code that add zero logic — it's pure boilerplate indirection.

`emit_generate_request_stats` (the only caller) could just construct and emit the event directly. Two levels of function calls for "create struct, call method" is over-abstraction.

model_gateway/src/routers/grpc/regular/streaming.rs
```diff
@@ -453,8 +547,37 @@ impl StreamingProcessor {
                     // Don't break - continue reading all Complete messages for n>1
                 }
                 ProtoResponseVariant::Error(error) => {
+                    if self.enable_request_statistics {
```
CatherineSue: This 20-line error stats emission block (lines 550–575) is copy-pasted **3 times** in this file alone:
1. Here (stream error in chat completions)
2. In `process_generate_streaming` (~line 907)
3. In `process_generate_logprob_streaming` (~line 1145)

The same pattern also appears in `harmony/streaming.rs`. That's 5+ copies of the same accumulate-and-emit logic. Extract it into a method.

model_gateway/src/routers/grpc/proto_wrapper.rs
```diff
@@ -861,6 +873,435 @@ impl ProtoGenerateComplete {
...
+fn max_timestamp(current: Option<f64>, candidate: Option<f64>) -> Option<f64> {
+    match (current, candidate) {
+        (Some(curr), Some(next)) => Some(curr.max(next)),
+        (None, Some(next)) => Some(next),
+        (some_curr, None) => some_curr,
+    }
+}
+
+#[inline]
+fn normalize_trtllm_timestamps(
```
CatherineSue: `normalize_trtllm_timestamps` uses a completely undocumented heuristic:

```rust
if first_val >= arrival_ts {
    Some(first_val)          // treat as absolute
} else {
    Some(arrival_ts + first_val)  // treat as relative offset
}
```

So if `first_token_time < arrival_time`, it assumes the value is a *relative offset* rather than an absolute timestamp, and adds it to arrival.

What if the clock just drifted? What if the value is in different units? This heuristic will silently produce wrong timestamps. At minimum, this needs:
1. A doc comment explaining *why* TRT-LLM sometimes returns relative vs absolute
2. A log warning when the heuristic triggers
3. A unit test proving the heuristic is correct for known TRT-LLM output

model_gateway/src/routers/grpc/proto_wrapper.rs
```diff
@@ -913,6 +1362,35 @@ impl ProtoStream {
             Self::Trtllm(stream) => stream.mark_completed(),
         }
     }
+
+    /// Abort the underlying request stream.
+    pub async fn abort(&mut self, reason: &str) -> Result<(), String> {
```
CatherineSue:
```rust
pub async fn abort(&mut self, reason: &str) -> Result<(), String> {
    self.abort_with_request_stats(reason).await.map(|_| ())
}
```

Two methods with nearly identical signatures where one wraps the other and discards data. This is a leaky abstraction — callers now have to decide between `abort()` and `abort_with_request_stats()`, and the only difference is whether they get stats back.

Just have one method that returns the stats. Callers who don't need stats simply ignore the return value. This is what `let _ =` is for.

model_gateway/src/main.rs
```diff
@@ -302,6 +302,10 @@ struct CliArgs {
     #[arg(long, default_value_t = 1800, help_heading = "Request Handling")]
     request_timeout_secs: u64,
 
+    /// Enable request-level statistics collection and logging
+    #[arg(long, default_value_t = false, help_heading = "Request Handling")]
+    enable_request_statistics: bool,
```
CatherineSue: `#[arg(long, default_value_t = false)]` — `default_value_t = false` is redundant. `bool` already defaults to `false` in clap. Other bool flags in this file (e.g., `disable_frontend`) don't specify `default_value_t = false` explicitly.

Also, this is under `help_heading = "Request Handling"` but request statistics is an observability concern, not request handling. Should be under an observability heading if one exists.

model_gateway/src/routers/grpc/pipeline.rs
```diff
@@ -63,12 +63,14 @@ impl RequestPipeline {
         reasoning_parser_factory: ReasoningParserFactory,
         configured_tool_parser: Option<String>,
         configured_reasoning_parser: Option<String>,
+        enable_request_statistics: bool,
```
CatherineSue: `enable_request_statistics: bool` — this parameter now appears in **four** pipeline constructors in this file:
- `new_regular` (line 66)
- `new_harmony_single` (line 113)
- `new_harmony_dual` (line 146)
- `new_pd` (line 178)

Each one passes it to `ResponseProcessor::new()` and `StreamingProcessor::new()`. This is the concrete manifestation of the threading problem — a single config boolean ripples through 10 constructor calls in one file alone.

Consider extracting a `PipelineConfig` struct or reading from a shared context rather than passing through every constructor.

model_gateway/src/routers/grpc/regular/processor.rs
```diff
@@ -35,6 +41,7 @@ pub(crate) struct ResponseProcessor {
     pub reasoning_parser_factory: ReasoningParserFactory,
     pub configured_tool_parser: Option<String>,
     pub configured_reasoning_parser: Option<String>,
+    pub enable_request_statistics: bool,
```
CatherineSue: `pub enable_request_statistics: bool` — config booleans should not be `pub` fields on a processor struct. This lets anyone mutate the stats-collection behavior at runtime. Make it private, or better yet, don't store it here — read it from the router context at the point where stats are actually emitted.

model_gateway/src/routers/grpc/regular/processor.rs
```diff
@@ -200,8 +209,19 @@ impl ResponseProcessor {
...
+            completes: all_responses,
+            request_stats,
+        } = match response_collection::collect_responses(
+            execution_result,
+            request_logprobs,
+            self.enable_request_statistics,
+        )
+        .await
+        {
+            Ok(collected) => collected,
```
CatherineSue: Same `match { Ok(collected) => collected, Err(err) => return Err(err) }` pattern that's literally `?`. This instance *and* the one at line 394. That's 4 total occurrences across two processor files (here and harmony) of a 5-line block that should be `?`.

crates/grpc_client/src/vllm_engine.rs
```diff
@@ -66,6 +66,17 @@ impl AbortOnDropStream {
         self.aborted.store(true, Ordering::Release);
         debug!("Request {} marked as completed", self.request_id);
     }
+
+    /// Manually abort the request and return the backend abort response.
+    pub async fn abort(
```
CatherineSue: Third copy-paste of the same `abort()` method. Three backends, three identical method signatures, three implementations that all do `self.aborted.store(true, ...); self.client.abort_request(...)`. This should be a trait method or at minimum a shared helper.

Also note: vLLM's `abort_request` takes different parameters than SGLang's (no `reason`), yet the shared signature forces it to accept `reason: String`. This mismatch is papered over by different `_reason` patterns across backends.

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -504,11 +584,11 @@ impl HarmonyStreamingProcessor {
         // Phase 1: Drain prefill stream, collecting cached_tokens from Complete messages
         let mut prefill_cached_tokens_by_index: HashMap<u32, u32> = HashMap::new();
         while let Some(result) = prefill_stream.next().await {
-            let response = result.map_err(|e| format!("Prefill stream error: {}", e.message()))?;
+            let response = result.map_err(|e| format!("Prefill stream error: {e}"))?;
```
CatherineSue: `format!("Prefill stream error: {e}")` — this was previously `format!("Prefill stream error: {}", e.message())`. This is not just a style change.

`e` here is a `tonic::Status`. `Display` for `tonic::Status` includes the status code (e.g., `"status: Internal, message: ..."`), while `.message()` returns just the message string. This changes the error string format, which could break log parsers or error matching.

This is an unrelated semantic change buried in a stats PR.

model_gateway/src/routers/grpc/harmony/streaming.rs
```diff
@@ -268,6 +300,7 @@ impl HarmonyStreamingProcessor {
                     }
                 }
                 ProtoResponseVariant::Complete(complete_wrapper) => {
+                    completed_responses.push(complete_wrapper.clone());
```
CatherineSue: `completed_responses.push(complete_wrapper.clone());` — same unnecessary full-proto clone as in `regular/streaming.rs:501`. This is on the streaming hot path inside a `while let Some(response)` loop.

model_gateway/src/routers/grpc/proto_wrapper.rs
```diff
@@ -861,6 +873,435 @@ impl ProtoGenerateComplete {
...
+    let mut cache_hit_rate_count = 0u64;
+    let mut spec_acceptance_rate_sum = 0.0f64;
+    let mut spec_acceptance_rate_count = 0u64;
+    let mut prompt_tokens = 0u64;
+    let mut completion_tokens = 0u64;
+    let mut cached_tokens = 0u64;
+    let mut error_message: Option<String> = None;
+
+    for sample in stats {
+        seen += 1;
```
CatherineSue:
```rust
let mut seen = 0u64;
// ...
for sample in stats {
    seen += 1;
    // ...
}
if seen == 0 { return None; }
```

The `seen` counter is unnecessary. You can just check `stats.is_empty()` before the loop, or use the fact that `stats.first()?` already returns `None` for empty slices (which you do at the top of the calling function anyway). This manual counter adds complexity for no reason.

CatherineSue (commented):
Round 2 — more line-by-line issues I caught on closer inspection.

model_gateway/src/routers/grpc/harmony/processor.rs
```diff
@@ -7,32 +7,41 @@ use openai_protocol::{
...
     },
 };
 
 /// Processor for non-streaming Harmony responses
 ///
 /// Collects all output tokens from execution and parses them using
 /// HarmonyParserAdapter to extract the complete response.
-pub(crate) struct HarmonyResponseProcessor;
+pub(crate) struct HarmonyResponseProcessor {
+    enable_request_statistics: bool,
```
CatherineSue: `HarmonyResponseProcessor` was a unit struct (`struct HarmonyResponseProcessor;`). Now it carries `enable_request_statistics: bool`.

This means a struct whose job is *response processing* now carries *configuration about whether to collect metrics*. This violates single responsibility — the processor should process responses, and something else should decide whether stats get emitted.

This is why the middleware/wrapper approach would be cleaner: the processor stays focused on its job, and the stats layer wraps around it.

model_gateway/src/routers/grpc/harmony/stages/response_processing.rs
```diff
@@ -29,17 +29,17 @@ pub(crate) struct HarmonyResponseProcessingStage {
...
     /// Create a new Harmony response processing stage
-    pub fn new() -> Self {
+    pub fn new(enable_request_statistics: bool) -> Self {
         Self {
-            processor: HarmonyResponseProcessor::new(),
-            streaming_processor: Arc::new(HarmonyStreamingProcessor::new()),
+            processor: HarmonyResponseProcessor::new(enable_request_statistics),
+            streaming_processor: Arc::new(HarmonyStreamingProcessor::new(enable_request_statistics)),
         }
     }
```
CatherineSue, line 37:
```rust
streaming_processor: Arc::new(HarmonyStreamingProcessor::new(enable_request_statistics)),
```

This line is over 80 chars (rustfmt would catch this) and shows the config boolean being passed into an `Arc`-wrapped processor. The `enable_request_statistics` value is now baked into an `Arc` that lives for the lifetime of the stage. If you ever wanted to toggle stats at runtime (e.g., via a config reload), this architecture makes it impossible.

scottjlee:
CI fails because the branch is on my fork, so API keys used in the CI tests are unavailable. Close in favor of https://github.com/lightseekorg/smg/pull/757 which should fix this issue.

PR #723, opened 2026-03-11
feat(interactions): Implement steps to handle non-stream interactions req without tool call

XinyueZhang369:
### Problem

 The Gemini router was registered in PR #679 with stub step implementations (returning 501). Users cannot actually route non-streaming requests through the Gemini Interactions API endpoint. 

### Solution

Implement the Phase 1 (non-streaming, no MCP tool) request path for the Gemini router. This covers the four state machine steps: worker selection, request building, upstream execution, and response processing that were defined in PR #417 
## Changes

- Worker selection: Wildcard mode — picks the least-loaded healthy external worker with an open circuit breaker, regardless of model. The upstream Gemini API validates model availability.
- Request building: Serializes InteractionsRequest as the upstream payload with model override from URL path and store: false for model requests. Rejects store: true model requests (persistence not yet implemented). Agent
  requests pass store through unchanged.
- Non-stream execution: POSTs payload to upstream /v1beta/interactions with circuit breaker recording on success/failure, Gemini auth header forwarding, and error sanitization. Trips circuit breaker only on 5xx, not 4xx.
- Response processing: Patches response metadata (model/agent, store, previous_interaction_id) and returns upstream response to client. Interaction ID generation deferred to Phase 3 (storage).
- Utility module: extract_gemini_auth_header extracts x-goog-api-key from client headers, falling back to worker-configured API key.
- Header utils: Added Gemini provider case to apply_provider_headers for x-goog-api-key header on googleapis.com URLs.
- Context: Removed #[expect(dead_code)] from structs now in use.
- Mock server: MockGeminiServer serving /v1beta/interactions and /v1beta/models with optional auth validation. Matches real Gemini API behavior (no id when store=false).
- Unit tests: 13 tests covering successful model/agent requests, store=true rejection, wildcard model forwarding, missing model/agent validation, previous_interaction_id rejection, auth header forwarding, wrong key
  rejection, multiple workers, model ID override, empty input validation, model+agent conflict, and full Axum app round-trip.

## Test Plan

- Create unit tests with a mock Gemini server to validate
- Manual e2e test to ensure the gemini router handles non-stream request without tool call successfully, and returns valid response. Will add e2e tests for real Gemini backend once we setup oauth
SMG started and handled request successfully
Screenshot 2026-03-10 at 7 17 49 PM
Interactions request to SMG succeeded and returned correct response comparing to response from Gemini endpoint call
Screenshot 2026-03-10 at 8 24 59 PM
Screenshot 2026-03-10 at 9 42 27 PM

Response from SMG
```
{
  "status": "completed",
  "outputs": [
    {
      "signature": "<XXXX>",
      "type": "thought"
    },
    {
      "text": "The capital of France is **Paris**.",
      "type": "text"
    }
  ],
  "usage": {
    "total_tokens": 41,
    "total_input_tokens": 8,
    "input_tokens_by_modality": [
      {
        "modality": "text",
        "tokens": 8
      }
    ],
    "total_cached_tokens": 0,
    "total_output_tokens": 8,
    "total_tool_use_tokens": 0,
    "total_thought_tokens": 25
  },
  "role": "model",
  "created": "2026-03-11T03:22:09Z",
  "updated": "2026-03-11T03:22:09Z",
  "object": "interaction",
  "model": "gemini-2.5-flash"
}
```
Response from direct gemini endpoint call
```
{
  "status": "completed",
  "outputs": [
    {
      "signature": "<XXX>",
      "type": "thought"
    },
    {
      "text": "Hello! How can I help you today?",
      "type": "text"
    }
  ],
  "usage": {
    "total_tokens": 620,
    "total_input_tokens": 2,
    "input_tokens_by_modality": [
      {
        "modality": "text",
        "tokens": 2
      }
    ],
    "total_cached_tokens": 0,
    "total_output_tokens": 9,
    "total_tool_use_tokens": 0,
    "total_thought_tokens": 609
  },
  "role": "model",
  "created": "2026-03-11T03:48:09Z",
  "updated": "2026-03-11T03:48:09Z",
  "object": "interaction",
  "model": "gemini-2.5-flash"
}
```
<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Implemented a retry mechanism for request handling.
  * Added provider-aware authentication header extraction and application.

* **Bug Fixes**
  * Enhanced error handling and circuit-breaker updates for upstream request failures.
  * Improved response metadata patching to ensure correct payload structure and store behavior.

* **Chores**
  * Updated upstream API endpoint to v1beta and added Phase 2 planning notes.


model_gateway/src/routers/gemini/steps/response_processing.rs
```diff
@@ -7,43 +7,67 @@ use axum::{
...
+    if let Some(agent) = &req.agent {
+        obj.insert("agent".to_string(), Value::String(agent.clone()));
+    } else if let Some(model) = ctx.input.model_id.as_deref().or(req.model.as_deref()) {
+        obj.insert("model".to_string(), Value::String(model.to_string()));
+    }
+
+    // Always set store=false for model requests in request building step, so set the original value back in response
+    if req.agent.is_none() && req.store {
+        obj.insert("store".to_string(), Value::Bool(req.store));
+    }
```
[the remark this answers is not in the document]
XinyueZhang369: This PR already restore the store value in line 53 of response_processing.rs, so don't have this comment

model_gateway/tests/api/interactions_api_test.rs
```diff
@@ -0,0 +1,483 @@
...
+            .route_interactions(None, &request, request.model.as_deref())
+            .await;
+
+        assert_eq!(
+            response.status(),
+            StatusCode::OK,
+            "request {i} should succeed"
+        );
+    }
+}
```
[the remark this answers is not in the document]
XinyueZhang369: Load distribution is tested in tests/routing/load_balancing_test.rs — this test verifies the Gemini router works with multiple backends. As we only implement a very basic request handling for gemini router, we can add this later when more implementation is done

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
```diff
@@ -1,69 +1,117 @@
...
+            .timeout(ctx.components.request_timeout),
+        upstream_url,
+        auth_header.as_ref(),
+    );
+
+    // Send the request
+    let response = match request_builder.send().await {
+        Ok(r) => r,
+        Err(e) => {
+            worker.circuit_breaker().record_failure();
```
CatherineSue: Circuit breaker should not record failure on all failed cases. For instance 400, or invalid argument doesn't count

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
```diff
@@ -1,69 +1,117 @@
...
+            tracing::error!(
+                url = %upstream_url,
+                error = %e,
+                "Failed to forward request to upstream worker"
+            );
+            return Err(error::bad_gateway(
+                "upstream_error",
+                format!("Failed to forward request to upstream worker: {e}"),
+            ));
```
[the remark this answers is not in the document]
CatherineSue: I also don't think this means bad_gateway in general?

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
```diff
@@ -1,69 +1,117 @@
...
-    //       If exceeded, return Err(max_tool_calls error).
-    //    e. Build resume payload: original input + conversation history + function results.
-    //    f. Set ctx.processing.payload = Some(resume_payload).
-    //    g. Set ctx.state = NonStreamRequest (loop back).
-    //    h. Return Ok(StepResult::Continue).
-    // 10. If no MCP tool calls:
-    //    a. Set ctx.state = ProcessResponse.
-    //    b. Return Ok(StepResult::Continue).
+    let payload = ctx
+        .processing
```
CatherineSue, line 33: Is processing basically request_builder.rs?

model_gateway/src/routers/gemini/steps/request_building.rs
```diff
@@ -2,70 +2,60 @@
...
-    // MCP tool rewriting will modify this payload in a later phase.
-    let payload = serde_json::to_value(ctx.input.original_request.as_ref()).map_err(|e| {
-        (
-            StatusCode::INTERNAL_SERVER_ERROR,
-            format!("failed to serialize request: {e}"),
+    let mut payload = serde_json::to_value(ctx.input.original_request.as_ref()).map_err(|e| {
+        error::bad_request(
+            "invalid_request",
+            format!("Failed to serialize request: {e}"),
         )
```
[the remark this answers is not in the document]
CatherineSue: This comment has a point.

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
XinyueZhang369: Not quite sure about the question, the `ctx.processing.payload` is populated in request_building.rs step, does this answer your question?

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
XinyueZhang369: Yep, I record failure for server error only at the next few lines when checking response status. For the upstream error here, it doesn't include 4xx errors I think, do you want me to just remove it? But I think maybe better to record for upstream error?

model_gateway/tests/api/interactions_api_test.rs
```diff
@@ -0,0 +1,418 @@
...
+    headers.insert("x-goog-api-key", "test-key-123".parse().unwrap());
+
+    let response = router
+        .route_interactions(Some(&headers), &request, request.model.as_deref())
+        .await;
+
+    assert_eq!(response.status(), StatusCode::OK);
+    let body = json_body(response).await;
+    assert_eq!(body["status"], "completed");
+}
```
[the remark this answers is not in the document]
XinyueZhang369: This will be covered in e2e tests against real gemini backend that will be added in follow up PR

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
```diff
@@ -1,69 +1,131 @@
...
+    let auth_header = extract_gemini_auth_header(ctx.input.headers.as_ref(), worker.api_key());
+    let request_builder = apply_provider_headers(
+        ctx.components
+            .client
+            .post(upstream_url)
+            .json(&payload)
+            .timeout(ctx.components.request_timeout),
+        upstream_url,
+        auth_header.as_ref(),
+    );
```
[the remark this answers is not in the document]
XinyueZhang369: Fixed at extract_gemini_auth_header, wrap worker api key with `Bearer` and don't strip `Bearer` prefix

XinyueZhang369:
@coderabbitai review

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
CatherineSue: I think the logic used here is not very straightforward. Do we have clear principles on how we deal with proxy workers?
Because in theory, circuit breaker should be used with retry executor. Here we are just simply bypassing the retry executor. 
Having this kind of diverged behavior in the details is not very maintainable. If we decide external workers can be treated differently, this logic should be better extracted into worker instead of here. 

Loop @slin1237

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
CatherineSue: Yes.

model_gateway/src/routers/header_utils.rs
```diff
@@ -174,7 +174,24 @@ pub fn apply_provider_headers(
                 }
             }
         }
-        ApiProvider::Gemini | ApiProvider::Xai | ApiProvider::OpenAi | ApiProvider::Generic => {
+        ApiProvider::Gemini => {
+            // Gemini uses x-goog-api-key for authentication
+            if let Some(auth) = auth_header {
```
CatherineSue: Can we have this kind of abstraction in `ApiProvider` instead? 

Is it true that other providers use the same auth as OpenAI?

cc @slin1237 @key4ng 

This can be done in a follow-up PR. I'd prefer to sort this one out here. As I keep seeing auth_header extractions everywhere.

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
CatherineSue: I think it is fine for now. 
But since you are using state machine, in the end you will need to use `RetryExecutor`. 
Another thing is, I saw only `record_failure` but not `record_success`. We have `worker.record_outcome`.

In general I'm not a fan of fine grained APIs like this. It causes differences and raises questions on how to use them. For instance, I'm only seeing `worker.circuit_breaker().record*` used in gemini and openai routers

CatherineSue: Third thing, calling `circuit_breaker().record_success` and failure will skip the metrics recording.

Same with current use case in openai router. @key4ng Please see if you can follow up on this one. 

I think it is best to make `record_success` and `record_failure` a private function.

Last, I'd say always have concerns when you call a function route like this. In general, if worker is a good API, we should not need to call something like `worker.a.b()`.

CatherineSue: Did some more investigation, I think `worker.record_outcome` should filter the status that doesn't count for circuit_breaker, such as 400.

Then we replace all the calls here and openai router. We can also eliminate the `is_cb_failure` in grpc router. 

cc @slin1237

CatherineSue (commented):
Overall LGTM. I only have two architecture concerns. I don't think it is a good idea to make them the scope of this PR.

Can we remove all the UTs though? This PR has a lot of UTs but with proxy router or any router, the integration tests are always better. And it is hard to review UTs.

model_gateway/src/routers/gemini/steps/non_stream_execution.rs
XinyueZhang369: The record_success is at the following lines, not sure the line number before my latest commit when you reviewing this file, it's right after parsing response body. 
Though other external worker router didn't use retry executor, but I think it's a better practice, so I added in my new commit. It will be added in router.rs, so we still need to record success/failure here for upstream error and server error. And yes we should use `worker.record_outcome`, I've made changed to use it in gemini router. Do we want modify openai router as well in this PR? Or maybe open another PR for it? Just to limit the size of each PR

model_gateway/src/routers/header_utils.rs
XinyueZhang369: Sure, I moved extract header and apply header to `ApiProvider`

model_gateway/src/routers/gemini/router.rs
```diff
@@ -72,13 +74,31 @@ impl RouterTrait for GeminiRouter {
...
+            },
+            |res, _attempt| is_retryable_status(res.status()),
+            |_delay, _attempt| {
+                // TODO: record retry metrics when Gemini metrics are added
+            },
+            || {
+                // TODO: record retries-exhausted metric when Gemini metrics are added
+            },
+        )
+        .await
```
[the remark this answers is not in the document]
XinyueZhang369: @CatherineSue what's your opinion about this? I saw grpc router retry the while pipeline with retry executor, is it better that we do the same for interactions request or we just retry sending the request to external worker?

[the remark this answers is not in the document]
XinyueZhang369: Is this the reason we don't retry for external workers?

PR #742, opened 2026-03-12
feat(grpc-servicer): support upstream vLLM KV event subscriptions

smfirmin:
### Summary

This PR adds SMG support for consuming vLLM KV cache events over SubscribeKvEvents through smg-grpc-servicer, with no source changes required in the vllm repo.

It also hardens the integration around two important edge cases:

  - backend restart/reconnect behavior
  - unsupported/null-block BlockStored layouts from vLLM

Finally, it documents the remaining protocol limitation and stabilizes one unrelated flaky routing test fixture.

### What changed

  - Added vLLM KV event bridge support in smg-grpc-servicer so SMG can subscribe to KV cache events from upstream vLLM when vLLM is started with --kv-events-config.
 - Added fail-closed handling for unsupported BlockStored layouts that cannot be translated losslessly.
 - Mapped those unsupported layouts to gRPC FAILED_PRECONDITION instead of allowing corrupted KV state.
 - Updated the Rust KV event monitor to:
   - detect replayed stale batches correctly
   - detect a regressed first fresh batch after reconnect as a backend restart
   - clear worker KV state and reconnect from sequence 0 on detected restart
   - stop retrying on terminal backend responses such as UNIMPLEMENTED and FAILED_PRECONDITION

- Added tests for:
  - future requested sequence replay behavior
  -  unsupported layout fail-closed behavior
  - restart detection in the KV monitor
  - stale replay skipping in the KV monitor
  
- Added operator docs explaining:
  - upstream vLLM works as-is
  - --kv-events-config is required for KV event streaming
  - current residual protocol limitations
- Stabilized test_openai_router_circuit_breaker by replacing an environment-dependent invalid hostname with a deterministic closed localhost port.

### Upstream vLLM compatibility

This works with upstream vLLM as-is. No vllm code changes are required.

To enable KV event streaming, run the stock vLLM gRPC server with --kv-events-config.

### Residual limitation

Restart recovery is currently sequence-based.

SMG can detect a restart when the first fresh batch after reconnect has a lower sequence number than the last applied batch. If a restarted backend catches back up before SMG reconnects, the current protocol has no producer epoch/generation identifier, so that restart is not distinguishable from normal continuation.

Unsupported null-block/sparse BlockStored layouts also remain fail-closed until the stream carries enough per-block token range metadata for lossless translation.

### Testing

Ran:

  - python -m compileall grpc_servicer/smg_grpc_servicer/vllm/kv_events.py
    grpc_servicer/smg_grpc_servicer/vllm/servicer.py
  - cargo +nightly fmt --all --check
  - cargo clippy --all-targets --all-features -- -D warnings
  - cargo test kv_event_monitor
  - cargo test -p smg --test routing_tests
    routing::test_openai_routing::test_openai_router_circuit_breaker -- --exact
    --nocapture

## Summary by CodeRabbit

* **New Features**
  * Added KV event streaming with a Subscribe endpoint to stream and replay KV cache events for cache-aware routing.

* **Documentation**
  * Added guides, examples, and usage notes for the --kv-events-config flag, including limitations and when it’s required.

* **Bug Fixes**
  * Improved restart detection and handling to avoid applying stale batches and recover state correctly.

* **Tests**
  * Added tests for streaming, replay, multi-rank sequencing, and layout validation.

* **Chores**
  * Added runtime dependencies needed for KV event streaming.

### DIFF
--- a/.github/workflows/e2e-gpu-job.yml
+++ b/.github/workflows/e2e-gpu-job.yml
@@ -113,7 +113,7 @@ jobs:
 
       # Run tests
       - name: Run E2E tests
-        timeout-minutes: 20
+        timeout-minutes: 25
         env:
           SHOW_WORKER_LOGS: "0"
           SHOW_ROUTER_LOGS: "1"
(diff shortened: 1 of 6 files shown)


slin1237:
Can you add perf differences 
we can try with smg with multiple vLLM gRPC workers
then run it with different lb policies, we can potentially use vLLM multi turn benchmark scripts

PR #755, opened 2026-03-13
fix(multimodal): fall back to config model type for aliased vision models

### LINKED ISSUE #754: [Bug]: Multimodal family detection fails for aliased/custom model IDs even when tokenizer/config point to the correct vision model
## Bug Description

Multimodal processor/spec selection is currently keyed off the request-side `model_id` string (for example `request.model` / `ctx.input.model_id`) rather than the loaded model config. This breaks multimodal handling when a backend worker exposes a vision model under a customized served name such as `custom-model` even though the underlying tokenizer/config files are still from a real supported family like `Qwen/Qwen3-VL-30B-A3B-Instruct`.

In that setup, tokenizer lookup and config loading can still succeed, but multimodal family detection fails because the hardcoded registrations match substrings of the request model name instead of falling back to `config.model_type`.

## Steps to Reproduce

1. Run a gRPC worker serving a Qwen3-VL family model, but expose it to SMG under a custom served name such as `custom-model`.
2. Ensure the worker still reports a valid `tokenizer_path` / model path pointing to the real Qwen3-VL files.
3. Let SMG register the tokenizer under the worker model ID (`custom-model`).
4. Send an OpenAI chat completion request with image content to SMG using `model: "custom-model"`.
5. Observe multimodal preparation failure before request dispatch.

Relevant code path:

- `model_gateway/src/routers/grpc/regular/stages/chat/preparation.rs`
- `model_gateway/src/routers/grpc/multimodal.rs`
- `crates/multimodal/src/registry/mod.rs`
- `crates/multimodal/src/registry/qwen3_vl.rs`
- `crates/multimodal/src/vision/image_processor.rs`

## Expected Behavior

If the loaded `config.json` / `preprocessor_config.json` clearly describe a supported multimodal family such as Qwen3-VL, SMG should select the correct multimodal spec and image processor even when the externally visible model name is customized.

At minimum, multimodal family detection should fall back to the loaded config metadata (for example `config.model_type`) when the request model name does not match any known family substring.

## Actual Behavior

SMG currently matches multimodal support using hardcoded substring checks against the request-side model name.

Examples:

- `Qwen3VLVisionSpec::matches()` requires the model ID string to contain both `qwen3` and `vl`
- `ImageProcessorRegistry::find()` uses case-insensitive `model_id.contains(pattern)` matching for patterns like `qwen3-vl`

When the request model is `custom-model`, both checks fail even though the loaded files are still the real Qwen3-VL model.

The request fails during multimodal preparation with an error equivalent to:

```text
Multimodal processing failed: Multimodal not supported for model: custom-model
```

## Component

multimodal

## Routing Policy (if applicable)

N/A

## Connection Mode

gRPC

## Configuration

```yaml
workers:
  - url: grpc://worker:port
    runtime_type: sglang
    models:
      - id: custom-model
```

Worker metadata / labels conceptually look like:

```yaml
served_model_name: custom-model
tokenizer_path: /models/Qwen3-VL-30B-A3B-Instruct
model_path: /models/Qwen3-VL-30B-A3B-Instruct
```

Example underlying model family:

- `Qwen/Qwen3-VL-30B-A3B-Instruct`

## Logs / Error Output

```text
Multimodal processing failed: Multimodal not supported for model: custom-model
```

Likely failure points in current code:

- `model_gateway/src/routers/grpc/multimodal.rs`: `model_registry.lookup(&metadata)`
- `crates/multimodal/src/registry/qwen3_vl.rs`: `matches()` uses `metadata.model_id`
- `crates/multimodal/src/vision/image_processor.rs`: `find(model_id)` uses substring matching on request model name

## Environment

OS: Darwin 25.3.0
cargo 1.93.1 (083ac5135 2025-12-15)
rustc 1.93.1 (01f6ddf75 2026-02-11)

## Deployment Environment

Local development

## Streaming Context

- [ ] Issue occurs with streaming (SSE) responses
- [ ] Issue occurs with non-streaming responses
- [x] Issue occurs in both modes

## Additional Context

This is a family-detection problem, not a tokenizer-loading problem.

`process_multimodal()` successfully loads `config.json` and `preprocessor_config.json` from the tokenizer source path before attempting multimodal family selection. However, the selection logic does not use that config to identify the model family.

A minimal fix would be:

1. Add a fallback family matcher that inspects loaded config metadata such as `config.model_type`.
2. Use that fallback in both the multimodal spec registry and image processor registry so the token-side and pixel-side selection stay aligned.
3. Preserve the current fast path based on request model name when it already matches.

This should allow aliases/custom served names to keep working without weakening the existing explicit registrations.

## Pre-submission Checklist

- [x] I have searched existing issues and discussions
- [x] I can reproduce this issue consistently
- [ ] I am using the latest version of SMG

YouNeedCryDear:
Fixes #754

### Problem

Multimodal family detection in the gRPC chat preparation path was keyed off the request-facing model name. When a worker exposed a supported vision model under an alias such as `custom-model`, tokenizer and config loading could succeed but multimodal spec selection and image processor lookup still failed before dispatch.

### Solution

Use the loaded model config as a fallback signal during multimodal family detection. The multimodal registry now checks `config.model_type` when `model_id` does not match, and the image processor registry does the same while preserving the existing fast path on `model_id`.

## Changes

- add `ModelMetadata::config_model_type()` and `matches_model_type_hints()` helpers for shared family matching logic
- update built-in multimodal specs to fall back from `model_id` to `config.model_type`
- collapse image processor lookup into `find(model_id, model_type)` with model-type fallback
- use the loaded `model_type` in `model_gateway` multimodal preprocessing before image processor selection
- add regression tests for aliased model IDs and fast-path preservation across multimodal specs and image processor lookup

## Test Plan

1. Configure a gRPC worker to serve a supported multimodal model under an aliased external model ID such as `custom-model`.
2. Keep `tokenizer_path` / `model_path` pointing at the real model artifacts containing the correct `config.json` and `preprocessor_config.json`.
3. Send a chat completion request with image content using the aliased model name.
4. Verify that SMG now resolves the correct multimodal spec and image processor instead of failing during preparation.
5. Run local verification:
   - `make fmt`
   - `make check`
   - `make test`

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Refactor**
  * Model matching now relies on explicit model-type hints for more accurate spec resolution instead of identifier substring checks.
  * Image processor lookup now accepts an optional model type and falls back to model-type-based selection when direct model-id resolution fails.

* **Tests**
  * Added unit tests verifying alias resolution via model-type hints and image-processor fallback and fast-path behavior.

### DIFF
--- a/crates/multimodal/src/vision/image_processor.rs
+++ b/crates/multimodal/src/vision/image_processor.rs
@@ -441,6 +441,10 @@ impl ImageProcessorRegistry {
             "phi3-vision",
             Box::new(super::processors::Phi3VisionProcessor::new()),
         );
+        registry.register(
+            "phi3_v",
+            Box::new(super::processors::Phi3VisionProcessor::new()),
+        );
 
         // Register LLaMA 4 Vision
         registry.register(
@@ -592,5 +596,15 @@ mod tests {
             .find("Qwen3-VL-30B-A3B-Instruct", Some("qwen2_vl"))
             .expect("qwen3 processor by model_id");
         assert_eq!(processor.model_name(), "qwen3-vl");
+    }
+
+    #[test]
+    fn test_registry_find_phi3_model_type_fallback() {
+        let registry = ImageProcessorRegistry::with_defaults();
+
+        let processor = registry
+            .find("custom-model", Some("phi3_v"))
+            .expect("phi3 processor by model_type");
+        assert_eq!(processor.model_name(), "phi3-vision");
     }
 }


CatherineSue:
For some reason I can't push to this branch. I think the overall changes are good. Just some small clean up with `traits.rs`. 

I made the new changes in PR 898

PR #882, opened 2026-03-24
fix(mesh): stop advertising 0.0.0.0 to peers

jshanson7:
### Problem

Mesh nodes used the same address for both listening and peer advertisement. When `--mesh-host` was left at `0.0.0.0`, gossip could propagate `0.0.0.0:<port>` back into cluster state and cause peers to dial an unroutable address.

### Solution

Split mesh bind and advertise addresses. The mesh server now listens on a bind address but advertises a separate routable address to peers. Add `--mesh-advertise-host` to the Rust and Python entrypoints, reject unspecified advertised addresses, and update HA docs/examples accordingly.

## Changes

- split mesh bind vs advertise addresses in mesh startup and gossip replies
- add `--mesh-advertise-host` to the Rust CLI and Python router args
- reject mesh configs that would advertise `0.0.0.0`
- preserve positional compatibility for the PyO3 `Router(...)` constructor
- update HA quick-start and Python entrypoint examples
- add a regression test covering advertised address in ping replies

## Test Plan

- `cargo +nightly fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test -p smg-mesh test_ping_advertises_configured_address -- --nocapture`
- `cargo check -p smg -p smg-python --all-targets`
- `pytest -v -s tests/test_arg_parser.py`

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -460,6 +460,7 @@ struct Router {
     enable_mesh: bool,
     mesh_server_name: Option<String>,
     mesh_host: String,
+    mesh_advertise_host: Option<String>,
     mesh_port: u16,
     mesh_peer_urls: Vec<String>,
 }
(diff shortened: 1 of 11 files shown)


jshanson7:
branch name fixed: https://github.com/lightseekorg/smg/pull/883

PR #884, opened 2026-03-24
fix(gateway): Update metric when removing unhealthy workers

ekzhang:
## Description
Bugfix: Updates the `smg_worker_pool_size` metric when removing unhealthy workers. 

### Problem

Previously removing unhealthy workers would not go through the same code path as `DELETE /workers/:id`. This means that the `smg_worker_pool_size` metric remains high even though workers were removed.

### Solution

This uses the same code path (JobQueue) to remove workers.

## Changes

Changed the unhealthy worker removal logic to submit a job queue request instead.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Refactor**
  * Health monitoring now enqueues worker-removal tasks to the background job queue when available and logs an error if enqueue fails.
  * If no job queue is available, the monitor marks workers as unhealthy but does not remove them.
  * Health monitoring is now connected to the background task system when present.
* **Tests**
  * Tests updated to assert healthy-worker counts and reflect the new removal-enqueue behavior.

### DIFF
--- a/model_gateway/src/core/worker_registry.rs
+++ b/model_gateway/src/core/worker_registry.rs
@@ -998,7 +998,10 @@ impl WorkerRegistry {
                                     "Removing unhealthy worker from registry"
                                 );
                                 next_check.remove(&url);
-                                if let Err(e) = job_queue.submit(Job::RemoveWorker { url: url.clone() }).await {
+                                if let Err(e) = job_queue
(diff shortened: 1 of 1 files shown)


ekzhang:
Sorry about initially missing this case when I added the option @slin1237 !

PR #922, opened 2026-03-26
perf(multimodal): pre-convert decoded images to RGB8 at decode time

slin1237:
## Summary
- Pre-converts decoded images to RGB8 format immediately after decoding in `MediaConnector::decode_image`
- Eliminates repeated downstream conversions when non-RGB8 images (RGBA, Luma, etc.) hit the fallback path in `rgb_bytes()`
- Single conversion at decode time replaces potentially multiple per-processor conversions

## Test plan
- [x] `cargo test -p llm-multimodal` — all 81 tests pass
- [x] `cargo test -p llm-multimodal -- vision_golden` — all 30 golden tests pass

## Summary by CodeRabbit

* **Refactor**
  * Adjusted internal image processing workflow.

PR #924, opened 2026-03-26
perf(multimodal): skip redundant to_rgb8 conversion when image is already RGB8

slin1237:
## Summary
- In `phi3_vision.rs`, `process_single_image` unconditionally called `image.to_rgb8()` even when the input was already `DynamicImage::ImageRgb8`, causing a needless full-pixel copy.
- Added a match guard that clones the image directly when it is already RGB8, and only converts otherwise.

## Test plan
- [x] `cargo test -p llm-multimodal` — all 81 tests pass
- [x] `cargo test -p llm-multimodal -- vision_golden` — all 30 golden tests pass

## Summary by CodeRabbit

* **Performance**
  * Optimized the vision image pipeline to skip unnecessary RGB conversions when images are already in RGB, reducing CPU and memory use and speeding up image processing for users.

### DIFF
--- a/crates/multimodal/src/vision/processors/phi3_vision.rs
+++ b/crates/multimodal/src/vision/processors/phi3_vision.rs
@@ -241,7 +241,10 @@ impl Phi3VisionProcessor {
         config: &PreProcessorConfig,
     ) -> (Array4<f32>, (usize, usize), usize) {
         // 1. Convert to RGB
-        let image = DynamicImage::ImageRgb8(image.to_rgb8());
+        let image = match image {
+            DynamicImage::ImageRgb8(_) => image.clone(),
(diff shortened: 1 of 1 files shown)


slin1237:
## Benchmark Results

Baseline (main) vs this PR — spot check (affects Phi3-Vision only):

| Benchmark | Baseline | This PR | Delta |
|---|---|---|---|
| qwen3_vl 640x480 | 1.564 ms | 1.554 ms | ~0% |
| llama4 640x480 | 3.417 ms | 3.400 ms | ~0% |

No regression. The optimization avoids a full pixel buffer copy in `process_single_image` when the input is already RGB8 (which is the common case, and guaranteed if PR #922 is also merged).

PR #928, opened 2026-03-26
perf(multimodal): avoid DynamicImage clone in no-op transform paths

slin1237:
## Summary

- **expand_to_square (LLaVA pad mode):** Skip the expand call entirely when the image is already square, using `Cow::Borrowed` to avoid both the expand operation and the clone. Also avoids computing the mean color when unnecessary.
- **center_crop (LLaVA standard mode, LlavaNext patch processor):** Check if the image dimensions already fit within the crop size before calling `center_crop`, avoiding a full pixel-buffer clone on the no-op path.

These clones copy the entire pixel buffer of a `DynamicImage`. In no-op cases (image already satisfies the constraint), skipping the function call eliminates a needless allocation + memcpy.

## Test plan
- [x] `cargo test -p llm-multimodal` -- all 81 tests pass
- [x] `cargo test -p llm-multimodal -- vision_golden` -- all 30 golden tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced image processing to skip unnecessary transformations when images already meet the required dimensions, reducing redundant operations.

### DIFF
--- a/crates/multimodal/src/vision/processors/llava.rs
+++ b/crates/multimodal/src/vision/processors/llava.rs
@@ -204,14 +204,19 @@ impl LlavaProcessor {
         let processed = match self.aspect_ratio {
             ImageAspectRatio::Pad => {
                 // Pad mode: expand to square with mean color padding, then resize
-                let mean_color = mean_to_rgb(&mean);
-                let squared = expand_to_square(image, mean_color);
+                let (w, h) = image.dimensions();
+                let squared = if w == h {
+                    std::borrow::Cow::Borrowed(image)
+                } else {
+                    let mean_color = mean_to_rgb(&mean);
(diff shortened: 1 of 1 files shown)


slin1237:
## Benchmark Results

Baseline (main) vs this PR — spot check (affects LLaVA code path):

| Benchmark | Baseline | This PR | Delta |
|---|---|---|---|
| qwen3_vl 640x480 | 1.564 ms | 1.554 ms | ~0% |
| llama4 640x480 | 3.417 ms | 3.396 ms | ~0% |

No regression. The optimization avoids `DynamicImage::clone()` (full pixel buffer copy) in no-op paths: `expand_to_square` when image is already square, `center_crop` when image already fits. Impact is proportional to how often these no-op conditions are hit in production — not exercised by the current benchmark suite which uses fixed-size synthetic images.

PR #944, opened 2026-03-27
fix(e2e): show worker logs inline on test failure

CatherineSue:
### Problem

When `SHOW_WORKER_LOGS=0` (CI default) and no `log_dir` is configured, worker logs are sent to `/dev/null`. When a test fails, there's no way to see what the backend worker was doing — making failures hard to debug.

### Solution

- Always write worker logs to `/tmp/smg_worker_logs/` even when `SHOW_WORKER_LOGS=0`
- Add a `pytest_runtest_makereport` hook that dumps the full worker log contents inline right when a test fails

## Changes

- `e2e_test/infra/worker.py` — Always write to log file instead of `/dev/null` when output is not shown
- `e2e_test/conftest.py` — Add `pytest_runtest_makereport` hook to dump worker logs on failure
- `.github/workflows/e2e-gpu-job.yml` — Revert SHOW_WORKER_LOGS back to 0 (no longer needed inline since the hook handles it)
- `e2e_test/chat_completions/test_worker_logs_on_failure.py` — Deliberate failure test to verify the hook works (DELETE AFTER VERIFICATION)

## Test Plan

The deliberate failure test (`test_deliberate_failure`) will fail in CI and should show worker logs inline in the pytest output immediately after the assertion error. Verify the logs appear, then delete the test file.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Worker logs are now automatically captured and displayed when E2E tests fail, providing enhanced visibility into test failures.

* **Tests**
  * Added E2E test to validate worker log dumping and reporting during test failures.

### DIFF
--- a/e2e_test/chat_completions/test_worker_logs_on_failure.py
+++ b/e2e_test/chat_completions/test_worker_logs_on_failure.py
@@ -1,0 +1,30 @@
+"""Temporary test to verify worker logs are dumped on failure.
+
+This test deliberately fails to confirm the pytest_runtest_makereport
+hook dumps worker logs inline. DELETE THIS FILE after verification.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+
+@pytest.mark.engine("sglang", "vllm")
+@pytest.mark.gpu(1)
+@pytest.mark.e2e
+@pytest.mark.parametrize("setup_backend", ["grpc"], indirect=True)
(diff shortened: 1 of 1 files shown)


CatherineSue:
This might cause the incorrect dump for same model_id + port. Close it for now.

PR #987, opened 2026-03-31
feat(sse): add shared SSE codec module

XinyueZhang369:
### Problem

The gateway has 3 duplicate SSE parsers and 2 duplicate chunk processors scattered across routers (gRPC, Anthropic, OpenAI, HTTP PD). Each uses its own encoding/decoding logic with a common pattern of `serde_json::to_string` + `format!("data: {json}\n\n")`, resulting in 2 heap allocations per SSE chunk. For a 500-token response (~50 chunks), this adds up.

### Solution

Create a shared SSE codec module at `model_gateway/src/routers/common/sse.rs` (Phase 1 of the SSE codec optimization design) with:

- **`SseEncoder`**: Serializes JSON directly into a reusable `Vec<u8>` buffer via `serde_json::to_writer`, then copies into `Bytes`. Reduces encoding from 2 allocations to 1 per chunk. Buffer grows to high-water mark and retains capacity across chunks via `.clear()`.
- **`SseDecoder`**: Byte buffer with cursor tracking and borrow-based frame parsing. Returns `SseFrame<'_>` with `Cow<str>` data — zero allocation for the common single-line case, only allocates when joining multi-line `data:` fields.
- **`parse_block`**: Standalone function for parsing a single already-extracted SSE block string.
- **`build_sse_response`**: Shared SSE response builder with standard headers.
- **Shared helpers**: `find_double_newline` using `memchr::memmem` (~10x faster than `windows(2)`), spec-compliant `parse_frame` with CRLF normalization, comment ignoring, and one-space strip.

## Changes

- Add `model_gateway/src/routers/common/mod.rs` — new module declaration
- Add `model_gateway/src/routers/common/sse.rs` — shared SSE codec (encoder, decoder, helpers, 35 tests)
- Modify `model_gateway/src/routers/mod.rs` — register `pub mod common`

## Test Plan

35 unit tests covering:
- Encoder: data-only format, event+data format, `[DONE]` sentinel, buffer reuse, output compatibility with existing `format!` patterns
- Decoder: single/multi-frame, chunk splitting, multi-byte UTF-8 splits, CRLF within frames, comments, `[DONE]` detection, buffer overflow (DoS protection), compact, flush, unknown fields ignored
- Roundtrip: encode → decode for data-only, event+data, done sentinel, multi-event sequences
- Parser: basic, data-only, empty, multi-line data, comments

```
cargo test -p smg --lib routers::common::sse
# test result: ok. 35 passed; 0 failed
```

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added Server‑Sent Events (SSE) streaming support: JSON message framing with optional event types and DONE sentinel, encoder/decoder with configurable buffer limits, robust newline/CRLF handling, UTF‑8 validation, compaction/flush semantics, comment/control handling, and an HTTP streaming-response helper.

* **Tests**
  * Extensive unit tests covering encoding/decoding, chunk-boundary and mixed-EOL cases, multiline data joining, buffer overflow/compaction/flush behavior, parsing rules, and round-trip integrity.


model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,912 @@
...
+        .status(StatusCode::OK)
+        .header(
+            CONTENT_TYPE,
+            HeaderValue::from_static("text/event-stream; charset=utf-8"),
+        )
+        .header("Cache-Control", HeaderValue::from_static("no-cache"))
+        .header("Connection", HeaderValue::from_static("keep-alive"))
+        .body(Body::from_stream(stream))
+        .expect("infallible: static headers and valid status code")
+}
```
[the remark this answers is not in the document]
XinyueZhang369: We will focus on creating the common sse module. Migrate to use the common sse module in existing code will be in next PR

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1044 @@
...
+    axum::response::Response::builder()
+        .status(StatusCode::OK)
+        .header(
+            CONTENT_TYPE,
+            HeaderValue::from_static("text/event-stream; charset=utf-8"),
+        )
+        .header("Cache-Control", HeaderValue::from_static("no-cache"))
+        .body(Body::from_stream(stream))
+        .expect("infallible: static headers and valid status code")
+}
```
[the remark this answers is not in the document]
XinyueZhang369: This is pre-existing pattern, thinking behavioral change for it might be out of scope for this PR

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1064 @@
...
+) -> axum::response::Response {
+    let stream = UnboundedReceiverStream::new(rx);
+    axum::response::Response::builder()
+        .status(StatusCode::OK)
+        .header(
+            CONTENT_TYPE,
+            HeaderValue::from_static("text/event-stream; charset=utf-8"),
+        )
+        .header(CACHE_CONTROL, HeaderValue::from_static("no-cache"))
+        .header(CONNECTION, HeaderValue::from_static("keep-alive"))
```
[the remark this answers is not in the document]
XinyueZhang369: Keep the same for now

model_gateway/src/routers/common/sse.rs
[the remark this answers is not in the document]
CatherineSue: I don't think this is gonna be a real concern. Hardly any client would be slow to consume tokens during streaming. The token generation is not gonna be that fast.

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1064 @@
...
+/// Errors that can occur during SSE decoding.
+#[derive(Debug)]
+pub enum SseDecodeError {
+    /// Buffer exceeded the configured maximum size.
+    BufferOverflow,
+    /// A complete frame contained invalid UTF-8.
+    InvalidUtf8(std::str::Utf8Error),
+    /// `flush()` called while complete frames remain in the buffer.
+    /// Drain `next_frame()` to `None` before calling `flush()`.
+    IncompleteFlush,
```
CatherineSue, line 165: I am not sure if this is an actual `SseDecodeError`, as in thoery, this should never happen if caller uses the `SseDecoder` correctly. It is ok to leave this one here. But I feel a panic or a `debug_assert` might be better.

model_gateway/src/routers/common/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod sse;
```
CatherineSue, line 30: I am not sure introducing a common just having sse is a good design.
If we look at our modules today, each router actually either serves as an actual router or an endpoint. 

If common is only gonna have sse.rs for now, I'd rather we just have sse.rs. Instead of having a common under routers.

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1064 @@
...
+        }
+        self.buf.extend_from_slice(chunk);
+        Ok(())
+    }
+
+    /// Yield the next complete SSE frame.
+    ///
+    /// Returns borrowed references into the internal buffer.
+    /// Call in a loop until `None`, then call `compact()`.
+    pub fn next_frame(&mut self) -> Option<Result<SseFrame<'_>, SseDecodeError>> {
```
CatherineSue: This signature means `SseFrame` holds a shared borrow on the decoder while `next_frame` needs a mutable borrow. The compiler forces sequential `get → own → drop → get` choreography. I feel it is not that intuiative. And I can't picture (in details) what this would mean for the integration of `SseDecoder` to the actual usage in routers today.

CatherineSue (commented):
The overall `SseEncoder` looks clean to me. However, it is a bit hard for me to understand `SseDecoder` use case.

For instance, in the codebase is `anthropic/sse.rs:208-264 (consume_and_forward)`:        

```                                                                              
                                                                                                                                                                                        
  // Current pattern — anthropic/sse.rs:238-248                                                                                                                                         
  while let Some(pos) = find_double_newline(&buffer) {                                                                                                                                  
      let frame_bytes = &buffer[..pos];                                                                                                                                                 
      let frame = std::str::from_utf8(frame_bytes)?;                                                                                                                                    
      if let Some((event_type, data)) = parse_sse_frame(frame) {                                                                                                                        
          processor.process(&event_type, &data).await?;  // <-- async!
      }                                                                                                                                                                                 
      buffer.drain(..pos + 2);
  }                                      
```

It would become 

```
  // Hypothetical migration to SseDecoder
  decoder.push(&chunk)?;                                                                                                                                                                
  loop {                                                                                                                                                                                
      match decoder.next_frame() {                                                                                                                                                      
          Some(Ok(frame)) => {                                                                                                                                                          
              // frame borrows from decoder — can't hold across .await
              // MUST clone/own everything before dropping                                                                                                                              
              let event_type = frame.event_type.map(str::to_owned);                                                                                                                     
              let data = frame.data.into_owned();                                                                                                                                       
              drop(frame);  // mandatory before next iteration                                                                                                                          
              processor.process(&event_type.as_deref().unwrap_or(""), &data).await?;                                                                                                    
          }                                                                                                                                                                             
          Some(Err(e)) => return Err(...),                                                                                                                                              
          None => break,                                                                                                                                                                
      }           
  }
  decoder.compact();
```

There is no lines change beneficial, and the entire pipeline is an ergonomics dance. We are doing `Extract → own → drop → await`. It is less clean than the `while let` existed in the code base.

XinyueZhang369:
Hi @CatherineSue thanks for the review, about the [decoder comment](https://github.com/lightseekorg/smg/pull/987#pullrequestreview-4046825268), Let me clarify the actual migration pattern and the concrete benefits.

   **The example in the review assumes we must copy, but we don't.** The frame borrow can be held across `.await` because the decoder is untouched during that time:

   ```rust
   decoder.push(&chunk)?;
   while let Some(Ok(frame)) = decoder.next_frame() {
       // frame borrows from decoder.buf — but decoder is not touched during await
       processor.process(frame.event_type.unwrap_or(""), &frame.data).await?;
       // frame dropped here at end of loop body → decoder free for next call
       drop(frame);
   }
   decoder.compact();
   ```

   No `.into_owned()`, no `to_owned()`, no explicit `drop()`. The compiler allows this because `next_frame()` is not called again until after the frame is dropped at the end of the loop body. This is the same `Extract →
   await → drop` flow as the current code, just without the intermediate allocations.

   **Concrete allocation savings per frame:**

   | Step | Current (`parse_sse_frame`) | `SseDecoder` |
   |------|---|---|
   | Event type | `value.trim().to_string()` → 1 alloc | `&str` borrowed from buffer → 0 |
   | Data (single line, 99% of traffic) | `data_lines.push(value.trim().to_string())` then `data_lines.join("\n")` → 2 allocs (String + join) | `Cow::Borrowed(&str)` → 0 |
   | Data (multi-line, rare) | Same pattern → 2+ allocs | `Cow::Owned` join → 1 alloc |
   | Buffer drain | `buffer.drain(..pos + 2)` per frame → O(n) memmove per frame | Cursor advance → O(1) per frame, one `compact()` memmove per chunk |

   **Other improvements over the current `find_double_newline` + `parse_sse_frame`:**
   - Frame boundary detection supports `\r\n\r\n`, `\r\r`, and mixed EOL combinations (current only handles `\n\n`)
   - Line splitting handles bare `\r` (current uses `str::lines()` which doesn't)
   - DoS protection: configurable max buffer size with auto-compact
   - Field parsing uses `split_once(':')` per spec instead of `strip_prefix("data:")`

model_gateway/src/routers/common/sse.rs
XinyueZhang369: If flush() is called with complete frames remaining (e.g. data: a\n\ndata: b), parse_frame would see the \n\n as an empty line separator and merge them into a single frame with data: "a\nb" — two events silently fused into 
  one. The caller would process wrong data without any indication something went wrong. So i'm thinking returning Err will let the caller detect and handle it, let me know if this make senses

model_gateway/src/routers/common/mod.rs
XinyueZhang369: Yeah okay, like just putting `sse.rs` under `model_gateway/src/routers`?

model_gateway/src/routers/common/sse.rs
XinyueZhang369: We don't have to copy the data out of the buffer each time, the sequence will be more like `get -> drop -> get`, more details is this [reply](https://github.com/lightseekorg/smg/pull/987#issuecomment-4173963692), let me know what do you think

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1064 @@
...
+    let mut saw_data = false;
+
+    for line in split_sse_lines(frame) {
+        if line.is_empty() || line.starts_with(':') {
+            continue;
+        }
+
+        // Per spec: split at first colon to get field name and value
+        let (field, value) = match line.split_once(':') {
+            Some((f, v)) => (f, v.strip_prefix(' ').unwrap_or(v)),
```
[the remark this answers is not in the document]
XinyueZhang369: Not applicable here — our upstream sources are HTTP JSON API streams (OpenAI, Anthropic, internal workers), not file-based streams. UTF-8 BOMs don't appear in HTTP response bodies from these providers. Adding BOM handling would be dead code with no test justification. If we ever integrate a BOM-producing source, we can add it then.

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1069 @@
...
+//!   Uses cursor tracking and borrow-based frame parsing to minimize allocations.
+
+use std::borrow::Cow;
+
+use axum::body::Body;
+use bytes::Bytes;
+use http::{
+    header::{HeaderValue, CACHE_CONTROL, CONNECTION, CONTENT_TYPE},
+    StatusCode,
+};
```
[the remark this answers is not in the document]
XinyueZhang369: `build_sse_response` is moved from grpc/common/responses/streaming.rs. Missing Connection header (present in some existing implementations) will need reconciliation when callers migrate. Want to keep the same as the existing one in this PR to avoid too many changes in one PR

model_gateway/src/routers/common/mod.rs
slin1237: yeah, if common is small enough, lets move this to under src/routers

CatherineSue: I saw @slin1237 created a common 2 days ago:
https://github.com/lightseekorg/smg/commits/main/model_gateway/src/routers/common 

So i guess it is ok for us to move this under common now.

Sorry for the back and forth.

model_gateway/src/routers/common/sse.rs
CatherineSue: @XinyueZhang369 Thanks for the detaled reply. You're right that NLL lets the borrow cross a single `.await` — that wasn't actually the crux.

Re-reading my comment, I can see "Extract → own → drop → await" made it sound like copying was the issue. It isn't. The concern is integration clarity. 

Could you paste the full migrated `consume_and_forward` (`anthropic/sse.rs:208-264`) side-by-side with `main`? 

My main concern is, even in your happy-path snippet, there is a transition dance, the `drop(frame)`. If we write the whole thing, I think there will be a pattern that shows this kind of loop, which is exactly the egonomics dance I referred to.

I agree with the time complexity in row 4 of the table you mentioned, but if the current API gives us from 3 lines of code to 5 lines or more, then that means we need to revise the API. 

Regarding the table:

Row 2 — Data, single line: ⚠️  Undercounted by 1. Actual current is 3 allocs, not 2:

```                                                                                                                                                                                                  
  let mut data_lines = Vec::new();                        // (lazy, no alloc yet)                                                                                                                                                                                                      
  data_lines.push(value.trim().to_string());              // 1 String alloc + 1 Vec buffer alloc (first push)                                                                                                                                                                          
  let data = data_lines.join("\n");                       // 1 String alloc (join always allocates a new String)                                                                                                                                                                       
  Vec::new() is lazy — the first push allocates the Vec buffer. XinyueZhang missed that. SseDecoder side (0) is correct — Some(Cow::Borrowed(value)) at line 457.                                                                                                                      
```

CatherineSue: Is it possible for us to change `SseFrame<'_>` to owned `SseFrame`? This way we make `compact` internal, collapse `flush` into a "drain remainder" at EOF.

CatherineSue: Some investigation from claude:

  **anthropic (streaming)** — `consume_and_forward` at sse.rs:244 already uses owned `(String, String)` from `parse_sse_frame` and feeds them to `processor.process(&event_type, &data).await`. Owned `SseFrame` is a drop-in, call site gets simpler.                                               
                                                                                                                                                                                                                                                                                       
  **openai/responses (block-based)** — all three sites (streaming.rs:771, accumulator.rs:89, utils.rs:139) pre-extract a block via their own `ChunkProcessor::next_block()` and then call `parse_sse_block(&raw_block)` which returns borrowed. They don't use a streaming decoder. Two options
   here, both fine:                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                       
  1. Keep `parse_block(&str) -> Option<SseFrame<'_>>` as-is in the PR — the frame borrows from the caller's `&str`, no `&mut self` involved, so no `compact/flush` dance. openai keeps its zero-alloc path.                                                                                    
  2. Or switch them to owned `SseFrame` — costs 1 `String::from` per event, which is noise next to the `serde_json::from_str(data)` that runs 2 lines later in all three sites.
                                                                                                                                                                                                                                                                                       
  The fix is surgical and asymmetric:                                                                                                                                                                                                                                                  
  - `SseDecoder::next_frame()` → owned SseFrame (kills the lifecycle dance)                                                                                                                                                                                                              
  - `parse_block(&str)` → keep `SseFrame<'_>` (no dance to kill; caller owns input)                                                                                                                                                                                                        
                                                                               
  Nothing in the existing code relies on borrowing from a `SseDecoder's` buffer specifically — because no such decoder exists yet.

CatherineSue: **gRPC encode sites** (`grpc/regular/streaming.rs:1331` `format_sse_chunk_into` + :2675):     
                                                                                                                                                                                               
 ```                                                                                                                                                                                                                                                                                      
  buffer.clear();                                                                                                                                                                                                                                                                      
  buffer.extend_from_slice(b"data: ");                                                                                                                                                                                                                                                 
  serde_json::to_writer(&mut *buffer, chunk)...;                                                                                                                                                                                                                                       
  buffer.extend_from_slice(b"\n\n");                                                                                                                                                                                                                                                   
 ```
                            
  This is literally the pattern `SseEncoder::encode_data generalizes`. Drop-in replacement, zero API concerns.                                                                                                                                                                           
                             
  **gRPC decode sites** (`grpc/regular/responses/streaming.rs:185` and :935):                                                                                                                                                                                                                
  ```                      
  let event_str = String::from_utf8_lossy(&chunk);                                                                                                                                                                                                                                     
  let event = event_str.trim();
  if let Some(json_str) = event.strip_prefix("data: ") {
      match serde_json::from_str::<ChatCompletionStreamResponse>(json_str) { ... }                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                                                                    
  ```
                                                                                                                                                                                                                                                                                 
  Two things worth flagging:                                                                                                                                                                                                                                         
                             
  1. **This code assumes one** `chunk` **= one complete SSE frame**. No buffering, no `find_double_newline`. If the gRPC→SSE translation ever produces a split frame (large payload, backpressure, proxy in the middle), these sites silently misparse. The new decoder would fix that. So         
  "consolidation" here isn't just dedup — it's correctness.
  2. Downstream usage is `serde_json::from_str(&frame.data)` immediately, same shape as anthropic and openai. Owned SseFrame fits perfectly:                                                                                                                                             
  ```                                                                                                                                                                                                                                                                                   
  decoder.push(&chunk)?;
  while let Some(frame) = decoder.next_frame()? {                                                                                                                                                                                                                                      
      if frame.is_done() { break; }                                                                                                                                                                                                                                                    
      match serde_json::from_str::<ChatCompletionStreamResponse>(&frame.data) { ... }
  }                                                                                                                                                                                                                                                                                    
  ```                      
  No lifecycle dance, and the existing `String::from_utf8_lossy(&chunk)` already allocates on every chunk today — an owned String in frame.data is a lateral move, not a regression.

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1055 @@
...
+            self.consumed += pos + delim_len;
+            if let Some(frame) = parse_frame(frame_str) {
+                return Some(Ok(frame));
+            }
+            // Control-only block (comment, event-only, id/retry) — skip it
+        }
+    }
+
+    /// Compact: shift unconsumed bytes to front. Call after draining frames.
+    /// Single O(n) memmove per batch instead of per-frame.
```
CatherineSue, line 225: is memmove a typo?

CatherineSue: Nvm, not a typo

model_gateway/src/routers/common/sse.rs
```diff
@@ -0,0 +1,1055 @@
...
+/// Errors that can occur during SSE encoding.
+#[derive(Debug)]
+pub enum SseEncodeError {
+    /// The event type contains newline characters, which would break SSE framing.
+    InvalidEventType,
+    /// JSON serialization failed.
+    Json(serde_json::Error),
+}
+
+impl std::fmt::Display for SseEncodeError {
```
CatherineSue: Can you check if we can use `#[derive(thiserror::Error)]` to remove the boilerplate Display + Error here? You can check https://github.com/lightseekorg/smg/blob/main/model_gateway/src/routers/common/retry.rs#L47-L51

model_gateway/src/routers/common/sse.rs
```diff
@@ -310,7 +303,7 @@ impl Default for SseDecoder {
 ///
 /// Use this when you already have a complete SSE block as a string (e.g. from
 /// an accumulator or rewrite path) and don't need the full [`SseDecoder`] machinery.
-pub fn parse_block(block: &str) -> Option<SseFrame<'_>> {
+pub fn parse_block(block: &str) -> Option<SseFrame> {
```
CatherineSue: Can we keep this to use borrowed?

XinyueZhang369: Yeah make sense since the block is not from the buffer, I changed here to be borrowed, and next frame stays as owned

model_gateway/src/routers/common/sse.rs
XinyueZhang369: Got it, this is cleaner, changed

CatherineSue (approved):

PR #1060, opened 2026-04-08
fix(mcp): avoid relisting tools on resumed responses

zhaowenzi:
### Problem

When `/v1/responses` resumed a prior turn with `previous_response_id`, SMG re-emitted `mcp_list_tools` for MCP bindings that were already established in earlier turns.

That differs from OpenAI behavior. On follow-up turns, OpenAI does not repeat `mcp_list_tools` for existing bindings. It only emits a new `mcp_list_tools` item when the request adds a new MCP binding.

This PR intentionally scopes the fix to the OpenAI Responses router path first. The gRPC Responses paths will be handled in a follow-up PR.

### Solution

SMG now reads prior response history for `mcp_list_tools.server_label` values and treats those bindings as already known.

On resumed turns:

- existing MCP bindings do not get a repeated `mcp_list_tools`
- newly added MCP bindings get exactly one new `mcp_list_tools`
- normal `mcp_call` behavior is unchanged for the current turn

This first PR only applies that behavior to the OpenAI Responses router path. A follow-up PR will align the gRPC Responses paths.

## Changes

- load existing MCP binding labels from stored response history referenced by `previous_response_id`
- pass existing binding labels through Responses routing, non-streaming, and streaming paths
- emit `mcp_list_tools` only for bindings not already seen in prior response output
- add regression coverage for resumed MCP flows in router/unit tests and e2e tests
- scope this PR to the OpenAI Responses router path only
- leave gRPC Responses alignment for a follow-up PR

### Before This Change

First request:

```bash
curl "$BASE_URL" \
  -d '{
    "model": "gpt-5.4",
    "input": "Search the web for Python programming language.",
    "tools": [
      {
        "type": "mcp",
        "server_label": "brave",
        "server_url": "http://127.0.0.1:8080/mcp",
        "require_approval": "never",
        "allowed_tools": ["brave_web_search"]
      }
    ]
  }'
```

Example response shape:

```json
{
  "id": "resp_1",
  "output": [
    {
      "type": "mcp_list_tools",
      "server_label": "brave"
    },
    {
      "type": "mcp_call",
      "server_label": "brave"
    },
    {
      "type": "message"
    }
  ]
}
```

Second request with the same MCP binding and `previous_response_id`:

```bash
curl "$BASE_URL" \
  -d '{
    "model": "gpt-5.4",
    "input": "Search the web for Rust programming language.",
    "previous_response_id": "resp_1",
    "tools": [
      {
        "type": "mcp",
        "server_label": "brave",
        "server_url": "http://127.0.0.1:8080/mcp",
        "require_approval": "never",
        "allowed_tools": ["brave_web_search"]
      }
    ]
  }'
```

Problematic response shape before this change:

```json
{
  "id": "resp_2",
  "output": [
    {
      "type": "mcp_list_tools",
      "server_label": "brave"
    },
    {
      "type": "mcp_call",
      "server_label": "brave"
    },
    {
      "type": "message"
    }
  ]
}
```

Why this is wrong:

- `brave` was already listed in the first turn
- resumed turn should not emit another `mcp_list_tools` for the same binding

Third request adds a new MCP binding:

```bash
curl "$BASE_URL" \
  -d '{
    "model": "gpt-5.4",
    "input": "Use deepwiki and brave to answer.",
    "previous_response_id": "resp_2",
    "tools": [
      {
        "type": "mcp",
        "server_label": "brave",
        "server_url": "http://127.0.0.1:8080/mcp",
        "require_approval": "never",
        "allowed_tools": ["brave_web_search"]
      },
      {
        "type": "mcp",
        "server_label": "deepwiki",
        "server_url": "https://mcp.deepwiki.com/mcp",
        "require_approval": "never"
      }
    ]
  }'
```

Problematic response shape before this change:

```json
{
  "id": "resp_3",
  "output": [
    {
      "type": "mcp_list_tools",
      "server_label": "brave"
    },
    {
      "type": "mcp_list_tools",
      "server_label": "deepwiki"
    },
    {
      "type": "mcp_call",
      "server_label": "brave"
    },
    {
      "type": "mcp_call",
      "server_label": "deepwiki"
    },
    {
      "type": "message"
    }
  ]
}
```

Why this is wrong:

- only `deepwiki` is new
- `brave` should not be relisted again

### After This Change

First request stays the same:

```json
{
  "id": "resp_1",
  "output": [
    {
      "type": "mcp_list_tools",
      "server_label": "brave"
    },
    {
      "type": "mcp_call",
      "server_label": "brave"
    },
    {
      "type": "message"
    }
  ]
}
```

Second request with the same binding now returns:

```json
{
  "id": "resp_2",
  "output": [
    {
      "type": "mcp_call",
      "server_label": "brave"
    },
    {
      "type": "message"
    }
  ]
}
```

Key change:

- no repeated `mcp_list_tools` for `brave`

Third request with a newly added binding now returns:

```json
{
  "id": "resp_3",
  "output": [
    {
      "type": "mcp_list_tools",
      "server_label": "deepwiki"
    },
    {
      "type": "mcp_call",
      "server_label": "brave"
    },
    {
      "type": "mcp_call",
      "server_label": "deepwiki"
    },
    {
      "type": "message"
    }
  ]
}
```

Key change:

- only the newly added `deepwiki` binding gets a new `mcp_list_tools`
- existing `brave` binding is not relisted

### Streaming Check

The same rule also applies to SSE event flow:

- before: resumed streams could emit `response.mcp_list_tools.in_progress` and `response.mcp_list_tools.completed` again for existing bindings
- after: resumed streams emit those events only for newly added bindings

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Resumption now preserves previously listed tools and only emits newly introduced tool listings when resuming.

* **Bug Fixes**
  * Resumed non-streaming and streaming responses no longer repeat already-seen tool-list emissions while still producing expected tool calls and completions.

* **Tests**
  * Added end-to-end and integration tests validating resumed behavior for streaming and non-streaming flows, exact listing, call, and completion events.

### DIFF
--- a/model_gateway/src/routers/openai/chat.rs
+++ b/model_gateway/src/routers/openai/chat.rs
@@ -134,8 +134,6 @@ pub(super) async fn route_chat(
     ctx.state.payload = Some(PayloadState {
         json: payload,
         url: url.clone(),
-        previous_response_id: None,
-        existing_mcp_list_tools_labels: Vec::new(),
     });
 
     // Wrap values in Arc to avoid cloning large objects on each retry attempt
--- a/model_gateway/src/routers/openai/chat.rs
+++ b/model_gateway/src/routers/openai/chat.rs
@@ -135,6 +135,7 @@ pub(super) async fn route_chat(
         json: payload,
         url: url.clone(),
         previous_response_id: None,
+        existing_mcp_list_tools_labels: Vec::new(),
     });
 
     // Wrap values in Arc to avoid cloning large objects on each retry attempt
--- a/model_gateway/src/routers/openai/mcp/tool_loop.rs
+++ b/model_gateway/src/routers/openai/mcp/tool_loop.rs
@@ -54,10 +54,9 @@ pub(crate) struct ToolLoopState {
 
 impl ToolLoopState {
     pub fn new(original_input: ResponseInput, prior_mcp_list_tools_labels: Vec<String>) -> Self {
-        let mut known_labels = prior_mcp_list_tools_labels
+        let known_labels = prior_mcp_list_tools_labels
             .into_iter()
             .collect::<HashSet<_>>();
-        known_labels.extend(existing_mcp_list_tools_labels(&original_input));
 
         Self {
             iteration: 0,
@@ -529,28 +528,6 @@ fn non_streaming_tool_item_id_source(item_id: &str, response_format: &ResponseFo
             .unwrap_or(item_id)
             .to_string(),
     }
-}
-
-fn existing_mcp_list_tools_labels(input: &ResponseInput) -> HashSet<String> {
-    let ResponseInput::Items(items) = input else {
-        return HashSet::new();
-    };
-
-    let Ok(items_value) = to_value(items) else {
-        return HashSet::new();
-    };
-
-    items_value
-        .as_array()
-        .into_iter()
-        .flatten()
-        .filter_map(|item| {
-            (item.get("type").and_then(|t| t.as_str()) == Some(ItemType::MCP_LIST_TOOLS))
-                .then(|| item.get("server_label").and_then(|v| v.as_str()))
-                .flatten()
-                .map(ToOwned::to_owned)
-        })
-        .collect()
 }
 
 pub(crate) fn mcp_list_tools_bindings_to_emit(
--- a/model_gateway/src/routers/openai/mcp/tool_loop.rs
+++ b/model_gateway/src/routers/openai/mcp/tool_loop.rs
@@ -8,7 +8,7 @@
 //! - Payload transformation for MCP tool interception
 //! - Metadata injection for MCP operations
 
-use std::io;
+use std::{collections::HashSet, io};
 
 use axum::http::HeaderMap;
 use bytes::Bytes;
@@ -21,7 +21,8 @@ use openai_protocol::{
 };
 use serde_json::{json, to_value, Value};
 use smg_mcp::{
-    mcp_response_item_id, McpToolSession, ResponseFormat, ResponseTransformer, ToolExecutionInput,
+    mcp_response_item_id, McpServerBinding, McpToolSession, ResponseFormat, ResponseTransformer,
+    ToolExecutionInput,
 };
 use tokio::sync::mpsc;
 use tracing::{debug, info, warn};
@@ -45,17 +46,25 @@ pub(crate) struct ToolLoopState {
     pub conversation_history: Vec<Value>,
     /// Original user input (preserved for building resume payloads)
     pub original_input: ResponseInput,
+    /// MCP bindings already represented by historical `mcp_list_tools` items.
+    pub existing_mcp_list_tools_labels: HashSet<String>,
     /// Transformed output items (mcp_call, web_search_call, etc.) - stored to avoid reconstruction
     pub mcp_call_items: Vec<Value>,
 }
 
 impl ToolLoopState {
-    pub fn new(original_input: ResponseInput) -> Self {
+    pub fn new(original_input: ResponseInput, prior_mcp_list_tools_labels: Vec<String>) -> Self {
+        let mut known_labels = prior_mcp_list_tools_labels
+            .into_iter()
+            .collect::<HashSet<_>>();
+        known_labels.extend(existing_mcp_list_tools_labels(&original_input));
+
         Self {
             iteration: 0,
             total_calls: 0,
             conversation_history: Vec::new(),
             original_input,
+            existing_mcp_list_tools_labels: known_labels,
             mcp_call_items: Vec::new(),
         }
     }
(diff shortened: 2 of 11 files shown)


model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -519,48 +528,97 @@ fn non_streaming_tool_item_id_source(item_id: &str, response_format: &ResponseFo
...
+}
+
+pub(crate) fn mcp_list_tools_bindings_to_emit(
+    existing_labels: &HashSet<String>,
+    bindings: &[McpServerBinding],
+) -> Vec<(String, String)> {
+    bindings
+        .iter()
+        .filter(|binding| !existing_labels.contains(&binding.label))
+        .map(|binding| (binding.label.clone(), binding.server_key.clone()))
```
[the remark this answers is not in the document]
zhaowenzi: Not a scope of this PR

model_gateway/tests/common/mock_worker.rs
```diff
@@ -666,8 +666,25 @@ async fn responses_handler(
...
+            .unwrap_or(false);
+        let has_previous_response_id = payload
+            .get("previous_response_id")
+            .and_then(|v| v.as_str())
+            .is_some_and(|id| !id.is_empty());
+        let has_prior_tool_context =
+            has_function_output || has_previous_response_id || has_mcp_history;
+
+        if has_tools && !has_prior_tool_context {
             // First turn: emit streaming tool call events using OpenAI-style function_call ids
```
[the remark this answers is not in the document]
zhaowenzi: Fixed

key4ng (approved):

model_gateway/src/routers/openai/chat.rs
```diff
@@ -132,6 +132,7 @@ pub(super) async fn route_chat(
         json: payload,
         url: url.clone(),
         previous_response_id: None,
+        existing_mcp_list_tools_labels: Vec::new(),
```
CatherineSue: Why is this in chat?

zhaowenzi: This is only here because `chat` and `responses` now share the same `PayloadState` / `RequestContext` plumbing. Just like chat won't use `previous_response_id` too

CatherineSue: Then the design here is a bit misleading or questionable at the beginning. We should not pass in these parameters to chat if they are not useful in chat pipeline.

model_gateway/tests/api/responses_api_test.rs
```diff
@@ -343,6 +343,171 @@ async fn test_final_response_hides_internal_mcp_trace_items() {
     mcp.stop().await;
 }
 
+#[tokio::test]
+async fn test_previous_response_id_does_not_repeat_mcp_list_tools_for_existing_binding() {
```
CatherineSue, line 347: Is this test added for ut coverage? I am a bit confused on why we still need `MockWorker` and the fake responses test here. cc @key4ng

zhaowenzi: There are already a few tests here at the same router/integration level. This one is intended to be in the same category: it covers a `previous_response_id` resume regression, which only shows up after one turn is routed/stored and a second turn resumes from that history, so it still needs `MockWorker`.

model_gateway/src/routers/openai/chat.rs
zhaowenzi: I split the Responses-only metadata out of the shared payload state.
`PayloadState` now only carries the prepared upstream request (`json + url`), and `previous_response_id` / `existing_mcp_list_tools_labels` are now kept only on the Responses path, so chat no longer needs to initialize unused fields.

Let me know if this does not work for you.

CatherineSue: Works for me.

model_gateway/src/routers/openai/responses/history.rs
```diff
@@ -225,6 +242,22 @@ fn deserialize_items_from_array(array: &Value) -> Vec<ResponseInputOutputItem> {
         .unwrap_or_default()
 }
 
+fn extract_mcp_list_tools_labels(array: &Value) -> Vec<String> {
```
CatherineSue, line 245: There is a `existing_mcp_list_tools_labels` in `tool_loop.rs` as well. 
I am not following why we need a similar function in `tool_loop.rs`.

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -519,23 +528,59 @@ fn non_streaming_tool_item_id_source(item_id: &str, response_format: &ResponseFo
     }
 }
 
+fn existing_mcp_list_tools_labels(input: &ResponseInput) -> HashSet<String> {
```
CatherineSue: The code looks unnecessary to me. `input: &ResponseInput` patches `ResponseInputOutputItem`. `ResponseInputOutputItem` has no `McpListTools` variant, the `Vec<ResponseInputOutputItem>` can never contain one. When you serialize it back with `to_value()`, you'll never get `{"type": "mcp_list_tools", ...}` in the output.

model_gateway/src/routers/openai/responses/history.rs
zhaowenzi: Resolved. We now keep this logic only in `history.rs`.

model_gateway/src/routers/openai/mcp/tool_loop.rs
zhaowenzi: Resolved. I removed the duplicate logic from `tool_loop.rs`.

CatherineSue (approved):

PR #1065, opened 2026-04-08
feat(storage): add conversation memory store substrate

zhoug9127:
### Problem

SMG already supports response, conversation, and conversation-item persistence, but it does not yet expose a generic conversation-memory store substrate that later request hooks can consume. We need that generic substrate in upstream first, without introducing specific planners, row semantics, or persistence behavior changes.

## Header Note (Placeholder)

Current memory request headers are placeholders and are still under discussion.
They are intentionally treated as provisional and may be renamed in a follow-up PR:

- `x-smg-memory-policy`
- `x-smg-memory-ltm-store-enabled`
- `x-smg-memory-subject-id`
- `x-smg-memory-recall-method`
- `x-smg-memory-embedding-model`
- `x-smg-memory-extraction-model`

### Solution

Add a generic conversation-memory store substrate and normalized request memory execution context.

This PR:
- adds `ConversationMemoryWriter` exposure through the storage factory and app context
- provides generic memory and none backend writers, while leaving other backends as `None`
- adds generic `memory_runtime` config plus validation for store readiness
- adds normalized request/header parsing into a `MemoryExecutionContext`
- threads the optional writer and normalized execution context through the OpenAI request context for later hook use

Non-goals in this PR:
- no conversation-memory row planning
- no response or conversation-item persistence hooks
- no Oracle-specific writers, schema bootstrapping, or job semantics

## Changes

- add `create_storage_bundle` plumbing and optional `conversation_memory_writer`
- add in-memory and no-op conversation memory writer implementations
- add `memory_runtime` config and validation
- add `model_gateway::memory` module for normalized request memory context
- wire router/request context to carry the optional writer and normalized execution context

## Test Plan

- `cargo test -p data-connector test_create_storage_bundle_memory_exposes_memory_writer -- --nocapture`
- `cargo test -p data-connector test_create_storage_bundle_none_exposes_memory_writer -- --nocapture`
- `cargo test -p smg test_builder_memory_runtime_flags_round_trip -- --nocapture`
- `cargo test -p smg test_reject_ltm_store_when_ltm_runtime_disabled -- --nocapture`
- `cargo test -p smg header_policy_store_and_recall_enables_store_request -- --nocapture`
- `cargo test -p smg missing_headers_produce_inactive_context -- --nocapture`
- `cargo test -p smg store_headers_become_active_when_runtime_is_ready -- --nocapture`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Per-request memory execution context via new memory module and HTTP headers
  * Storage now returns an optional conversation-memory writer via a bundle API
  * In-memory and no-op conversation-memory writers implemented

* **Behavioral Changes**
  * Storage creation preserves memory-writer handles and only wraps core storages when hooks are present
  * Startup/runtime validation rejects incompatible memory flags and disallowed hook+store combinations

* **Tests**
  * Added tests for config validation, header parsing, memory-writer presence, and writer behavior


model_gateway/src/memory/context.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl Policy {
+    fn from_value(value: Option<&str>) -> Self {
+        match value.map(normalize) {
+            Some("store_only") => Self::StoreOnly,
+            Some("store_and_recall") => Self::StoreAndRecall,
+            Some("none") | Some("no_policy") => Self::None,
+            _ => Self::None,
+        }
+    }
```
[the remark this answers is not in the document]
zhoug9127: Implemented. Header flag parsing is now case-insensitive for textual values.
  is_enabled(...) normalizes input and uses eq_ignore_ascii_case for "true", "yes", and "enabled" (while still accepting "1"), so mixed-case client values are handled correctly.

model_gateway/src/memory/context.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    fn allows_ltm_store(self) -> bool {
+        matches!(self, Self::StoreOnly | Self::StoreAndRecall)
+    }
+}
+
+fn is_enabled(value: Option<&str>) -> bool {
+    matches!(value.map(normalize), Some("1" | "true" | "yes" | "enabled"))
+}
```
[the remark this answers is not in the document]
zhoug9127: Implemented. Same fix applied in the shared is_enabled(...) helper: parsing is now robust to case variations (TRUE, Yes, ENABLED, etc.) via eq_ignore_ascii_case.

model_gateway/src/routers/openai/context.rs
```diff
@@ -139,11 +157,18 @@ impl RequestContext {
...
         }
     }
 }
 
 impl RequestContext {
+    pub fn refresh_memory_execution_context(&mut self) {
+        let headers = self.headers().cloned().unwrap_or_default();
+        self.memory_execution_context =
+            middleware::build_memory_execution_context(self.components.router_config(), &headers);
+    }
```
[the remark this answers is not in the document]
zhoug9127: Implemented. Removed the HeaderMap clone in refresh_memory_execution_context(). It now uses a borrowed reference (self.headers().unwrap_or(&empty_headers)) and passes that directly into build_memory_execution_context(...), avoiding the extra allocation/copy.

crates/data_connector/src/factory.rs
```diff
@@ -139,19 +167,36 @@ pub async fn create_storage(config: StorageFactoryConfig<'_>) -> Result<StorageT
 
     // Wrap backends in hooked storage when a hook is provided
     if let Some(hook) = config.hook {
+        if bundle.conversation_memory_writer.is_some() {
+            return Err(
+                "storage hooks are not yet supported with conversation memory writer backends"
+                    .to_string(),
+            );
```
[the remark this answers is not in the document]
zhoug9127: Implemented. Agreed this was a regression. The unconditional rejection was removed from create_storage_bundle(), so hook-enabled Memory/None backends no longer fail startup just because conversation_memory_writer exists. The factory now continues wrapping response/conversation/conversation_item storages as before. The guard was moved to runtime-aware validation in model_gateway/src/app_context.rs and only errors when memory_runtime.ltm_store_enabled is enabled with an unsupported hook+memory-writer combination.

model_gateway/src/routers/openai/context.rs
```diff
@@ -6,112 +6,177 @@ use axum::http::HeaderMap;
...
 #[derive(Clone)]
+/// Components shared by all OpenAI router request types.
 pub struct SharedComponents {
+    /// HTTP client used for worker/proxy calls.
     pub client: reqwest::Client,
+    /// Router runtime configuration.
+    pub router_config: Arc<RouterConfig>,
 }
 
+/// Components required by Responses-specific flows.
```
key4ng: can we remove unrelated added comments in this file, same comments for other files

model_gateway/src/app_context.rs
```diff
@@ -542,12 +562,13 @@ impl AppContextBuilder {
             redis: config.redis.as_ref(),
             hook,
         };
-        let (response_storage, conversation_storage, conversation_item_storage) =
-            create_storage(storage_config).await?;
+        let bundle = create_storage_bundle(storage_config).await?;
+        validate_memory_writer_configuration(config, bundle.conversation_memory_writer.as_ref())?;
```
key4ng: is it duplicated with line 321? just keep the one in build

model_gateway/src/app_context.rs
```diff
@@ -666,3 +687,113 @@ impl Default for AppContextBuilder {
...
+) -> Result<(), String> {
+    if config.memory_runtime.ltm_store_enabled && memory_writer.is_none() {
+        return Err(
+            "memory_runtime.ltm_store_enabled is true but selected storage backend does not provide conversation memory writer".to_string(),
+        );
+    }
+
+    if config.memory_runtime.ltm_store_enabled
+        && config.storage_hook_wasm_path.is_some()
+        && memory_writer.is_some()
```
key4ng: is it always true

model_gateway/src/routers/openai/context.rs
```diff
@@ -139,73 +212,107 @@ impl RequestContext {
...
             state: ProcessingState::default(),
             storage_request_context: None,
+            memory_execution_context,
         }
     }
 }
 
 impl RequestContext {
+    /// Recompute memory execution context from current headers and router runtime settings.
+    pub fn refresh_memory_execution_context(&mut self) {
```
key4ng, line 182: double check if this method really working?

model_gateway/src/app_context.rs
zhoug9127: I feel this might be suggested by AI reviewer...

If `ltm_store_enabled=true + storage hook enabled + writer present `-> `fail fast` **only for now**, because hooked memory-writer path is not implemented yet...

Once we implement hooked ConversationMemoryWriter, we should remove/relax that guard and allow the combination.

model_gateway/src/routers/openai/context.rs
zhoug9127: Sure, I will do that!

Having those crazy lines trying to satisfy the 80% requirements docs coverage...

model_gateway/src/routers/openai/context.rs
zhoug9127: It’s a safety method for “headers changed later than handler builds RequestContext” scenarios.

Current code doesn’t do this in OpenAI routes, but the design allows it, so refresh_memory_execution_context() is the recompute step for that pattern.

Let me know if that is redundant.

model_gateway/src/app_context.rs
zhoug9127: Good catch, will do a full re-check

model_gateway/src/memory/context.rs
```diff
@@ -0,0 +1,187 @@
+use axum::http::HeaderMap;
+use serde::{Deserialize, Serialize};
+
+pub use super::headers::MemoryHeaderView;
+
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+/// Runtime feature flags that gate memory store/recall behavior.
+pub struct MemoryRuntimeConfig {
+    pub ltm_enabled: bool,
+    pub ltm_store_enabled: bool,
```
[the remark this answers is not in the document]
zhoug9127: Addressed in 155b1b06: MemoryRuntimeConfig now uses #[serde(default)] and has a regression test for partial deserialization defaults.

model_gateway/src/routers/openai/context.rs
key4ng: >  80% requirements docs coverage...

oh really where is the requirement

crates/data_connector/src/factory.rs
```diff
@@ -53,22 +73,36 @@ pub struct StorageFactoryConfig<'a> {
 /// # Errors
 /// Returns error string if required configuration is missing or initialization fails
 pub async fn create_storage(config: StorageFactoryConfig<'_>) -> Result<StorageTuple, String> {
```
key4ng: do we need keep both create_storage and create_storage_bundle? I feel it redundant, how about just delete create_storage and rename the create_storage_bundle

model_gateway/src/memory/context.rs
```diff
@@ -0,0 +1,202 @@
+use axum::http::HeaderMap;
+use serde::{Deserialize, Serialize};
+
+pub use super::headers::MemoryHeaderView;
+
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+/// Runtime feature flags that gate memory store/recall behavior.
+#[serde(default)]
+pub struct MemoryRuntimeConfig {
```
slin1237: if this is config, move this to config
config should be self contained instead of depending on other code in gw

slin1237 (changes requested):

model_gateway/src/memory/context.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub ltm_enabled: bool,
+    pub ltm_store_enabled: bool,
+}
+
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+/// Normalized per-request memory execution context derived from headers and runtime flags.
+pub struct MemoryExecutionContext {
```
slin1237, line 14: can you elaborate on those
i cant barely understand any of those means by just reading the variable names

model_gateway/src/memory/context.rs
```diff
@@ -0,0 +1,202 @@
...
+        || value.eq_ignore_ascii_case("enabled")
+}
+
+/// Trim header input before policy/flag parsing.
+fn normalize(value: &str) -> &str {
+    value.trim()
+}
+
+#[cfg(test)]
+mod tests {
```
slin1237, line 115: are those ut meaningful?

model_gateway/src/memory/headers.rs
```diff
@@ -0,0 +1,45 @@
+use axum::http::HeaderMap;
+
+// Placeholder header names for memory controls.
+// These names are under active discussion and may change in a follow-up PR.
+pub const MEMORY_POLICY_HEADER: &str = "x-smg-memory-policy";
```
slin1237: move header specifics to
`model_gateway/src/routers/header_utils.rs`

model_gateway/src/routers/openai/context.rs
zhoug9127: Screenshot 2026-04-14 at 8 40 38 AM

Removed the redundant comment lines.

crates/data_connector/src/factory.rs
zhoug9127: create_storage_bundle was redundant, so I removed the split and kept a single create_storage() that returns StorageBundle. That keeps the API surface smaller and avoids two names for the same operation.

model_gateway/src/memory/context.rs
zhoug9127: > Moved MemoryRuntimeConfig into the config module so config stays self-contained and does not depend on the gateway memory module anymore. memory/context.rs now imports it from crate::config.

Revised since latest change pushed:
MemoryRuntimeConfig now lives under the config layer (model_gateway/src/config/types.rs) and is re-exported via crate::config, so memory/context.rs depends on config rather than defining config-owned types itself.

model_gateway/src/memory/context.rs
zhoug9127: This module’s main responsibility is translating memory headers + runtime flags into MemoryExecutionContext, and these tests cover that decision logic directly:

- missing headers => inactive context
- store headers => store requested/active only when runtime permits it
- store_and_recall => recall requested/active
- store_only => no recall

Let me know if you find them redundant.

model_gateway/src/memory/context.rs
zhoug9127: Sure I expanded the docs on MemoryExecutionContext and its fields. The code now explains the distinction between *_requested (caller intent from headers/policy) and *_active (actually enabled after runtime gating), and also documents the purpose of subject_id, recall_method, embedding_model, and extraction_model.

For the Header related code - I think Sai is taking care of them and should have a PR shortly.

crates/data_connector/src/factory.rs
```diff
@@ -7,10 +7,15 @@ use url::Url;
...
+        ConversationItemStorage, ConversationMemoryWriter, ConversationStorage, ResponseStorage,
+    },
     hooked::{HookedConversationItemStorage, HookedConversationStorage, HookedResponseStorage},
     hooks::StorageHook,
-    memory::{MemoryConversationItemStorage, MemoryConversationStorage, MemoryResponseStorage},
+    memory::{
+        MemoryConversationItemStorage, MemoryConversationMemoryWriter, MemoryConversationStorage,
+        MemoryResponseStorage,
+    },
     noop::{NoOpConversationItemStorage, NoOpConversationStorage, NoOpResponseStorage},
```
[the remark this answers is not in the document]
zhoug9127: That would reintroduce the issue just fixed. A store-enabled config would pass startup checks but silently drop memory writes. The current fail-fast behavior is intentional and safer.

zhoug9127:
Superseded by https://github.com/lightseekorg/smg/pull/1149, which contains the updated implementation and follow-up fixes. Closing this PR to keep review history focused in the new thread.

PR #1074, opened 2026-04-09
fix(tokenizer): load merged EOS token IDs from config.json + generation_config.json

### LINKED ISSUE #1112: [Bug]: gRPC TRT-LLM path misses split EOS tokens for Kimi-K2.5, causing overgeneration past turn boundaries
## Bug Description

Kimi-K2.5 defines different EOS tokens in two config files:
- `config.json`: `eos_token_id: 163585` (`[EOS]`)
- `generation_config.json`: `eos_token_id: 163586` (`<|im_end|>`)

The model needs to stop at **both** tokens. In the affected gRPC path, only one EOS token is propagated as `stop_token_ids`, while Kimi-K2.5 requires both `[EOS]` and `<|im_end|>`. SGLang handles this internally by merging both sources, but the TRT-LLM gRPC path does not.

This causes severe overgeneration — the model generates past `<|im_end|>` turn boundaries, producing multi-turn spillover output.

## Steps to Reproduce

**Environment:**
- SMG: `v1.4.1` (commit `5319dc6e`)
- Image: `ghcr.io/lightseekorg/smg:1.4.1-trtllm-1.3.0rc10`
- TRT-LLM: commit `a2557ad9`
- Model: `nvidia/Kimi-K2.5-NVFP4`

**Launch:**
```bash
smg serve \
  --model <path-to-Kimi-K2.5-NVFP4> \
  --router-model-path nvidia/Kimi-K2.5-NVFP4 \
  --router-tool-call-parser kimik2 \
  --router-reasoning-parser kimi_k25 \
  --backend trtllm --tp_size 8 \
  --host 0.0.0.0 --port 8990
```

**Request:**
```bash
curl -s http://localhost:8990/v1/chat/completions \
  -d '{
    "model": "nvidia/Kimi-K2.5-NVFP4",
    "messages": [{"role": "user", "content": "What is the weather in San Francisco?"}],
    "tools": [
      {"type": "function", "function": {"name": "getCurrentWeather", "description": "Get the current weather in a given location", "parameters": {"type": "object", "properties": {"location": {"type": "string", "description": "City and state"}}, "required": ["location"]}}}
    ],
    "max_tokens": 500
  }'
```

## Expected Behavior

Model generates one assistant turn (~10 tokens for a single tool call), stops at `<|im_end|>` (163586).

## Actual Behavior

Model generates **253 tokens** for what should be a single tool call. Output is 723 chars of multi-turn spillover — the model generates past `<|im_end|>`, hallucinating a user turn, a tool response, and a follow-up assistant message:

```
content length: 723 chars
completion_tokens: 253 (expected ~10)
leaked markers: ['<|im_end|>', '<|im_assistant|>', '<|im_middle|>']
```

The model doesn't stop at `<|im_end|>` (163586) because only `[EOS]` (163585) is in the gRPC `stop_token_ids`. The `<|im_end|>` token is defined in `generation_config.json` but SMG doesn't merge it into the stop tokens sent to TRT-LLM.

SGLang backend is **not** affected (handles EOS merging internally).

## Component

- Tokenizer (`crates/tokenizer/src/tiktoken.rs`, `crates/tokenizer/src/traits.rs`)
- gRPC Client (`crates/grpc_client/src/trtllm_service.rs`)
- Model Gateway (request building stages)

## Connection Mode

gRPC (TRT-LLM backend only)

## Streaming Context

- [x] Issue occurs with streaming (SSE) responses
- [x] Issue occurs with non-streaming responses
- [x] Issue occurs in both modes

ConnorLi96:
## Problem

Models like Kimi-K2.5 define different EOS tokens in `config.json` (EOS=163585) and `generation_config.json` (`<|im_end|>`=163586). TRT-LLM's gRPC path only uses one source, causing infinite generation past turn boundaries (18/60 → 55/60 on fc-dash).

https://huggingface.co/nvidia/Kimi-K2.5-NVFP4/blob/main/config.json#L12
https://huggingface.co/nvidia/Kimi-K2.5-NVFP4/blob/main/generation_config.json#L3

## Solution

- Add `load_eos_token_ids()` to merge EOS IDs from both config files
- Expose `eos_token_ids()` on the `Tokenizer` trait
- Pass merged IDs as `stop_token_ids` in TRT-LLM gRPC requests
- Remove the hardcoded `<|im_end|>` stop string workaround

## Changes

- `crates/tokenizer/src/traits.rs` — add `eos_token_ids()` to trait
- `crates/tokenizer/src/tiktoken.rs` — `load_eos_token_ids()` impl
- `crates/tokenizer/src/huggingface.rs` — impl for HF tokenizer
- `crates/tokenizer/src/mock.rs` — mock impl
- `crates/grpc_client/src/trtllm_service.rs` — pass merged IDs, remove hardcoded stop string
- `model_gateway/src/routers/grpc/client.rs` — plumb new field
- `model_gateway/.../harmony/stages/request_building.rs` — pass EOS IDs
- `model_gateway/.../regular/stages/chat/request_building.rs` — pass EOS IDs

## Test Plan

- Verified Kimi-K2.5 on TRT-LLM: fc-dash 55/60 (was 18/60 without EOS fix)
- SGLang path unaffected (handles EOS merging internally)

Checklist:

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Support for discovering and loading End-of-Sequence (EOS) token IDs from model config files.
  * Tokenizer trait and implementations now expose EOS token IDs and propagate them through caching wrappers.
  * Request builders and request construction paths now accept and forward explicit EOS token ID data.
  * Tokenizer download logic expanded to include additional config files and a new public EOS helper module.

* **Tests**
  * Added unit tests covering EOS ID extraction, merging, and edge cases.

Closes #1112

### DIFF
--- a/crates/grpc_client/src/trtllm_service.rs
+++ b/crates/grpc_client/src/trtllm_service.rs
@@ -265,6 +265,10 @@ impl TrtllmServiceClient {
         clippy::unused_self,
         reason = "method receiver kept for consistent public API across gRPC backends"
     )]
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "mirrors vLLM backend API; refactoring to a params struct is tracked separately"
+    )]
     pub fn build_generate_request_from_chat(
         &self,
         request_id: String,
@@ -273,6 +277,7 @@ impl TrtllmServiceClient {
         token_ids: Vec<u32>,
         multimodal_input: Option<proto::MultimodalInput>,
         tool_call_constraint: Option<(String, String)>, // (constraint_type, constraint_value)
+        eos_token_ids: &[u32],
     ) -> Result<proto::GenerateRequest, String> {
         // Build sampling config
         let sampling_config = Self::build_sampling_config_from_chat(body);
@@ -287,6 +292,12 @@ impl TrtllmServiceClient {
 
         let max_tokens = body.max_completion_tokens.unwrap_or(2048);
 
+        let stop_token_ids: Vec<u32> = if body.ignore_eos {
+            vec![]
+        } else {
+            eos_token_ids.to_vec()
+        };
+
         let grpc_request = proto::GenerateRequest {
             request_id,
             tokenized: Some(proto::TokenizedInput {
@@ -299,7 +310,7 @@ impl TrtllmServiceClient {
             max_tokens,
             streaming: body.stream,
             stop,
-            stop_token_ids: vec![],
+            stop_token_ids,
             ignore_eos: body.ignore_eos,
             bad: vec![],
             bad_token_ids: vec![],
@@ -414,6 +425,7 @@ impl TrtllmServiceClient {
         processed_text: String,
         token_ids: Vec<u32>,
         constraint: Option<(String, String)>,
+        eos_token_ids: &[u32],
     ) -> Result<proto::GenerateRequest, String> {
         let sampling_config = Self::build_sampling_config_from_responses(body);
         let output_config = proto::OutputConfig {
@@ -442,7 +454,7 @@ impl TrtllmServiceClient {
             max_tokens,
             streaming: body.stream.unwrap_or(false),
             stop: vec![],
-            stop_token_ids: vec![],
+            stop_token_ids: eos_token_ids.to_vec(),
             ignore_eos: false,
             bad: vec![],
             bad_token_ids: vec![],
--- a/crates/tokenizer/src/huggingface.rs
+++ b/crates/tokenizer/src/huggingface.rs
@@ -82,6 +82,12 @@ impl HuggingFaceTokenizer {
             }
         }
 
+        // Load merged EOS token IDs from config.json + generation_config.json
+        let mut special_tokens = special_tokens;
+        if let Some(dir) = std::path::Path::new(file_path).parent() {
+            special_tokens.eos_token_ids = crate::eos::load_eos_token_ids(dir);
+        }
+
         Ok(HuggingFaceTokenizer {
             tokenizer,
             special_tokens,
@@ -208,6 +214,7 @@ impl HuggingFaceTokenizer {
             cls_token: find_token(&["[CLS]", "<cls>", "<CLS>"]),
             mask_token: find_token(&["[MASK]", "<mask>", "<MASK>"]),
             additional_special_tokens,
+            ..Default::default()
         }
     }
 
@@ -382,6 +389,10 @@ impl TokenizerTrait for HuggingFaceTokenizer {
     }
     fn think_in_prefill(&self) -> bool {
         self.chat_template.think_in_prefill()
+    }
+
+    fn eos_token_ids(&self) -> &[TokenIdType] {
+        &self.special_tokens.eos_token_ids
     }
 
     fn set_chat_template(&mut self, template: String) -> Result<()> {
--- a/crates/tokenizer/src/tiktoken.rs
+++ b/crates/tokenizer/src/tiktoken.rs
@@ -119,6 +119,7 @@ fn parse_special_tokens(config: &serde_json::Value) -> SpecialTokens {
         cls_token: get_str("cls_token"),
         mask_token: get_str("mask_token"),
         additional_special_tokens: additional,
+        ..Default::default()
     }
 }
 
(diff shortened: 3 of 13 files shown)


crates/tokenizer/src/huggingface.rs
```diff
@@ -82,6 +82,12 @@ impl HuggingFaceTokenizer {
             }
         }
 
+        // Load merged EOS token IDs from config.json + generation_config.json
+        let mut special_tokens = special_tokens;
+        if let Some(dir) = std::path::Path::new(file_path).parent() {
```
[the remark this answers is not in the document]
slin1237: this is not a nit
this is a critical piece that should be fixed

crates/tokenizer/src/huggingface.rs
```diff
@@ -82,6 +82,12 @@ impl HuggingFaceTokenizer {
             }
         }
 
+        // Load merged EOS token IDs from config.json + generation_config.json
+        let mut special_tokens = special_tokens;
```
slin1237, line 86: when loading hf or tiktokenizer
do we not have those special tokens already?

crates/tokenizer/src/tiktoken.rs
```diff
@@ -42,6 +42,60 @@ struct TiktokenConfig {
...
+                            ids.insert(id as TokenIdType);
+                        }
+                    }
+                }
+                _ => {}
+            }
+        }
+    }
+
+    ids.into_iter().collect()
```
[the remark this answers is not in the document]
slin1237: this should not be nit
it should be addressed before merging

crates/grpc_client/src/trtllm_service.rs
```diff
@@ -273,6 +277,7 @@ impl TrtllmServiceClient {
         token_ids: Vec<u32>,
         multimodal_input: Option<proto::MultimodalInput>,
         tool_call_constraint: Option<(String, String)>, // (constraint_type, constraint_value)
+        eos_token_ids: &[u32],
```
slin1237, line 280: why do we only change trt?
and is this specific to kimi-k2?

slin1237 (changes requested):

model_gateway/src/routers/grpc/harmony/stages/request_building.rs
```diff
@@ -175,6 +175,10 @@ impl PipelineStage for HarmonyRequestBuildingStage {
                 ProtoGenerateRequest::Vllm(Box::new(req))
             }
             GrpcClient::Trtllm(trtllm_client) => {
+                let eos_ids = ctx
```
slin1237: this is not even supported for harmony afaik

crates/grpc_client/src/trtllm_service.rs
ConnorLi96: since SGLang and vLLM merge EOS tokens internally — they read both config files in their Python startup code. so only TRTLLM. 

not Kimi-specific, the fix is generic, just Kimi-K2.5 is the first model we see this issue.

crates/tokenizer/src/huggingface.rs
ConnorLi96: `tokenizer_config.json` only provides `eos_token` as a string (e.g. "[EOS]"), but not the token ID. The actual EOS token IDs come from `config.json` and `generation_config.json`. Again, SGLang and vLLM handle this merge internally in their Python code. TRT-LLM's gRPC path doesn't, so SMG needs to read and merge them explicitly.

Also found and fixed a related issue: when SMG loads the tokenizer via HF Hub (by model ID rather than local path), the hub downloader only fetched tokenizer files, not `config.json/generation_config.json`. Fixed in hub.rs so these are now included in the download.

crates/tokenizer/src/huggingface.rs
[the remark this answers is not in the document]
ConnorLi96: Agreed, fixed. Extracted `collect_eos_ids()` helper and replaced the duplication with a loop over both config filenames. Also moved the function out of `tiktoken.rs` into a shared crates/tokenizer/src/eos.rs module.

model_gateway/src/routers/grpc/harmony/stages/request_building.rs
ConnorLi96: You're right, Harmony doesn't run on TRT-LLM. the TRT-LLM arm just passes `&[]` now to satisfy the function signature.

since we change the signature of `build_generate_request_from_chat` and `build_generate_request_from_responses`, we have to do this otherwise can't compile IIUC.

crates/grpc_client/src/trtllm_service.rs
CatherineSue: Fix this in trt instead. It is not in our scope to pass this to their backend, just because this is a bug in their end.

uralik:
@slin1237  could you take a look on the latest update here? this is quite huge issue that prevents seamlessly using Kimi models served via vllm with smg. thanks!

PR #1091, opened 2026-04-10
fix(mcp): re-add image generation tool routing

TingtingZhou7:
### Problem
Remove gRPC from https://github.com/lightseekorg/smg/pull/1057/changes#diff-477dfd1a461c069d89a1fc610c9eb9e8d6f98fc6924ebc8de092a8ceca00d007

### Solution
Disable image generation tool in gRPC.

## Changes
Disable image generation tool in gRPC.

## Test Plan
Tested with gRPC and image-generation-tool is disabled now. 

Screenshot 2026-04-10 at 2 00 05 PM

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added first-class image generation tool with configurable parameters (size, quality, background, format, compression, moderation, model).
  * Image generation calls stream progress events (in-progress, generating, partial image, completed).

* **Bug Fixes / Reliability**
  * Improved error handling with fallback text extraction for failed image-generation responses.

* **Behavioral**
  * gRPC Responses routes now explicitly reject image-generation tools (returns unsupported_tool).


CatherineSue:
Seems lint is failing. @TingtingZhou7 Can you take a look?

model_gateway/src/routers/tool_output_context.rs
```diff
@@ -0,0 +1,36 @@
...
+    response_format: &ResponseFormat,
+    output: &Value,
+) -> String {
+    match response_format {
+        ResponseFormat::ImageGenerationCall => {
+            let is_error = is_image_generation_error(output);
+            let note = if is_error {
+                extract_image_generation_fallback_text(output).unwrap_or_default()
+            } else {
+                "Successfully generated the image".to_string()
```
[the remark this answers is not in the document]
TingtingZhou7: It will always be there, for oci mcp server

model_gateway/src/routers/tool_output_context.rs
```diff
@@ -0,0 +1,36 @@
...
+            };
+            let summary = json!({
+                "tool": "generate_image",
+                "status": if is_error { "failed" } else { "completed" },
+                "note": note
+            });
+            summary.to_string()
```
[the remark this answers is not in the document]
TingtingZhou7: applied change

model_gateway/src/routers/tool_output_context.rs
```diff
@@ -0,0 +1,36 @@
+use serde_json::{json, Value};
+use smg_mcp::{extract_image_generation_fallback_text, is_image_generation_error, ResponseFormat};
+
+/// Build tool output text for model context.
+///
+/// This is format-driven and intended to support per-tool compaction policies.
+/// Currently only `ResponseFormat::ImageGenerationCall` is compacted into a
+/// minimal fixed summary (no payload/status) to avoid feeding large binary
+/// image data back into the next model turn. Other formats are currently no-op
+/// and return `output.to_string()` unchanged.
```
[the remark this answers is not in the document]
CatherineSue: @TingtingZhou7 Please take a look

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -160,12 +163,12 @@ pub(crate) async fn execute_streaming_tool_calls(
                 continue;
             }
         };
-
+        apply_request_tool_overrides(&response_format, original_body, &mut arguments);
```
CatherineSue, line 227: Why do we need to override tool info in the requests?

TingtingZhou7:
> Seems lint is failing. @TingtingZhou7 Can you take a look?
Fixed

crates/mcp/src/transform/transformer.rs
```diff
@@ -177,8 +265,69 @@ impl ResponseTransformer {
...
+                    .as_ref()
+                    .and_then(|v| v.as_object())
+                    .and_then(|obj| obj.get("result"))
+                    .and_then(|v| v.as_str())
+                    .map(String::from)
+            })
+            .or_else(|| extract_image_generation_fallback_text(result))
+            .or_else(|| payload.as_str().map(String::from))
+            .or_else(|| result.as_str().map(String::from))
+            .or_else(|| Some(payload.to_string()));
```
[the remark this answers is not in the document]
TingtingZhou7: Fixed.

crates/mcp/src/transform/transformer.rs
```diff
@@ -177,8 +265,69 @@ impl ResponseTransformer {
...
+        let payload = Self::image_payload_from_wrapped_content(result)
+            .unwrap_or_else(|| Value::String(Self::flatten_mcp_output(result)));
+        let parsed_payload = payload
+            .as_str()
+            .and_then(|s| serde_json::from_str::<Value>(s).ok());
+        let obj = payload
+            .as_object()
+            .or_else(|| parsed_payload.as_ref().and_then(|v| v.as_object()));
+
+        let status = ImageGenerationCallStatus::Completed;
```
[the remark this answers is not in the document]
TingtingZhou7: Align with other tools to mark as Completed no matter fail or succeed

model_gateway/src/routers/grpc/harmony/builder.rs
```diff
@@ -425,6 +425,7 @@ impl HarmonyBuilder {
                             ResponseTool::Function(_) => "function",
                             ResponseTool::WebSearchPreview(_) => "web_search_preview",
                             ResponseTool::CodeInterpreter(_) => "code_interpreter",
+                            ResponseTool::ImageGeneration(_) => "image_generation",
```
CatherineSue: I'm not following the design here. It seems Harmony still accepts the image generation tool call?

model_gateway/src/routers/tool_output_context.rs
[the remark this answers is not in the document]
TingtingZhou7: Fixed. I have updated the comment to include the "status" field.

model_gateway/src/routers/grpc/harmony/builder.rs
CatherineSue: Nvm. I saw this is just the builder

TingtingZhou7: Yes. Because harmony and openai share this ResponseTool file. And if i delete it, complition will fail. I have added it to make build succeed. But if you check the changes in "Utils", I have reject the request with image_generation_tool.

TingtingZhou7: Image

model_gateway/src/routers/grpc/common/responses/streaming.rs
```diff
@@ -474,6 +476,7 @@ impl ResponseStreamEventEmitter {
         let event_type = match response_format {
             ResponseFormat::WebSearchCall => WebSearchCallEvent::IN_PROGRESS,
             ResponseFormat::CodeInterpreterCall => CodeInterpreterCallEvent::IN_PROGRESS,
+            ResponseFormat::ImageGenerationCall => ImageGenerationCallEvent::IN_PROGRESS,
```
CatherineSue: In theory, the code path here won't  be triggered, as we are disabling the image_generation_tool and returning a 400. Is that right?

model_gateway/src/routers/openai/mcp/tool_loop.rs
TingtingZhou7: To forword the parameters like image quality / compression/ from response api to mcp client. 

Example as below: 
```
tingzhou@tingzhou-mac .ssh %  curl -sS "http://127.0.0.1:9999/v1/responses" \
    -H "Authorization: Bearer $OPENAI_API_KEY" \
    -H "opc-request-id: $(uuidgen)" \
    -H "opc-compartment-id: ocid1.compartment.oc1..aaaaaaaathgntpo75bdehisnl6wkxfc4slkd6rpheafbt5a6ekm2ri4bmeva" \
    -d '{
      "model":"gpt-5.1",
      "input":"Generate an image of a red panda reading a book under a maple tree, watercolor style.",
      "tools":[
        {
          "type":"image_generation",
          "model":"openai.gpt-image-1.5",
          "quality":"high",
          "output_format":"png",
          "output_compression":100,
          "size":"1024x1024",
          "background":"auto"
        }
      ],
      "stream":false
    }' | tee ~/Desktop/response-new2.json
```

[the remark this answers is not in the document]
TingtingZhou7: Streaming change will be in seperate PR.

model_gateway/src/routers/grpc/common/responses/streaming.rs
TingtingZhou7: Yes. User request are gated before this.

model_gateway/src/routers/grpc/common/responses/streaming.rs
```diff
@@ -532,6 +538,7 @@ impl ResponseStreamEventEmitter {
         match response_format {
             Some(ResponseFormat::WebSearchCall) => OutputItemType::WebSearchCall,
             Some(ResponseFormat::CodeInterpreterCall) => OutputItemType::CodeInterpreterCall,
+            Some(ResponseFormat::ImageGenerationCall) => OutputItemType::ImageGenerationCall,
```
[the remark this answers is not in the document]
TingtingZhou7: Image generation is now gated at the request level.

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -253,6 +261,58 @@ pub(crate) fn prepare_mcp_tools_as_functions(payload: &mut Value, session: &McpT
...
+/// Extract request-level builtin tool overrides to merge into tool-call arguments.
+///
+/// Currently this is intentionally scoped to image generation only.
+/// We can extend this to other builtin tools later if needed.
+fn request_tool_overrides(
+    response_format: &ResponseFormat,
+    original_body: &ResponsesRequest,
+) -> Option<Value> {
+    if !matches!(response_format, ResponseFormat::ImageGenerationCall) {
+        return None;
```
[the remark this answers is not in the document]
TingtingZhou7: we don't suppoprt any other image-formatted mcp tool. OCI format only.

model_gateway/src/routers/openai/responses/streaming.rs
```diff
@@ -147,6 +148,9 @@ pub(super) fn apply_event_transformations_inplace(
                             // Determine item type and ID prefix based on response_format
                             let (new_type, id_prefix) = match response_format {
                                 ResponseFormat::WebSearchCall => (ItemType::WEB_SEARCH_CALL, "ws_"),
+                                ResponseFormat::ImageGenerationCall => {
+                                    (ItemType::IMAGE_GENERATION_CALL, "ig_")
+                                }
```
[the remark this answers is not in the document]
TingtingZhou7: out of scope of this PR. Will validate whether this is true in follow up PR. RIght now, we are just following how web search tool is doing.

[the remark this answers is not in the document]
TingtingZhou7: out of scope of this PR. Will validate whether this is true in follow up PR. RIght now, we are just following how web search tool is doing.

crates/mcp/src/transform/transformer.rs
```diff
@@ -177,8 +265,60 @@ impl ResponseTransformer {
         }
     }
 
+    /// Transform MCP image generation results to OpenAI image_generation_call format.
+    fn to_image_generation_call(result: &Value, tool_call_id: &str) -> ResponseOutputItem {
+        let payload = Self::image_payload_from_wrapped_content(result)
+            .unwrap_or_else(|| Value::String(Self::flatten_mcp_output(result)));
```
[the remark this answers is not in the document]
TingtingZhou7: Right now I don't see mutiple text blocks whening testing using oci mcp server. Will evaluate this comment if we see format like this.

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -33,6 +33,20 @@ pub(crate) async fn ensure_mcp_connection(
     mcp_orchestrator: &Arc<McpOrchestrator>,
     tools: Option<&[ResponseTool]>,
 ) -> Result<(bool, Vec<McpServerBinding>), Response> {
+    // grpc Responses currently does not support image_generation tool calls.
+    if tools
+        .map(|t| {
+            t.iter()
+                .any(|tool| matches!(tool, ResponseTool::ImageGeneration(_)))
+        })
```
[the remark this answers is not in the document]
TingtingZhou7: I need to confirm with the team whether cases like this is supported now. If it's a valid case, we can same check later.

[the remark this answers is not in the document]
TingtingZhou7: I need to confirm with the team whether cases like this is supported now. If it's a valid case, we can same check later.

crates/mcp/src/transform/transformer.rs
```diff
@@ -26,15 +29,126 @@ pub fn mcp_response_item_id(source_id: &str) -> String {
...
+                    .filter(|o| o.get("type").and_then(|v| v.as_str()) == Some("text"))
+                    .and_then(|o| o.get("text"))
+                    .and_then(|v| v.as_str())
+                    .map(|text| match serde_json::from_str::<Value>(text) {
+                        Ok(parsed) => parsed
+                            .as_object()
+                            .is_some_and(ResponseTransformer::is_image_payload_candidate)
+                            .then_some(false)
+                            .unwrap_or(false),
+                        Err(_) => true,
```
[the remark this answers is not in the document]
TingtingZhou7: tested with non-streaming case. only one element in the array for both succeed/failed case. If this is a valid concern for streaming case, I will test and fix in a seperate pr.

model_gateway/src/routers/tool_output_context.rs
```diff
@@ -0,0 +1,36 @@
...
+/// string outputs are preserved as-is, non-strings use `output.to_string()`.
+pub fn compact_tool_output_for_model_context(
+    response_format: &ResponseFormat,
+    is_error: bool,
+    output: &Value,
+) -> String {
+    match response_format {
+        ResponseFormat::ImageGenerationCall => {
+            let note = if is_error {
+                extract_image_generation_fallback_text(output).unwrap_or_default()
```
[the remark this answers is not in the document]
TingtingZhou7: Not applied to us. OCI MCP server does not have such payload.

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -256,6 +266,58 @@ pub(crate) fn prepare_mcp_tools_as_functions(payload: &mut Value, session: &McpT
...
+) {
+    if let (Some(overrides), Some(args_obj)) = (
+        request_tool_overrides(response_format, original_body),
+        arguments.as_object_mut(),
+    ) {
+        let Some(override_obj) = overrides.as_object() else {
+            return;
+        };
+        for (k, v) in override_obj {
+            args_obj.insert(k.clone(), v.clone());
```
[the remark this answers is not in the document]
TingtingZhou7: Make sense. Will delete the extra field from request for now.

model_gateway/src/routers/grpc/harmony/builder.rs
CatherineSue: Please address this in a follow up PR. If Harmony should not support image generation, there are cleaner compile-safe options than “pretend it is a supported Harmony tool type here, then reject later in utils”.

For example,

```
ResponseTool::ImageGeneration(_) => continue
```

Or we could add a image generation mcp server, even a fake one so we can test this can work on harmony or other models.

model_gateway/src/routers/grpc/common/responses/streaming.rs
CatherineSue: Same here. We can address this in a follow up PR.

crates/protocols/src/responses.rs
```diff
@@ -80,6 +84,19 @@ pub struct CodeInterpreterTool {
     pub container: Option<Value>,
 }
 
+/// Built-in image generation tool.
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Deserialize, Serialize, Default, schemars::JsonSchema)]
+pub struct ImageGenerationTool {
```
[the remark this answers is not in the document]
TingtingZhou7: Will fix in next pr. Will keep this PR untouched now.

model_gateway/src/routers/grpc/harmony/builder.rs
TingtingZhou7: ACK. Will address in follow up pr! Thanks!

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -58,6 +82,20 @@ pub(crate) async fn ensure_mcp_connection(
     if let Some(tools) = tools {
         match ensure_request_mcp_client(mcp_orchestrator, tools).await {
             Some(mcp_servers) => {
+                if has_explicit_mcp_tools {
+                    let server_keys: Vec<String> =
+                        mcp_servers.iter().map(|binding| binding.server_key.clone()).collect();
+                    let has_image_generation_response_format = mcp_orchestrator
+                        .list_tools_for_servers(&server_keys)
```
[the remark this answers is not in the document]
TingtingZhou7: outdated comment.

slin1237 (commented):
Tingting — went through this carefully on a local checkout. Image gen feature itself looks reasonable end to end, the protocol/transformer/openai-router additions follow the existing pattern for builtin tools well, and the `extract_image_generation_fallback_text` work in the transformer is the kind of robustness that's easy to skip and you didn't.

The biggest concern isn't the image gen logic. It's that the gRPC gating introduced the same hardcoded-allow-list pattern in 4 different places, and one of them is a latent bug that will bite the day the rejection comes off. Inline comments cover the details — quick severity breakdown so you can navigate:

**Blocker — fix before merge**
- `grpc/common/responses/utils.rs:64` — `has_builtin_tools` omits ImageGeneration. Latent bug under the rejection. This is the one I'd most want fixed.
- `grpc/common/responses/utils.rs:39-51` — rejection in the wrong function, single-variant `matches!`, stale doc above.
- `grpc/harmony/builder.rs:100-107` — `is_builtin()` is both wrong and dead code (zero call sites in `model_gateway/src`).
- No tests for the new gRPC rejection. At minimum: a request with `image_generation` in tools returns 400 with code `unsupported_tool`.

**Major — should fix in this PR**
- `openai/mcp/tool_loop.rs:209` — silent serialization corruption (`unwrap_or_else` ships an empty object as a response item).
- `openai/mcp/tool_loop.rs:280-308` — `request_tool_overrides` name doesn't match its scope.
- `routers/tool_output_context.rs` — wrong location, same anti-pattern in micro, magic strings flowing into the model context.
- `grpc/harmony/builder.rs:428` — unreachable branch should be `unreachable!()`.

**Low — clean up if you're touching the file anyway**
- `utils.rs:40-68` — three separate iterations of the same tools vec.
- `utils.rs:48` — stringly-typed error code, no const, hardcoded tool name.
- `tool_loop.rs:310-326` — tool override merge precedence not documented or tested.

---

**The fix you actually want is per-model capability, not a route allow-list**

Route-level rejection ("grpc doesn't support image_gen") is the wrong axis. Same gRPC route can serve x-model (does image gen) and qwen-2.5-72b (doesn't), and a worker swap can change the answer overnight for the same model ID. The capability has to live with the model.

The shape that works:

1. `ResponseToolKind` enum + `ResponseTool::kind()` exhaustive match in `crates/protocols/src/responses.rs`. One place that classifies every tool variant, compiler-enforced.
2. `SupportedTools` bitset (`Copy`, `Default`, `Serialize`) — one bit per kind, with a `validate(&[ResponseTool])` helper.
3. Extend `ModelCard` with `supported_tools: SupportedTools`, populated through the existing label pipeline. Workers register with `smg.tool.image_generation: false`-style labels — same shape as the rest of the model card metadata.
4. Validate at every route entry handler: `model_card.supported_tools.validate(request.tools)`. One call site per route, runs before any MCP / harmony / transformer code.

If you also need a transport ceiling (e.g., grpc has a hard cap that's narrower than the model's max because of binary streaming limitations), define a `const SupportedTools` mask per transport and intersect with the model card's set. The image gen rejection then becomes one bit cleared in the grpc mask with a tracking issue in the comment, and the day it ships you flip one bit. No fan-out, no `matches!`, compile-time exhaustiveness at one place. Adding a new tool variant means updating the enum, and the compiler walks you through every site that needs a decision.

This is bigger than #1091 should carry — it's a follow-up refactor PR after this one lands.

---

**How to break this PR apart**

17 files / 4 different concerns / 1 latent bug is too much for one review. Splitting it four ways:

| PR | Files | Scope | Risk |
|---|---|---|---|
| **#1091a** | `crates/protocols/src/{responses,event_types}.rs` | Pure protocol additions: `ImageGenerationTool`, `ResponseOutputItem::ImageGenerationCall`, `ImageGenerationCallEvent`, `ItemType::ImageGenerationCall`. No behavior change. | None. 5-minute review, land first. |
| **#1091b** | `crates/mcp/*` (`core/config.rs`, `lib.rs`, `transform/{mod,types,transformer}.rs`) | MCP transformer for image gen, including the +253 line transformer hunk and `extract_image_generation_fallback_text`. Depends on #1091a. | Low. Pure addition, has tests. |
| **#1091c** | `model_gateway/src/routers/openai/*`, `routers/common/mcp_utils.rs`, `routers/tool_output_context.rs` (relocated next to its consumer), `routers/mod.rs`, openai `responses/streaming.rs` and `responses/utils.rs` | OpenAI router integration end-to-end. Depends on #1091b. The actual feature. | Medium. Touches the openai hot path, needs an end-to-end test. |
| **#1091d** | `model_gateway/src/routers/grpc/*` only | gRPC capability gate, refactored as a route-entry validator + the latent bug fix at `utils.rs:64` + the `is_builtin()` cleanup + the missing tests. Conceptually independent of #1091c — can land before or after. | Medium. This is where the bugs are. Do it last and carefully. |

Each split is 3-6 files with one clear scope and one obvious reviewer. The current 17-file PR has the reviewer spending most of the time chasing how things connect across crates instead of looking at the code.

If you want, split the protocol additions out today and I'll review and approve #1091a in five minutes — no-risk merge that unblocks the rest. The harder split is #1091d (gRPC), but that's also where the latent bug lives, so it's the one most worth doing carefully.

Nothing here is meant to slow you down. It's meant to make sure that when you delete the rejection in six weeks, image gen actually works on grpc instead of silently skipping MCP routing.

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -36,6 +36,20 @@ pub(crate) async fn ensure_mcp_connection(
     mcp_orchestrator: &Arc<McpOrchestrator>,
```
slin1237, line 55: This rejection is in the wrong function. `ensure_mcp_connection` is supposed to set up MCP connections, and `image_generation` isn't necessarily an MCP tool — gating capabilities here couples two unrelated concerns. The doc above (lines 28-34) is also now stale: it still says this function checks for MCP and `web_search_preview`/`code_interpreter` and doesn't mention any rejection.

Move the gate to a route-entry validator that runs before any MCP setup. Then this function keeps doing one thing and the doc stays accurate without needing an update.

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -36,6 +36,20 @@ pub(crate) async fn ensure_mcp_connection(
...
     tools: Option<&[ResponseTool]>,
 ) -> Result<(bool, Vec<McpServerBinding>), Response> {
+    // grpc Responses currently does not support image_generation tool calls.
+    if tools
+        .map(|t| {
+            t.iter()
+                .any(|tool| matches!(tool, ResponseTool::ImageGeneration(_)))
```
slin1237: Single-variant `matches!` won't catch you the next time someone adds a `ResponseTool` variant — tomorrow's `ResponseTool::TextToSpeech` slips through silently because the compiler has nothing to say.

The fix is `tool.kind() == ResponseToolKind::ImageGeneration` with `kind()` defined as one exhaustive `match` on `ResponseTool` in `crates/protocols/src/responses.rs`. The compiler then forces every new variant through that one match arm. Same fix retires the same anti-pattern at `harmony/builder.rs:105` and `utils.rs:64` below — one structural change instead of three scattered band-aids.

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -36,6 +36,20 @@ pub(crate) async fn ensure_mcp_connection(
...
+        })
+        .unwrap_or(false)
+    {
+        return Err(error::bad_request(
+            "unsupported_tool",
```
slin1237: `"unsupported_tool"` and the message string are both magic. No const, no enum. Two consequences: if another route uses a slightly different spelling for the same condition, clients can't disambiguate, and the message hardcodes `image_generation`, which lies the day `ResponseTool::ImageGeneration` gets renamed.

Promote the error code to a const. Derive the tool name from `tool.kind().as_str()`. Bonus: include the list of supported tools in the error body so clients can self-correct without grepping logs.

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -36,6 +36,20 @@ pub(crate) async fn ensure_mcp_connection(
...
+            "image_generation tool is not supported on grpc responses routes.",
+        ));
+    }
+
     // Check for explicit MCP tools (must error if connection fails)
     let has_explicit_mcp_tools = tools
         .map(|t| t.iter().any(|tool| matches!(tool, ResponseTool::Mcp(_))))
```
slin1237: Two things in this region of the function — the first is the most important thing in the PR.

The latent bug: `has_builtin_tools` 9 lines below this checks `WebSearchPreview | CodeInterpreter` only — ImageGeneration is missing. Today that's harmless because the rejection 16 lines up kills the request first. The day you delete the rejection (when image_gen is "fully tested"), this branch will silently mis-classify image_gen as not-a-builtin-tool, fall through the early-return at line 71, and skip MCP routing entirely — even when an MCP image-gen server is configured. You're shipping a latent bug under a temporary safety net. Fix it in this PR, or move the rejection somewhere it can never drift away from this check.

The cleanup: this function does `tools.iter()` three separate times for three different classifications (image_gen, mcp, builtin). One classification pass into a struct prevents the next addition from being copy-pasted as a fourth walk:

```rust
struct ToolSummary { has_image_gen: bool, has_mcp: bool, has_builtin: bool }
let summary = ToolSummary::classify(tools);
```

model_gateway/src/routers/grpc/harmony/builder.rs
slin1237: Two issues with how this file handles ImageGeneration.

`is_builtin()` at line 105 (above this hunk) is wrong. It says ImageGeneration isn't a builtin tool, but `BuiltinToolType::ImageGeneration` exists in `crates/mcp/src/core/config.rs:303` and `is_builtin_tool_call()` in `crates/protocols/src/event_types.rs:343` both include it. So this method now disagrees with the rest of the codebase about what "builtin" means. The compiler didn't catch it because `matches!` is non-exhaustive — same shape of bug as `grpc/common/responses/utils.rs:43`.

It's also dead code. Zero call sites in `model_gateway/src` — verified with `grep -rn '\.is_builtin()' model_gateway/src` (returns nothing). So you have a dead trait method that's also wrong. Pick one: delete it, or rewrite as exhaustive `match`.

Separately, the branch you just added on this line is unreachable in production — `ensure_mcp_connection` rejects image_generation before harmony ever runs. Returning `"image_generation"` here makes it look like harmony has a plan for this tool when it doesn't. Replace with `unreachable!("image_generation rejected upstream by ensure_mcp_connection")`. Documents the invariant, fails loud if someone removes the rejection without revisiting harmony.

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -196,7 +201,11 @@ pub(crate) async fn execute_streaming_tool_calls(
             },
         );
 
-        let output_str = tool_output.output.to_string();
+        let model_context_output = compact_tool_output_for_model_context(
+            &response_format,
+            tool_output.is_error,
+            &tool_output.output,
+        );
         let mut mcp_call_item = to_value(tool_output.to_response_item()).unwrap_or_else(|e| {
```
slin1237, line 258: Silent corruption. If `to_value(...)` fails, this hands back `json!({})`, the code below inserts an `id` into the empty object, and ships `{"id": "..."}` as the response item. The client sees a malformed item with no other fields instead of any indication something went wrong.

Decide what failed serialization should do here. Either bubble the error up — the tool effectively failed, treat it as such — or build a well-formed error item with `status: "failed"` and a note about the conversion problem. The `warn!` log isn't enough; by the time anyone reads it, the wrong response has already gone out.

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -264,6 +273,58 @@ pub(crate) fn prepare_mcp_tools_as_functions(payload: &mut Value, session: &McpT
...
 
+/// Extract request-level builtin tool overrides to merge into tool-call arguments.
+///
+/// Currently this is intentionally scoped to image generation only.
+/// We can extend this to other builtin tools later if needed.
+fn request_tool_overrides(
+    response_format: &ResponseFormat,
+    original_body: &ResponsesRequest,
+) -> Option<Value> {
+    if !matches!(response_format, ResponseFormat::ImageGenerationCall) {
```
slin1237: The function name promises something general — `request_tool_overrides` — but the body early-returns for everything that isn't `ImageGenerationCall`, with a doc comment saying "intentionally scoped to image generation only". Pick one. Either rename to `image_generation_request_overrides` and drop the `response_format` parameter (it's load-bearing for one branch), or build the mechanism generically so the next builtin tool is a one-line entry rather than another early-return.

The current shape wastes a future contributor's afternoon when they grep for `request_tool_overrides` expecting something polymorphic and find a function that's secretly single-purpose. Same anti-pattern as the grpc rejection in micro: a function pretending to dispatch but actually switching on one variant.

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -264,6 +273,58 @@ pub(crate) fn prepare_mcp_tools_as_functions(payload: &mut Value, session: &McpT
...
+    arguments: &mut Value,
+) {
+    if let (Some(overrides), Some(args_obj)) = (
+        request_tool_overrides(response_format, original_body),
+        arguments.as_object_mut(),
+    ) {
+        let Some(override_obj) = overrides.as_object() else {
+            return;
+        };
+        for (k, v) in override_obj {
```
slin1237: `HashMap::insert` overwrites unconditionally. So if the model issues a tool call with `size: "1024x1024"` and the request specified `size: "512x512"`, the request value wins.

That might be the right priority — request-level config beats model choice — but it's not documented and there's no test. The opposite priority (model knows best, request fills in gaps) is equally defensible. A doc comment on `apply_request_tool_overrides` pinning down which way it goes plus a unit test would prevent someone flipping it accidentally next quarter when they're cleaning up the merge logic.

model_gateway/src/routers/tool_output_context.rs
```diff
@@ -0,0 +1,36 @@
...
+/// minimal fixed summary (tool/status/note) to avoid feeding large binary
+/// image data back into the next model turn. Other formats are currently no-op:
+/// string outputs are preserved as-is, non-strings use `output.to_string()`.
+pub fn compact_tool_output_for_model_context(
+    response_format: &ResponseFormat,
+    is_error: bool,
+    output: &Value,
+) -> String {
+    match response_format {
+        ResponseFormat::ImageGenerationCall => {
```
slin1237: Three things in this small file. Worth fixing all of them now since it's brand new and there's no migration cost.

The location is wrong. This file lives at the top of `routers/`, sibling of `grpc/` and `openai/`, but the only consumer is `routers/openai/mcp/tool_loop.rs:204`. Two cleaner places: move it next to its consumer, or — better — make it a method on `ResponseFormat` in `crates/mcp/src/transform/types.rs` so each variant defines its own compaction policy at the variant site, right next to where the variant is defined.

It also repeats the same anti-pattern as the grpc rejection in micro. `match { ImageGenerationCall => special; _ => no-op }` — one real branch, default no-op, no exhaustive enforcement. As more tools need compaction, this file grows linearly with hardcoded branches and the default arm hides the question "does this tool need compaction?" from the compiler. A trait method on `ResponseFormat` would force a compile-time decision at every variant.

And the magic strings flow into the model's next-turn context. `"Successfully generated the image"` on line 21 and the `unwrap_or_default()` on line 19 (which gives you an empty string when there's no fallback) both end up in the JSON `note` field that the model reads. Promote them to named consts at the top of the file with a doc comment that says "this is part of the model prompt — changes affect model behavior". And replace the empty-string failure path with a sensible default like `"Image generation failed; no error details available"` — an empty note tells the model nothing.

slin1237 (changes requested):

model_gateway/src/routers/grpc/common/responses/utils.rs
TingtingZhou7: Moved this rejection under model_gateway/src/routers/grpc/router.rs in commit https://github.com/lightseekorg/smg/pull/1091/changes/edd6146f66c01daa289e5e3a5a3926a41d4e3db9#diff-f58a9a1a3c50bc6f3b056e82e59f1713986fd2197a652131ac0be98fa31e929e

model_gateway/src/routers/tool_output_context.rs
TingtingZhou7: 1. Moved this method compact_tool_output_for_model_context to crates/mcp/src/transform/types.rs.
2. Fixed by force exhaustive handling of all tools in one place. Changes in https://github.com/lightseekorg/smg/pull/1091/changes/141560dd6f319fb8b645520f88a8cc74cf2bcd7d#diff-952b664df2e0327e678bc486d2a43cdc19d2c2c88e99a3315f00788ed115a0bf
3. Fixed when replying on @CatherineSue's comment. The latest compact method remove the result field from the json object to keep with dp logic we have right now (Search for sanitizeOutputs method in genai-dp code base) - https://github.com/lightseekorg/smg/pull/1091/changes/141560dd6f319fb8b645520f88a8cc74cf2bcd7d#diff-a0211844de8da3fa7e723ca40ad0c4d2bcdf17b3039959546d168d7e9dd18e91

model_gateway/src/routers/openai/mcp/tool_loop.rs
TingtingZhou7: Not in scope in this PR. I did not make any change to line 209. Please ping the original code contributor to fix and not adding scope to this pr.

model_gateway/src/routers/grpc/harmony/builder.rs
TingtingZhou7: 1.2. Fixed by delete the is_builtin method - 381a17feac966611de63b9596fa9f9637b1a797b
3. Done. And add  
```#[expect(
    539 +            clippy::unreachable,
    540 +            reason = "image generation is rejected upstream by ensure_mcp_connection"
    541 +        )] 
```
to bypass the format check.

model_gateway/src/routers/grpc/common/responses/utils.rs
TingtingZhou7: Fixed.  Add ImageGeneration also to has_builtin_tools in [765d79bac4c0415b9b2b2530f64719386bd2ba15](https://github.com/lightseekorg/smg/pull/1091/changes/edd6146f66c01daa289e5e3a5a3926a41d4e3db9#diff-0885c1bdaee33f34d6f225221f586eddb411e1fe6e5f994d072eeadc01c91dbd)

model_gateway/src/routers/openai/mcp/tool_loop.rs
TingtingZhou7: Fixed under  model_gateway/src/routers/common/tool_overrides.r in https://github.com/lightseekorg/smg/pull/1091/changes#diff-22e8cba26b31471a6948fa63a38211df7ad12269cde22d3cf3233ebdb72c9d6c to make it more generic to other tools. Any new tools added in the future requires a one-line change here. 

Image

model_gateway/src/routers/grpc/common/responses/utils.rs
TingtingZhou7: Fixed in [765d79bac4c0415b9b2b2530f64719386bd2ba15](https://github.com/lightseekorg/smg/pull/1091/changes/edd6146f66c01daa289e5e3a5a3926a41d4e3db9#diff-0885c1bdaee33f34d6f225221f586eddb411e1fe6e5f994d072eeadc01c91dbd) by add error code to const and add supported tools in the body.

model_gateway/src/routers/grpc/common/responses/utils.rs
TingtingZhou7: Added a ResponseToolKind in crates/protocols/src/responses.rs in https://github.com/lightseekorg/smg/pull/1091/changes/edd6146f66c01daa289e5e3a5a3926a41d4e3db9#diff-c3029f8c0f800d9ebe9cfe6b6ba81d54386beec8000da123e1477126d77fb0c0

model_gateway/src/routers/openai/mcp/tool_loop.rs
TingtingZhou7: Updated with comment and unit test. 
https://github.com/lightseekorg/smg/pull/1091/changes#diff-22e8cba26b31471a6948fa63a38211df7ad12269cde22d3cf3233ebdb72c9d6cR19-R20

model_gateway/src/routers/grpc/harmony/builder.rs
```diff
@@ -425,6 +407,7 @@ impl HarmonyBuilder {
...
```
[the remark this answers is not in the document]
TingtingZhou7: fixed

crates/mcp/src/transform/types.rs
```diff
@@ -15,16 +17,47 @@ pub enum ResponseFormat {
...
+impl ResponseFormat {
+    /// Compact tool output before feeding it back into model context.
+    pub fn compact_tool_output_for_model_context(&self, is_error: bool, output: &Value) -> String {
+        if matches!(self, ResponseFormat::ImageGenerationCall) {
+            if is_error {
+                if let Some(text) = extract_image_generation_fallback_text(output) {
+                    return text;
+                }
+            } else {
+                return "Successfully generate image".to_string();
```
CatherineSue: Is this a standard in OpenAI format? The same hardcoded string exists in the original PR: https://github.com/lightseekorg/smg/pull/1057/changes#diff-6b25b656ee0b91d20ef43114ad9d8c09a9adacef6a4dfa668704106b7c572dfdR20-R24

crates/mcp/src/transform/transformer.rs
```diff
@@ -177,8 +253,53 @@ impl ResponseTransformer {
         }
     }
 
+    /// Transform MCP image generation results to OpenAI image_generation_call format.
+    fn to_image_generation_call(result: &Value, tool_call_id: &str) -> ResponseOutputItem {
```
CatherineSue, line 302: Nit: the whole section in this file all do the same `.get().as_str().to_string()` dance.

This is non-blocker but I'd suggest to derive `#[derive(Deserialize, Default)]` on a private `ImagePayload` struct and call `serde_json::from_value(...)` once — one line, type‑checked, all fields handled

crates/mcp/src/transform/types.rs
CatherineSue: The grammar is also wrong. `generated`?

crates/mcp/src/transform/types.rs
```diff
@@ -15,16 +17,47 @@ pub enum ResponseFormat {
...
+            }
+        }
+        if is_error {
+            match output {
+                Value::String(text) => text.clone(),
+                _ => serde_json::to_string(output).unwrap_or_else(|e| {
+                    format!("{{\"error\": \"Failed to serialize tool output: {e}\"}}")
```
CatherineSue: Is the `|e|` here reachable? `serde_json::to_string(v)` on an already‑parsed `Value` is infallible.

And I am not sure if the raw `\"` works if `e` ever has a quote, then the json would be invalid.

crates/mcp/src/transform/types.rs
```diff
@@ -15,16 +17,47 @@ pub enum ResponseFormat {
...
+                }),
+            }
+        } else {
+            match output {
+                Value::String(text) => text.clone(),
+                _ => output.to_string(),
+            }
+        }
+    }
+}
```
CatherineSue, line 85: Can we simply do 

```
match output {                                                                                                                                                                                                                          
          Value::String(s) => s.clone(),                                                                                                                                                                                                      
          _ => output.to_string(),                                                                                                                                                                                                            
      }   
```

The above two branch is doing the exact same thing.

model_gateway/src/routers/common/tool_overrides.rs
```diff
@@ -0,0 +1,117 @@
+use openai_protocol::responses::{ResponseTool, ResponsesRequest};
```
CatherineSue, line 1: Is it cleaner to move this inside `McpToolSession`? 
Because right now this `apply_request_tool_overrides` is scattered everywhere. And if we ever gonna support Anthropic messages API v2, which also has mcp servers, this function will need to be added there as well but right now there are no ways to enforce that, not in complier, not in runtime. 

Since we need to override it every turn in Mcp tool loop to pin the user requested fields, it makes more sense to add this in `McpToolSession`.

```
pub struct McpToolSession<'a> {                                                   
      orchestrator: &'a McpOrchestrator,                                                                                                                                                                                                      
      request_ctx: McpRequestContext<'a>,                  
      all_mcp_servers: Vec<McpServerBinding>,                                                                                                                                                                                                 
      mcp_servers: Vec<McpServerBinding>,                   
      mcp_tools: Vec<ToolEntry>,                                                                                                                                                                                                              
      exposed_name_map: HashMap<String, ExposedToolBinding>,                                                                                                                                                                                  
      exposed_name_by_qualified: HashMap<QualifiedToolName, String>,                                                                                                                                                                          
      internal_server_keys: HashSet<String>,                                                                                                                                                                                                  
      // ── NEW ──                                                                                                                                                                                                                            
      /// Caller-level pins from the Responses request, keyed by response format.                                                                                                                                                             
      /// Merged over model-emitted arguments inside `execute_tool`.                                                                                                                                                                          
      request_tool_overrides: HashMap<ResponseFormat, Value>,                                                                                                                                                                                 
  }              
```

Or if we worry about memory occupation, we can even do `image_generation_pins: Option<Value>,` instead to make this totally optional for non-image-generation tools.

In this way, we don't need this file. We move all the changes to `session.rs`.
It may require us to make `ResponseFormat` to derive `Eq, Hash` 

Then all the current 5 (or more) call sites which has 

```
let response_format = session.tool_response_format(&tc.function.name);
  apply_request_tool_overrides(&response_format, original_request, &mut args);   // ← gone                                                                                                                                                    
  ToolExecutionInput {                                      
      call_id: tc.id.clone(),                                                                                                                                                                                                                 
      tool_name: tc.function.name.clone(),          
      arguments: args,                                                                                                                                                                                                                        
  }         
```

becomes

```
ToolExecutionInput {                                                                                                                                                                                                                        
      call_id: tc.id.clone(),        
      tool_name: tc.function.name.clone(),                                                                                                                                                                                                    
      arguments: args,                                                                                                                                                                                                                        
  } 
```

Only thing we need is to update the `session` constructive calls.

Essentially it would make the PR smaller. This is a strong improvement. If not shipping in this PR, we can do a follow up.

CatherineSue (commented):
1. Can we update the PR description to include the latest status of the PR? As it supports gRPC now. Please try to follow PULL_REQUEST_TEMPLATE.md
2. The PR has a nice improvement on `compact_tool_output_for_model_context`. And a nice improvement comparing to the previous PR is there is no hardcoded `generation_image` for tool name anymore.
3. There are 2 non-blocker comments but they should be addressed soon.
4. My biggest concern is this the hardcoded `"Successfully generate image"`

crates/mcp/src/transform/types.rs
TingtingZhou7: Good point. There is no public doc that how openAI santisize data but I just found how we did that in our dp - if you search for sanitizeOutputs() in our dataplane codebase. I will update this logic to keep it same with dp.

TingtingZhou7: Here is the log that shows the result after compact, the base64 is removed but other fields like output_compression, etc are still preserved. 

Image


Changes are included in this commit - https://github.com/lightseekorg/smg/pull/1091/changes/3790345de34ff044000ced5e27aab5b5b903fcaa

crates/mcp/src/transform/types.rs
```diff
@@ -15,16 +17,56 @@ pub enum ResponseFormat {
...
+                format!("{{\"error\": \"Failed to serialize tool output: {e}\"}}")
+            })
+        } else {
+            match output {
+                Value::String(text) => text.clone(),
+                _ => output.to_string(),
+            }
+        };
+
+        debug!(
```
TingtingZhou7, line 76: Will delete this debug log before merege.

model_gateway/src/routers/grpc/harmony/responses/non_streaming.rs
```diff
@@ -412,8 +412,9 @@ fn build_tool_response(
 
     // Add MCP tool calls WITH output (these were executed)
     for (tool_call, result) in mcp_tool_calls.iter().zip(mcp_results.iter()) {
-        let output_str = to_string(&result.output)
-            .unwrap_or_else(|e| format!("{{\"error\": \"Failed to serialize tool output: {e}\"}}"));
+        let output_str = result
+            .response_format
+            .compact_tool_output_for_model_context(result.is_error, &result.output);
```
CatherineSue: I don't see how is_error is needed here. The original path is directly using to_string and catch the error there instead.

crates/mcp/src/transform/types.rs
CatherineSue: Please apply the changes as I mentioned above. The current three branches are exactly doing the same thing. You can check `ToolResult.is_error`, not how it is supposed to be used here. You can also ask code agent if you have doubts.

model_gateway/src/routers/common/tool_overrides.rs
CatherineSue: If decided to address this in a next PR, please acknowledge here. This has to be addressed but I want to move this PR forward first and leave refactors next

crates/mcp/src/transform/transformer.rs
CatherineSue: Same here. Please acknowledge if this is gonna be addressed in a next PR.

CatherineSue (commented):
Please update PR description. Thanks.

crates/mcp/src/transform/types.rs
TingtingZhou7: Fixed in https://github.com/lightseekorg/smg/pull/1091/changes/3790345de34ff044000ced5e27aab5b5b903fcaa.

crates/mcp/src/transform/types.rs
TingtingZhou7: I have removed the is_error from this method and right now only two conditions for this method. Latest changes are in https://github.com/lightseekorg/smg/pull/1091/changes/3790345de34ff044000ced5e27aab5b5b903fcaa. 

So right now the logic is: 

- If the tool is image-gen, we just drop the result no matter whether it's error or not. (Keep align with dp) 
- If the tool is not image-gen, we return 
```
- serde_json::to_string(output).unwrap_or_else(|e| {
                format!("{{\"error\": \"Failed to serialize tool output: {e}\"}}")
            })
```

And this is what currently have in main branch for all other tools. 

Image

crates/mcp/src/transform/transformer.rs
TingtingZhou7: Good point. I fixed in this commit - https://github.com/lightseekorg/smg/pull/1091/changes/f602f9d90c6ea5ceeb4ab9b5c839493e6918aa9d

TingtingZhou7: And here is the test result after this change. Output of image_gen_call is formatted correctly;

[smg-new.json](https://github.com/user-attachments/files/26841421/smg-new.json)

model_gateway/src/routers/common/tool_overrides.rs
TingtingZhou7: Ack, this might be a big change, i can clean up in next PR>

crates/mcp/src/transform/types.rs
CatherineSue: I said I think this is not needed. You already have `output: &Value` in the input, how could `serde_json::to_string` return error? If the original main has a bug, we can fix it here right now.

model_gateway/src/routers/grpc/harmony/responses/non_streaming.rs
TingtingZhou7: is_error is deleted in https://github.com/lightseekorg/smg/pull/1091/commits.

crates/mcp/src/transform/types.rs
TingtingZhou7: Got your point. Removed the exception, the output here should alwasy be json format. It's defined as Value before: 

Image

TingtingZhou7: Updated in https://github.com/lightseekorg/smg/pull/1091/changes/f602f9d90c6ea5ceeb4ab9b5c839493e6918aa9d

model_gateway/src/routers/grpc/common/responses/utils.rs
TingtingZhou7: Here is how the new error response looks like: 

```
    }' | tee ~/Desktop/smg-new.json
{"error":{"type":"Not Found","code":"model_not_found","message":"No worker available for model 'gpt-5.1'","param":null}}%                                                                                       tingzhou@tingzhou-mac workspace % curl -sS "http://127.0.0.1:9999/v1/responses" \
    -H "Authorization: Bearer $OPENAI_API_KEY" \
    -H "opc-request-id: $(uuidgen)" \
    -H "opc-compartment-id: ocid1.compartment.oc1..aaaaaaaathgntpo75bdehisnl6wkxfc4slkd6rpheafbt5a6ekm2ri4bmeva" \
    -d '{
      "model":"openai/gpt-oss-20b",
      "input":"Generate an image of a red panda reading a book under a maple tree, watercolor style.",
      "tools":[
        {
          "type":"image_generation",
          "model":"openai.gpt-image-1.5",
          "quality":"high",
          "output_format":"png",
          "output_compression":100,
          "size":"1024x1024",
          "background":"auto"
        }
      ],
      "stream":false
    }' | tee ~/Desktop/smg-new.json
{"error":{"type":"Bad Request","code":"unsupported_tool","message":"Tool 'image_generation' is not supported for gRPC Responses routes. Supported tools: function, web_search_preview, code_interpreter, mcp.","param":null}}%                                                                                                                                                                                                  tingzhou@tingzhou-mac workspace % 

```

model_gateway/src/routers/openai/mcp/tool_loop.rs
```diff
@@ -805,39 +764,16 @@ fn build_incomplete_response(
...
 
-    // Convert MCP function_call items in output to mcp_call format.
+    // Convert any function_call in output to mcp_call format
     if let Some(output_array) = obj.get_mut("output").and_then(|v| v.as_array_mut()) {
         // Find any function_call items and convert them to mcp_call (incomplete)
         let mut incomplete_items = Vec::new();
         for item in output_array.iter() {
             let item_type = item.get("type").and_then(|t| t.as_str());
             if item_type.is_some_and(is_function_call_type) {
                 let tool_name = item.get("name").and_then(|v| v.as_str()).unwrap_or("");
```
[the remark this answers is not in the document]
TingtingZhou7: not in scope. No code change here.

model_gateway/src/routers/grpc/harmony/builder.rs
TingtingZhou7: Changed to use unreachable - https://github.com/lightseekorg/smg/pull/1091/changes/52cbda6c1de9956b12958ca430cbde44c63ecebc

model_gateway/src/routers/grpc/common/responses/streaming.rs
TingtingZhou7: It's hard to remove it since we use an exhaustive match. If we remove exhaustive match, do you see any concern here?

model_gateway/src/routers/openai/responses/streaming.rs
```diff
@@ -149,6 +150,9 @@ pub(super) fn apply_event_transformations_inplace(
...
```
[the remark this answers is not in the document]
TingtingZhou7: Right now, we are follow the same pattern as other tools. WIll evaliudate whether this is a bug and fix in follow up pr

model_gateway/src/routers/grpc/common/responses/streaming.rs
```diff
@@ -133,6 +136,7 @@ impl ResponseStreamEventEmitter {
                         Some("mcp_call")
                             | Some("web_search_call")
                             | Some("code_interpreter_call")
+                            | Some("image_generation_call")
                             | Some("file_search_call")
```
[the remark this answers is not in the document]
TingtingZhou7: grpc will be disabled , not applicable

model_gateway/src/routers/grpc/harmony/builder.rs
TingtingZhou7: Commit which include the change - https://github.com/lightseekorg/smg/pull/1091/changes/52cbda6c1de9956b12958ca430cbde44c63ecebc

model_gateway/src/routers/grpc/common/responses/utils.rs
```diff
@@ -25,10 +25,29 @@ use crate::{
     worker::WorkerRegistry,
 };
 
+const ERROR_CODE_UNSUPPORTED_TOOL: &str = "unsupported_tool";
+
+const GRPC_RESPONSES_SUPPORTED_TOOL_KINDS: &[ResponseToolKind] = &[
+    ResponseToolKind::Function,
+    ResponseToolKind::WebSearchPreview,
```
TingtingZhou7, line 32: I should remove this also to make it exclusive match, so a new tool added later won't be missed here

crates/mcp/src/transform/transformer.rs
```diff
@@ -27,8 +30,22 @@ pub fn mcp_response_item_id(source_id: &str) -> String {
...
+/// `[{"type":"text","text":"..."}]`.
+pub fn extract_image_generation_fallback_text(value: &Value) -> Option<String> {
+    value.as_array().and_then(|content| {
+        content.iter().find_map(|item| {
+            item.as_object()
+                .filter(|o| o.get("type").and_then(|v| v.as_str()) == Some("text"))
+                .and_then(|o| o.get("text"))
+                .and_then(|v| v.as_str())
+                .filter(|text| !text.trim().is_empty())
+                .map(str::to_string)
```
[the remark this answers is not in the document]
TingtingZhou7: valid point. I need to test case when mcp return multiple results. Noted down and will follow up on this. Should be good for this pr since it's a p2

Issue #10, opened 2026-01-15
Setup Self-Hosted H100 Runners


slin1237:
## Description
Add self-hosted runners with H100 GPUs for CI/CD.

## Actions
- [ ] Set up H100 machines with:
  - Ubuntu 22.04/24.04
  - CUDA toolkit
  - Docker
  - Rust toolchain
  - Required dependencies
- [ ] Install GitHub Actions runner on each machine
- [ ] Configure runner labels:
  - `self-hosted`
  - `H100`
  - `gpu`
  - Consider memory/GPU count labels
- [ ] Add machines to repository settings
- [ ] Update workflows to use new runner labels
- [ ] Test runner connectivity and capabilities
- [ ] Set up monitoring/alerting for runners

## Notes
This is a manual configuration task in GitHub repository settings.

CatherineSue:
Is this issue completed? @XinyueZhang369

PR #1118, opened 2026-04-14
refactor(worker): create event-driven WorkerMonitor (PR 8)

slin1237:
## Summary

Consolidates `LoadMonitor` (per-group polling loops in `worker/manager.rs`) and `WorkerLoadManager` (DP rank load cache in `worker/load.rs`) into a single `WorkerMonitor` coordinator that subscribes to `WorkerRegistry` events and reconciles per-group polling state on its own. Workflow steps no longer push group lifecycle notifications to the monitor — the registry's broadcast channel does it.

This is PR 8 of the worker module deep refactor (tracked in \`.claude/plans/2026-04-09-worker-module-deep-refactor.md\`). It lands on top of #1113 (PR 7.5 — registry method reorganization).

Closes the polling side of the lifecycle work introduced in #1100 / #1101 / #1102 / #1105 / #1113.

## What changed

**New file `model_gateway/src/worker/monitor.rs`**

- \`WorkerLoadManager\` moves here unchanged (pure DP cache, no I/O).
- \`WorkerMonitor::new()\` constructs the struct without spawning anything — the factory layer reads \`worker_load_manager\` immediately so policies that need the DP cache can be wired before any workers exist.
- \`WorkerMonitor::start_event_loop()\` subscribes to the registry, runs a synchronous bootstrap reconcile from the registry snapshot, and spawns the background event handler. The subscribe-before-reconcile order means any registration that lands during reconcile is replayed by the event task, matching \`WorkerManager::start\`.
- Event handler:
  - \`Registered\` / \`Removed\` / \`Replaced\` → reconcile every group the worker participates in (\`worker.models() × worker_type × connection_mode\`). Empty groups stop their loop and clear cached state; new groups start one with the per-worker \`load_monitor_interval_secs\` override (or the default); existing groups whose desired interval changed restart with the new cadence.
  - \`StatusChanged\` away from \`Ready\` → evict the worker's URL from both the watch-channel snapshot and the DP cache. Transitions *into* \`Ready\` are handled implicitly: the group's polling loop reads from the registry on every tick.
  - \`RecvError::Lagged(_)\` → \`reconcile_from_registry()\` stops every loop, clears the cached snapshot, and rebuilds polling state from the current registry contents.
- Polling loop now filters to \`status == Ready\` workers, so \`Pending\` / \`NotReady\` / \`Failed\` workers no longer contribute spurious load samples.
- \`fetch_http_load\` and \`fetch_grpc_load\` move here as \`pub(crate)\` helpers that \`WorkerManager::get_all_worker_loads\` continues to call.
- \`Drop\` aborts the event task and every per-group polling loop.

**Removed**

- \`model_gateway/src/worker/load.rs\` — \`WorkerLoadManager\` lives in \`monitor.rs\` now.
- \`model_gateway/src/worker/manager.rs\` — entire \`LoadMonitor\` struct, impl, and \`Drop\` are gone (~250 lines). \`get_all_worker_loads\` calls the new \`WorkerMonitor::fetch_*\` helpers.
- \`model_gateway/src/workflow/steps/shared/register.rs\` — the \`load_monitor.on_group_added(...)\` push notification block. WorkerMonitor picks the new group up via the \`WorkerEvent::Registered\` broadcast emitted by \`worker_registry.register()\` directly above.
- \`model_gateway/src/workflow/steps/local/remove_from_worker_registry.rs\` — symmetric removal of the \`load_monitor.on_group_removed(...)\` block.

**Wiring**

- \`model_gateway/src/worker/mod.rs\` — drops \`pub mod load\`, adds \`pub mod monitor\`, re-exports \`WorkerLoadManager\` and \`WorkerMonitor\` from \`monitor\`, drops the \`LoadMonitor\` re-export.
- \`model_gateway/src/app_context.rs\` — renames the \`load_monitor\` field/builder/method/import to \`worker_monitor\`. The \`with_load_monitor\` builder step becomes \`with_worker_monitor\` and constructs a \`WorkerMonitor\` (no background task yet).
- \`model_gateway/src/server.rs\` — replaces the old \"LoadMonitor groups start dynamically\" debug log with an explicit \`worker_monitor.start_event_loop()\` call placed after \`RouterManager::from_config\`, so the bootstrap reconcile captures every worker that exists at that point and the event task picks up everything registered afterwards.
- \`model_gateway/src/routers/factory.rs\` — \`MinimumTokensPolicy\` constructor reads \`worker_load_manager\` via \`ctx.worker_monitor\` instead of \`ctx.load_monitor\`.
- \`model_gateway/src/service_discovery.rs\` — renames the \`load_monitor: None\` literal in the test \`AppContext\`.
- Test scaffolding (\`tests/common/{mod,test_app}.rs\`, \`tests/wasm_test.rs\`, \`tests/routing/test_pd_routing.rs\`) — \`LoadMonitor\` → \`WorkerMonitor\` / \`load_monitor\` → \`worker_monitor\`.

## Why

The old design required the workflow layer to push group lifecycle events to LoadMonitor at the right moments. That coupled two otherwise-independent subsystems (worker state and load polling) and made it easy to forget a notification or to duplicate one. After the WorkerStatus / \`WorkerEvent\` work in PR 5–7.5, the registry already broadcasts the events the monitor needs, so the monitor can subscribe directly and the workflow layer no longer has to know it exists. This also matches \`WorkerManager\`'s structure (subscribe-and-reconcile with bootstrap + lag recovery).

## How

The implementation mirrors \`WorkerManager\`'s lifecycle pattern:

1. Synchronous bootstrap reconcile on the caller's thread.
2. Subscribe to the broadcast channel **before** the reconcile, so any registration that lands during reconcile is replayed.
3. Spawn the background event task.

Per-group reconcile is sync (no awaits) and protected by a \`parking_lot::Mutex\` so \`Drop\` can abort everything cleanly. The monitor exposes \`worker_load_manager\` as an \`Arc\` so factories can wire load-aware policies before \`start_event_loop()\` is called.

\`KvEventMonitor\` is intentionally untouched and remains driven by the workflow/policy path — its lifecycle is policy-ordered today, and a future \"policy lifecycle coordinator\" PR can fold it in.

## Test plan

- [x] \`cargo check -p smg --lib\` — clean
- [x] \`cargo check -p smg --tests\` — clean
- [x] \`cargo clippy -p smg --lib --tests -- -D warnings\` — clean
- [x] \`cargo fmt -p smg -- --check\` — clean
- [x] \`cargo test -p smg --lib worker::monitor\` — 8 passed, 0 failed (4 \`WorkerLoadManager\` cache tests + 4 \`WorkerMonitor\` lifecycle tests covering bootstrap reconcile, register event, remove event, and \`StatusChanged\` eviction)
- [x] \`cargo test -p smg --lib\` — 535 passed, 0 failed, 4 ignored
- [x] \`cargo test -p smg --tests\` — 16 integration binaries, 468 tests, 0 failed

<details>
<summary>Checklist</summary>

- [x] Documentation updated (module docs on \`monitor.rs\` describe lifecycle, group reconciliation, and polling rules)

</details>

## Summary by CodeRabbit

* **Chores**
  * Consolidated worker monitoring into a unified, event-driven monitor for consistent real-time load tracking.
  * Startup now activates the monitor event loop so routing reacts promptly to worker state changes.
  * Test harnesses updated to exercise and validate the new monitoring behavior during setup.

* **Bug Fixes**
  * Removed stale per-worker load cache sources and ensured cache eviction on worker removal or status changes.


slin1237:
Addressed the full review pass in 494548c8. Per-comment breakdown:

### 🔴 Critical / 🟠 Major

**`Removed` event must evict cached loads** — @claude (#1), @gemini-code-assist (#5), @coderabbitai (#10): all three converged on the same bug. Real bug, confirmed by reading the code: when the removed worker was the last in its group, `reconcile_group` stopped the polling loop, and the stopped loop never had a chance to prune the stale entry. Fix: the event loop now calls \`monitor.evict_worker_loads(worker.url())\` *before* \`reconcile_group\` for \`Removed\`. Added the regression test \`removed_event_evicts_cached_loads\` so this can't regress.

**`Replaced` event must evict the old URL** — @gemini-code-assist (#6): registry's \`replace()\` rejects URL changes, so \`old.url() == new.url()\`, but the worker's group set can shrink to empty (model list change). In that case the disappearing groups would leak the entry. Fix: same eviction-before-reconcile pattern, using \`old.url()\`. Surviving / new groups repopulate the entry on their next tick.

**`stop_all_groups` left DP cache stale** — @gemini-code-assist (#8), @coderabbitai (#9): real bug. After \`RecvError::Lagged\` triggered \`reconcile_from_registry()\`, stale per-rank entries survived the rebuild. Fix: added \`WorkerLoadManager::clear()\` and call it inside \`stop_all_groups()\`. Also dropped the \`if drained.is_empty() { return; }\` early return so the function always resets both caches and behaves as a proper "reset." Added the regression test \`stop_all_groups_clears_dp_cache\`.

**Test helpers don't start the event loop + wrong interval field** — @coderabbitai (outside-diff): real test fidelity issue. Tests that depend on event-driven group reconciliation were silently running against an inert monitor, and they passed \`worker_startup_check_interval_secs\` instead of \`load_monitor_interval_secs\`. Fix: every helper in \`tests/common/{mod,test_app}.rs\`, \`tests/wasm_test.rs\`, and \`tests/routing/test_pd_routing.rs\` now starts the event loop after the \`AppContext\` is built and uses the production interval field.

### 🟡 Medium / 🟢 Minor

**\`WorkerLoadManager::clear()\` API** — @gemini-code-assist (#7), @coderabbitai (#9): added. Used by \`stop_all_groups()\` above.

**\`evict_worker_loads\` allocates a String** — @claude (#2): real nit. \`std::slice::from_ref(&url.to_string())\` allocated just to satisfy a \`&[String]\` API. Fix: added \`WorkerLoadManager::remove_worker(url: &str)\` and updated \`evict_worker_loads\` to use it. Allocation-free now.

**\`stop_group\` doc claims it evicts but doesn't** — @claude (#3): real nit. Updated the doc to match the actual behavior and document that callers must invoke \`evict_worker_loads\` separately when they know the affected URLs.

**Stale comments referencing \`LoadMonitor\`** — @coderabbitai (outside-diff in \`register.rs\` and \`remove_from_worker_registry.rs\`): refreshed both. Also collapsed the dead \`urls_by_group: HashMap<_, Vec<String>>\` in \`remove_from_worker_registry.rs\` into a \`HashSet\` since the value side stopped being read after the load-monitor push call was removed in this PR.

### Verification (on 494548c8)

- \`cargo check -p smg --lib --tests\` — clean
- \`cargo clippy -p smg --lib --tests -- -D warnings\` — clean
- \`cargo fmt -p smg -- --check\` — clean
- \`cargo test -p smg --lib worker::monitor\` — **10 passed, 0 failed** (4 cache tests + 4 original lifecycle tests + 2 new regression tests)
- \`cargo test -p smg --lib\` — **537 passed, 0 failed, 4 ignored**
- \`cargo test -p smg --tests\` — 16 integration binaries, **468 tests, 0 failed**

slin1237:
Both new findings addressed in f9b003ca. Per-comment breakdown:

### 🟠 Major — Arc cycle in spawned tasks (`Weak<Self>` fix)

@coderabbitai: confirmed real bug. Trace:
- \`start_event_loop\` and \`spawn_group_loop\` cloned \`Arc<Self>\` into the spawned closures
- The \`JoinHandle\`s for those tasks were stored back inside \`WorkerMonitor\` (\`event_task\` / \`group_handles\`)
- That formed a cycle: \`monitor → JoinHandle → spawned task closure → Arc<monitor>\`
- When \`AppContext\` dropped its \`Arc<WorkerMonitor>\`, the strong count never reached zero because the spawned tasks still held their own \`Arc\` clones
- \`Drop\` was therefore unreachable except via forceful \`tokio\` runtime shutdown — fine for production server lifetime, but a real cycle in any scenario where the runtime outlives the monitor

**Fix:** both spawn sites now downgrade to \`Weak<Self>\`. The spawned tasks upgrade the \`Weak\` once per iteration (after \`events_rx.recv().await\` for the event loop, after \`interval_timer.tick().await\` for each group loop), do their work with the temporary \`Arc\`, and explicitly \`drop\` it before the next \`await\` so the strong count only ever blips up to \`1 + external\` for the duration of an iteration. If the upgrade fails the task exits immediately — exactly what we want when the monitor has been dropped.

**Regression test:** \`spawned_tasks_use_weak_references\` asserts \`Arc::strong_count(&monitor) == 1\` and \`Arc::weak_count(&monitor) >= 2\` (one Weak from the event task, one from the group loop) after \`start_event_loop\`. This locks in the cycle-free invariant.

### 🟠 Major — Empty fetch tick left stale loads in the watch snapshot

@coderabbitai: confirmed real bug. \`group_monitor_loop\`'s "all fetches failed" branch hit \`continue\` without removing the group's URLs from \`load_tx\`. A group that started timing out kept publishing its previous tick's loads forever — subscribers iterating the snapshot saw stale entries indefinitely.

**Fix:** hoisted the \`all_group_urls\` computation above the empty branch and prune the snapshot in both paths. Per your guidance, the DP cache (\`worker_load_manager\`) is intentionally **not** pruned on this path: routing decisions can fall back to last-known-good per-rank loads when an upstream is briefly unreachable, matching the existing load-aware policy contract.

### Verification (on f9b003ca)

- \`cargo check -p smg --lib\` — clean
- \`cargo clippy -p smg --lib --tests -- -D warnings\` — clean
- \`cargo fmt -p smg -- --check\` — clean
- \`cargo test -p smg --lib worker::monitor\` — **11 passed, 0 failed** (10 previous tests + the new \`spawned_tasks_use_weak_references\`)
- \`cargo test -p smg --lib\` — **538 passed, 0 failed, 4 ignored**

PR #1119, opened 2026-04-14
feat(protocols): add serde(flatten) to ChatCompletionRequest for engine-specific fields

### LINKED ISSUE #22740: sgl-model-gateway drops return_routed_experts field from chat/completion requests
## Problem

The sgl-model-gateway drops the `return_routed_experts` request parameter when proxying chat/completion requests to sglang workers. This prevents MoE routing replay (recording expert routing decisions during inference and replaying them during training) from working through the gateway.

## Root cause

The `ChatCompletionRequest` struct in the external [`openai-protocol`](https://crates.io/crates/openai-protocol) crate (source: [lightseekorg/smg](https://github.com/lightseekorg/smg)) does not have a catch-all for unknown fields. When the gateway deserializes incoming JSON into this typed struct, any fields not explicitly defined in the struct are silently dropped. When the struct is re-serialized and forwarded to the worker, `return_routed_experts` is missing.

Notably, `CompletionRequest` in the same crate already has `#[serde(flatten)] pub other: Map<String, Value>`, which preserves unknown fields. `ChatCompletionRequest` is simply missing this same pattern.

The sglang Python server already supports `return_routed_experts`:
- `serving_completions.py`: reads `request.return_routed_experts` and passes it to the scheduler
- `RoutedExpertsCapturer`: captures per-layer `topk_ids` during MoE forward pass
- Response includes `routed_experts` in `sgl_ext` on each choice

The response side works fine (the gateway forwards response bytes as-is), so only the request-side deserialization is the problem.

## Fix

The fix is upstream in the `openai-protocol` crate, not in this repo:

**PR: https://github.com/lightseekorg/smg/pull/1119**

This adds `#[serde(flatten)] pub other: Map<String, Value>` to `ChatCompletionRequest`, matching the existing pattern on `CompletionRequest`. Once that PR is merged and a new crate version is published, sglang will need to bump its `openai-protocol` dependency version in `sgl-model-gateway/Cargo.toml` (currently pinned at `=1.0.0`).

## Context

- Used for MoE expert routing replay during RL training
- The `sglang-miles` branch already has `enable_return_routed_experts` in `ServerArgs`
- Tested on LLM360's fork (llm360/sglang#1)

DavidBellamy:
## Summary

- `ChatCompletionRequest` silently drops request fields not explicitly defined in the struct. This prevents engine-specific parameters (e.g. sglang's `return_routed_experts` for MoE routing replay) from passing through the gateway to backend workers.
- `CompletionRequest` already has a `#[serde(flatten)] pub other: Map<String, Value>` catch-all. This PR makes `ChatCompletionRequest` consistent by adding the same pattern.
- All existing construction sites use `..Default::default()`, so the new field is automatically initialized to an empty map with no code changes needed elsewhere.

Fixes: https://github.com/sgl-project/sglang/issues/22740

## Summary by CodeRabbit

* **New Features**
  * Chat completion requests and responses now accept and include additional engine-specific JSON properties. This improves compatibility with varied chat engines, lets custom parameters pass through unchanged, and increases flexibility without altering existing validation or normalization behavior.

### DIFF
--- a/crates/protocols/src/chat.rs
+++ b/crates/protocols/src/chat.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
+use serde_json::{Map, Value};
 use validator::Validate;
 
 use super::{
@@ -316,6 +316,10 @@ pub struct ChatCompletionRequest {
 
     /// Random seed for sampling for deterministic outputs
     pub sampling_seed: Option<u64>,
+
+    /// Additional fields not explicitly defined above (e.g. engine-specific parameters)
+    #[serde(flatten)]
+    pub other: Map<String, Value>,
 }
 
 // ============================================================================


slin1237:
thanks for the fix, lgtm
please address ci failures

DavidBellamy:
@slin1237 Thanks for the review! I've addressed the CI failures:

- **DCO sign-off**: Added via rebase
- **PR title**: Updated to conventional commits format (`feat(protocols): ...`)

The new workflow runs need maintainer approval to execute (fork PR after force push). Could you approve them?

Re: the `openai-responses` failures from the previous run, all 5 were flaky MCP tool-calling tests (`test_mcp_basic_tool_call_streaming`, `test_mcp_multi_server_tool_call`, `test_mcp_multi_server_tool_call_streaming`, etc.) that failed after multiple retries. Unrelated to this one-line serde change. Should clear on re-run.

DavidBellamy:
Replaced by #1130 (branch renamed to conform to Mergify naming convention).

PR #1150, opened 2026-04-15
 Optimize TTFT in PD Disaggregation: Decouple prefill response and decode streaming

usernamehaha2022:
### Problem
This PR improves the TTFT of PD HTTP routing in the common streaming path.
In the previous implementation of `execute_dual_dispatch_internal`, the router waited for both prefill and decode HTTP responses before starting to proxy the decode stream. This adds avoidable latency to the first streamed token for the common case where:
* the request is streaming, and
* input logprobs are not requested.
Since input logprob merging is only needed when return_logprob=true, waiting for the prefill HTTP response in the no-logprob streaming path does not provide value on the critical path, but does delay stream startup.

The goal of this PR is to reduce that overhead and improve end-to-end streaming responsiveness in PD mode.

### Solution

This PR updates the PD HTTP router in pd_router.rs:

* Added a fast path in [execute_dual_dispatch_internal()](vscode-webview://03hgedb0gcobonia4pisqfi8qgriano10tbqlibain2qjqsl3mib/sgl-model-gateway/src/routers/http/pd_router.rs:533-794) for: context.is_stream == true context.return_logprob == false
* In this fast path: the prefill request is dispatched in the background; the router only waits for the decode response,
decode streaming is proxied to the client as soon as the decode side is ready.
* Kept the original behavior for: non-streaming requests and streaming requests with logprobs
* Preserved prefill response consumption in the background path so the HTTP connection can still be released back to the pool.
* Kept logprob-dependent behavior unchanged by continuing to use `process_prefill_response()` and the original dual-response path when return_logprob=true.

In short, this change narrows the optimization to the most common latency-sensitive streaming scenario while keeping the logprob path and non-streaming path behavior unchanged.

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Performance**
  * Optimized streaming request handling for improved response delivery times.

### DIFF
--- a/model_gateway/src/routers/http/pd_router.rs
+++ b/model_gateway/src/routers/http/pd_router.rs
@@ -626,110 +626,209 @@ impl PDRouter {
             false,
         );
 
-        // Send both requests concurrently and wait for both
-        // Note: Using borrowed references avoids heap allocation
+        // Send both requests concurrently
         events::RequestPDSentEvent {
             prefill_url: prefill.url(),
             decode_url: decode.url(),
         }
         .emit();
 
-        // Send both requests concurrently. Use try_join so that if either side
-        // hits a transport error, the other is cancelled immediately — otherwise
-        // the surviving request hangs waiting for a PD bootstrap that will never
-        // come (see #831).
-        let pd_result = tokio::try_join!(prefill_request.send(), decode_request.send());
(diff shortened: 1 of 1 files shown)


CatherineSue:
@usernamehaha2022 thanks for contributing. Could you fix the failed CI checks?

PR #1130, opened 2026-04-14
feat(protocols): add serde(flatten) to ChatCompletionRequest for engine-specific fields

DavidBellamy:
## Summary

Adds `#[serde(flatten)]` catch-all to `ChatCompletionRequest` so unknown engine-specific JSON properties (e.g. sglang's `return_routed_experts` for MoE routing replay) are preserved during deserialization and forwarded to backend workers rather than silently dropped.

`CompletionRequest` already has this flatten catch-all via its `other` field. This change makes `ChatCompletionRequest` consistent with that pattern.

Fixes #22740 (upstream SGLang issue).

## Note

This replaces #1119, which had a non-conforming branch name. All CI checks (DCO, conventional commit title, unit/integration/e2e tests) passed on the prior PR except for unrelated flaky MCP tool-calling tests.

PR #1160, opened 2026-04-15
feat(protocols): add serde(flatten) to chat response structs for engine-specific fields

DavidBellamy:
## Summary
Mirrors the request-side flatten pattern landed in #1130 for \`ChatCompletionRequest\`, extending \`#[serde(flatten)] pub other: Map<String, Value>\` to the response-side structs in \`crates/protocols/src/chat.rs\`:

- \`ChatCompletionResponse\`
- \`ChatCompletionMessage\`
- \`ChatChoice\`
- \`ChatCompletionStreamResponse\`
- \`ChatMessageDelta\`
- \`ChatStreamChoice\`

Needed so engine-specific response fields (e.g. SGLang's \`routed_experts\`, \`completion_token_ids\`) round-trip through the smg gateway without being silently dropped by serde's default unknown-field behavior on the response path.

This is the paired follow-on to #1130. Same pattern, same doc comment, same idiom.

## Summary by CodeRabbit

* **New Features**
  * Improved chat response compatibility by adding support for additional engine-specific and unmodeled JSON fields. Chat API responses with extra data will now be properly preserved during serialization and deserialization, enabling better integration with diverse chat platforms.

### DIFF
--- a/crates/protocols/src/chat.rs
+++ b/crates/protocols/src/chat.rs
@@ -669,6 +669,10 @@ pub struct ChatCompletionResponse {
     pub choices: Vec<ChatChoice>,
     pub usage: Option<Usage>,
     pub system_fingerprint: Option<String>,
+
+    /// Additional fields not explicitly defined above (e.g. engine-specific parameters)
+    #[serde(flatten)]
+    pub other: Map<String, Value>,
 }
 
 impl ChatCompletionResponse {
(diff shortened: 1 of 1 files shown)


CatherineSue (commented):
Can you include exactly which use case and what fields you need for each struct in the pr description?

We should be cautious on this instead of a big roll out.

DavidBellamy:
Re-opened from LLM360/smg head as PR #1239 so our internal deploy automation can pick up the branch.

PR #1231, opened 2026-04-17
fix(headers): drop content-length in preserve_response_headers

DavidBellamy:
## Problem

`preserve_response_headers` (model_gateway/src/routers/common/header_utils.rs:52-81) forwards upstream response headers with a hop-by-hop filter, but doesn't strip `content-length`. If a caller modifies the response body while preserving the upstream's `content-length` header, clients see `peer closed connection without sending complete message body` when the two sizes don't match.

This is a latent defensive bug today: the PD merge path (`pd_router.rs:955-999`) builds a fresh `(status, body).into_response()` without attaching upstream headers, so hyper recomputes Content-Length from the `Vec<u8>` body. But any future caller that copies the passthrough idiom (e.g., from lines 720-730) into a body-mutating path would trigger the mismatch.

Observed downstream: sglang PD + logprob merging returning Content-Length = decode_body.len() but body = serde_json::to_vec(merged) — 4-byte-short truncation on every request when the merge changes the `input_token_logprobs` shape (e.g., `null → []` or `[0.0] → [0]`).

## Fix

One line added to the hop-by-hop drop list:

```diff
         || name.eq_ignore_ascii_case("content-encoding")
+        || name.eq_ignore_ascii_case("content-length")
         || name.eq_ignore_ascii_case("host"))
```

This is standard reverse-proxy hygiene: equivalent to nginx's `proxy_hide_header Content-Length`. Lets hyper set Content-Length from the actual returned body size, eliminating the mismatch class entirely regardless of which path mutates the body.

## Test plan

1. Unit test against `preserve_response_headers` verifying `content-length` is dropped.
2. Integration test: stand up mock prefill + decode via existing `model_gateway/tests/common/mock_worker.rs`, issue a PD chat request with `logprobs=true`, assert the response's `content-length` equals the actual body length.

## Summary by CodeRabbit

* **Chores**
  * Standardized internal component versions to 1.0.0 across the workspace.

* **Bug Fixes**
  * Fixed header forwarding behavior to properly exclude the Content-Length header.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,19 +4,19 @@ resolver = "2"
 
 [workspace.dependencies]
 # Internal workspace crates
-openai-protocol = { version = "1.7.0", path = "crates/protocols" }
-reasoning-parser = { version = "1.2.3", path = "crates/reasoning_parser" }
-tool-parser = { version = "1.2.0", path = "crates/tool_parser" }
-wfaas = { version = "1.0.3", path = "crates/workflow" }
-llm-tokenizer = { version = "1.3.2", path = "crates/tokenizer" }
-smg-auth = { version = "1.1.1", path = "crates/auth" }
-smg-mcp = { version = "2.2.0", path = "crates/mcp" }
-kv-index = { version = "1.2.0", path = "crates/kv_index" }
-smg-data-connector = { version = "2.2.0", path = "crates/data_connector", package = "data-connector" }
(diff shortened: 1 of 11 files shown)


slin1237:
can you revert other changes
thanks

DavidBellamy:
Re-opened from LLM360/smg head as PR #1238 so our internal deploy automation can pick up the branch.

PR #1267, opened 2026-04-21
fix(protocol): honor OpenAI-style `reasoning: {enabled: bool}` on /v1/chat/completions

ConnorLi96:
OpenAI clients send

    {"reasoning": {"enabled": false}}

on `/v1/chat/completions` to explicitly skip thinking for reasoning- capable models.  SMG had no matching struct field (the existing `reasoning_effort: Option<String>` only covers the sibling `effort` key), so serde silently dropped the object.

Downstream, the Kimi-K2.5 chat template defaults to injecting `<think>` and the model kept producing thinking tokens on traffic that had explicitly disabled reasoning.  This was the root cause of the SMG+TRT vs TGL output-sequence-length gap flagged during shadow traffic investigation (Peter Kim, OSL thread).

```rust
pub struct ChatCompletionRequest {
    pub reasoning_effort: Option<String>,  // only the effort string
    // ...no `reasoning: {enabled}` field...
}
```

With `#[serde(deny_unknown_fields)]` off (we keep OpenAI-compatible forward compatibility) the entire `reasoning` object was discarded.

- Add a new `ReasoningConfig { enabled: bool, effort: Option<String> }` type in `chat.rs`, colocated with its only use site.
- Add `reasoning: Option<ReasoningConfig>` on `ChatCompletionRequest`.
- In `Normalizable::normalize`, migrate it into `chat_template_kwargs` (`enable_thinking` / `thinking`).  `.or_insert()` leaves any value operators have placed in `chat_template_kwargs` directly untouched.

- Only the OpenAI Chat Completions API.  The Anthropic Messages API (`CreateMessageRequest`) has its own first-class `thinking` field already; the Responses API has `ResponseReasoningParam` already.
- No wrapper/delegation changes required: `Normalizable` is only implemented on `ChatCompletionRequest`, and `chat_template_kwargs` flows verbatim to the gRPC backend.

- 6 new unit tests in `chat.rs`:
  - `reasoning: {enabled: false}` sets `enable_thinking=false`
  - `reasoning: {enabled: true}` sets `enable_thinking=true`
  - `reasoning: {}` (missing `enabled`) is rejected by serde
  - `reasoning: {enabled: true, effort: "low"}` parses effort
  - No `reasoning` field leaves `chat_template_kwargs` untouched
  - Explicit `chat_template_kwargs` wins over `reasoning`
- `cargo test -p openai-protocol --lib` -> 61 passed
- `cargo clippy -p openai-protocol --all-features -- -D warnings` -> clean
- `cargo +nightly fmt --all` -> no diff
- End-to-end against SMG+TRT tp=4, Kimi-K2.5-NVFP4:

    # before fix (field silently dropped by serde)
    reasoning.enabled=false -> 54 completion tokens, reasoning_content populated

    # after fix
    reasoning.enabled=false -> 1 completion token, reasoning_content=''
    no `reasoning` field    -> 48 completion tokens (Kimi default thinking on)
    reasoning.enabled=true  -> 48 completion tokens

## Test Plan

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added reasoning configuration for chat completion requests, allowing users to enable/disable reasoning and optionally configure effort levels.

* **Tests**
  * Added unit tests validating reasoning configuration handling, including configuration precedence and edge cases.

### DIFF
--- a/crates/protocols/src/chat.rs
+++ b/crates/protocols/src/chat.rs
@@ -136,6 +136,35 @@ impl MessageContent {
 }
 
 // ============================================================================
+// Reasoning Configuration (OpenAI-style)
+// ============================================================================
+
+/// Configuration for reasoning models on the Chat Completions API.
+///
+/// OpenAI clients send this as ``{"reasoning": {"enabled": bool}}`` to
+/// explicitly toggle thinking for reasoning-capable models such as
+/// Kimi-K2.5 or Qwen3.  The value is mapped to ``chat_template_kwargs``
+/// (``enable_thinking`` / ``thinking``) by ``ChatCompletionRequest::
+/// normalize()`` so the chat template sees the caller's intent.
+///
+/// Without this struct, serde silently drops the field and reasoning
+/// models think by default -- inflating output sequence length on
+/// traffic that explicitly disabled it.
+#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema)]
+pub struct ReasoningConfig {
+    /// Whether reasoning / thinking should run for this request.
+    pub enabled: bool,
+
+    /// Optional effort hint (``low`` / ``medium`` / ``high``).  Migrated
+    /// to the top-level ``reasoning_effort`` field by ``normalize()``
+    /// (which is the field ``chat_utils.rs`` forwards to the chat
+    /// template).  The top-level ``reasoning_effort`` wins when both
+    /// are set.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub effort: Option<String>,
+}
+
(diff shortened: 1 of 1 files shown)


CatherineSue (commented):
What does OpenAI-style reasoning mean here? This is not an official OpenAI API
https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create

ConnorLi96:
> What does OpenAI-style reasoning mean here? This is not an official OpenAI API

yes, you're right. I should check OpenAI's doc more carefully, will close this PR. ty!

PR #1351, opened 2026-04-23
feat(grpc): add TokenSpeed gRPC client and router wiring (Part 1/3)

yetone:
### Problem

SMG has no first-class TokenSpeed backend. Routing to a TokenSpeed worker requires the gateway to speak TokenSpeed's dedicated gRPC protocol, plumb it through the router's per-backend IR, and pick TokenSpeed up correctly during runtime detection.

### Solution

This PR (Part 1 of 3) adds the Rust side of TokenSpeed support: a self-contained `tokenspeed.grpc.scheduler.TokenSpeedScheduler` proto, a `TokenSpeedSchedulerClient` that produces and consumes native `tokenspeed_proto` types end-to-end, dedicated `ProtoGenerateRequest::TokenSpeed` / `ProtoGenerateStreamChunk::TokenSpeed` / `ProtoGenerateComplete::TokenSpeed` IR arms in the router (same shape as the existing per-backend arms), and the `RuntimeType::TokenSpeed` variant.

PR1 is functionally inert without PR2 — the router can dial a TokenSpeed worker, but the worker process is in PR2's Python servicer.

### 3-PR Stack

- **PR1 (this):** Rust gRPC + protocol — base `main`
- **PR2** (#1464): Python servicer (`grpc_servicer/smg_grpc_servicer/tokenspeed/`) + unit tests — base `feat/grpc-servicer-tokenspeed`
- **PR3** (#1465): CI workflows + e2e tests + install scripts — base `feat/grpc-tokenspeed-servicer`

## Changes

- New `crates/grpc_client/proto/tokenspeed_scheduler.proto` — fully self-contained wire definition (package `tokenspeed.grpc.scheduler`)
- New `crates/grpc_client/src/tokenspeed_scheduler.rs` — `TokenSpeedSchedulerClient` returning native `tokenspeed_proto::*` types
- `crates/protocols/src/worker.rs` — `RuntimeType::TokenSpeed` variant
- `model_gateway/src/routers/grpc/proto_wrapper.rs` — dedicated `TokenSpeed` arms across the IR enums + per-arm accessors
- `model_gateway/src/routers/grpc/client.rs` — dispatch routes the new variant through `TokenSpeedSchedulerClient`
- `model_gateway/src/workflow/steps/local/detect_backend.rs` — health-check ordering extended with tokenspeed
- Stage updates: harmony/regular request building, multimodal, embedding

## Test Plan

- [x] `cargo +nightly fmt --all -- --check` passes
- [x] `cargo clippy -p smg-grpc-client --all-targets --all-features -- -D warnings` passes
- [x] `cargo clippy -p smg --all-targets --all-features -- -D warnings` passes
- [x] `cargo check -p smg --bin smg` succeeds against current `main`

End-to-end coverage runs in PR3 once both the Rust router and the Python servicer (PR2) are in place.

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/crates/grpc_client/build.rs
+++ b/crates/grpc_client/build.rs
@@ -2,6 +2,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Rebuild triggers
     println!("cargo:rerun-if-changed=proto/common.proto");
     println!("cargo:rerun-if-changed=proto/sglang_scheduler.proto");
+    println!("cargo:rerun-if-changed=proto/tokenspeed_scheduler.proto");
     println!("cargo:rerun-if-changed=proto/vllm_engine.proto");
     println!("cargo:rerun-if-changed=proto/trtllm_service.proto");
     println!("cargo:rerun-if-changed=proto/mlx_engine.proto");
@@ -43,6 +44,28 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             ],
             &["proto"],
         )?;
+
+    // Pass 3: compile the TokenSpeed scheduler proto. It imports SGLang
+    // message types (see proto/tokenspeed_scheduler.proto), so we point
+    // tonic at the already-generated ``crate::sglang_scheduler::proto``
+    // module via ``extern_path`` — otherwise tonic would try to generate a
+    // parallel ``sglang.grpc.scheduler`` module under tokenspeed's output
+    // file and produce ``too many leading super keywords`` errors.
+    //
+    // Write into a dedicated sub-directory so this pass's ``extern_path``
+    // on the SGLang package doesn't clobber Pass 2's ``sglang.grpc.scheduler.rs``
+    // — tonic still emits a stub file for every compiled proto even when
+    // the message types inside are extern-pathed out.
+    let ts_out_dir = std::path::PathBuf::from(std::env::var("OUT_DIR")?).join("tokenspeed");
+    std::fs::create_dir_all(&ts_out_dir)?;
+    tonic_prost_build::configure()
+        .build_server(true)
+        .build_client(true)
+        .out_dir(&ts_out_dir)
+        .extern_path(".smg.grpc.common", "crate::common_proto")
+        .extern_path(".sglang.grpc.scheduler", "crate::sglang_scheduler::proto")
+        .protoc_arg("--experimental_allow_proto3_optional")
+        .compile_protos(&["proto/tokenspeed_scheduler.proto"], &["proto"])?;
 
     Ok(())
 }
--- a/crates/grpc_client/build.rs
+++ b/crates/grpc_client/build.rs
@@ -14,14 +14,20 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .protoc_arg("--experimental_allow_proto3_optional")
         .compile_protos(&["proto/common.proto"], &["proto"])?;
 
-    // Pass 2: compile engine protos, referencing common types via extern_path
+    // Pass 2: compile engine protos, referencing common types via extern_path.
+    //
+    // Note: TokenSpeed defines its own message catalog (see
+    // proto/tokenspeed_scheduler.proto) and is fully independent from SGLang —
+    // no shared types, no cross-proto imports. It compiles in the same pass
+    // as the rest because there's nothing special about it anymore.
     tonic_prost_build::configure()
         .build_server(true)
         .build_client(true)
         .extern_path(".smg.grpc.common", "crate::common_proto")
         .type_attribute("GetModelInfoResponse", "#[derive(serde::Serialize)]")
         // vllm + trtllm ServerInfo have only primitive fields.
-        // sglang's contains prost_types::{Struct,Timestamp} so it's handled separately.
+        // sglang's and tokenspeed's contain prost_types::{Struct,Timestamp};
+        // those are handled separately at the wrapper layer.
         .type_attribute(
             "vllm.grpc.engine.GetServerInfoResponse",
             "#[derive(serde::Serialize)]",
@@ -41,31 +47,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 "proto/vllm_engine.proto",
                 "proto/trtllm_service.proto",
                 "proto/mlx_engine.proto",
+                "proto/tokenspeed_scheduler.proto",
             ],
             &["proto"],
         )?;
-
-    // Pass 3: compile the TokenSpeed scheduler proto. It imports SGLang
-    // message types (see proto/tokenspeed_scheduler.proto), so we point
-    // tonic at the already-generated ``crate::sglang_scheduler::proto``
-    // module via ``extern_path`` — otherwise tonic would try to generate a
-    // parallel ``sglang.grpc.scheduler`` module under tokenspeed's output
-    // file and produce ``too many leading super keywords`` errors.
-    //
-    // Write into a dedicated sub-directory so this pass's ``extern_path``
-    // on the SGLang package doesn't clobber Pass 2's ``sglang.grpc.scheduler.rs``
-    // — tonic still emits a stub file for every compiled proto even when
-    // the message types inside are extern-pathed out.
-    let ts_out_dir = std::path::PathBuf::from(std::env::var("OUT_DIR")?).join("tokenspeed");
-    std::fs::create_dir_all(&ts_out_dir)?;
-    tonic_prost_build::configure()
-        .build_server(true)
-        .build_client(true)
-        .out_dir(&ts_out_dir)
-        .extern_path(".smg.grpc.common", "crate::common_proto")
-        .extern_path(".sglang.grpc.scheduler", "crate::sglang_scheduler::proto")
-        .protoc_arg("--experimental_allow_proto3_optional")
-        .compile_protos(&["proto/tokenspeed_scheduler.proto"], &["proto"])?;
 
     Ok(())
 }
--- a/crates/grpc_client/build.rs
+++ b/crates/grpc_client/build.rs
@@ -2,6 +2,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Rebuild triggers
     println!("cargo:rerun-if-changed=proto/common.proto");
     println!("cargo:rerun-if-changed=proto/sglang_scheduler.proto");
+    println!("cargo:rerun-if-changed=proto/tokenspeed_scheduler.proto");
     println!("cargo:rerun-if-changed=proto/vllm_engine.proto");
     println!("cargo:rerun-if-changed=proto/trtllm_service.proto");
     println!("cargo:rerun-if-changed=proto/mlx_engine.proto");
@@ -19,8 +20,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .build_client(true)
         .extern_path(".smg.grpc.common", "crate::common_proto")
         .type_attribute("GetModelInfoResponse", "#[derive(serde::Serialize)]")
-        // vllm + trtllm ServerInfo have only primitive fields.
-        // sglang's contains prost_types::{Struct,Timestamp} so it's handled separately.
+        // Some ServerInfo protos contain prost_types::{Struct, Timestamp};
+        // those are handled separately at the wrapper layer.
         .type_attribute(
             "vllm.grpc.engine.GetServerInfoResponse",
             "#[derive(serde::Serialize)]",
@@ -40,6 +41,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 "proto/vllm_engine.proto",
                 "proto/trtllm_service.proto",
                 "proto/mlx_engine.proto",
+                "proto/tokenspeed_scheduler.proto",
             ],
             &["proto"],
         )?;
--- a/crates/grpc_client/src/lib.rs
+++ b/crates/grpc_client/src/lib.rs
@@ -20,8 +20,11 @@ use std::sync::Arc;
 pub use mlx_engine::{proto as mlx_proto, MlxEngineClient};
 pub use sglang_scheduler::{proto as sglang_proto, SglangSchedulerClient};
 use tonic::metadata::MetadataMap;
-// TokenSpeed reuses ``sglang_proto`` for message types today; the module
-// below owns the ``tokenspeed.grpc.scheduler`` service stub and its client.
+// TokenSpeed has a fully independent wire definition (see
+// ``proto/tokenspeed_scheduler.proto``) — distinct service, distinct
+// messages with intentionally trimmed field sets aimed at top-tier LLM
+// workloads. The client wraps that wire and translates to/from SGLang-shaped
+// types at the boundary so the router's dispatch enums don't proliferate.
 pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
 pub use trtllm_service::{proto as trtllm_proto, TrtllmServiceClient};
 pub use vllm_engine::{proto as vllm_proto, VllmEngineClient};
--- a/crates/grpc_client/src/lib.rs
+++ b/crates/grpc_client/src/lib.rs
@@ -10,6 +10,7 @@ pub mod common_proto {
 pub mod mlx_engine;
 pub mod sglang_scheduler;
 pub mod tokenizer_bundle;
+pub mod tokenspeed_scheduler;
 pub mod trtllm_service;
 pub mod vllm_engine;
 
@@ -19,6 +20,9 @@ use std::sync::Arc;
 pub use mlx_engine::{proto as mlx_proto, MlxEngineClient};
 pub use sglang_scheduler::{proto as sglang_proto, SglangSchedulerClient};
 use tonic::metadata::MetadataMap;
+// TokenSpeed reuses ``sglang_proto`` for message types today; the module
+// below owns the ``tokenspeed.grpc.scheduler`` service stub and its client.
+pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
 pub use trtllm_service::{proto as trtllm_proto, TrtllmServiceClient};
 pub use vllm_engine::{proto as vllm_proto, VllmEngineClient};
 
--- a/crates/grpc_client/src/lib.rs
+++ b/crates/grpc_client/src/lib.rs
@@ -25,9 +25,9 @@ pub use sglang_scheduler::{proto as sglang_proto, SglangSchedulerClient};
 // workloads. The client wraps that wire and translates to/from SGLang-shaped
 // types at the boundary so the router's dispatch enums don't proliferate.
 pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
+use tonic::metadata::MetadataMap;
 pub use trtllm_service::{proto as trtllm_proto, TrtllmServiceClient};
 pub use vllm_engine::{proto as vllm_proto, VllmEngineClient};
-use tonic::metadata::MetadataMap;
 
 /// Shared `get_tokenizer()` implementation for all engine clients.
 ///
--- a/crates/grpc_client/src/lib.rs
+++ b/crates/grpc_client/src/lib.rs
@@ -19,7 +19,6 @@ use std::sync::Arc;
 
 pub use mlx_engine::{proto as mlx_proto, MlxEngineClient};
 pub use sglang_scheduler::{proto as sglang_proto, SglangSchedulerClient};
-use tonic::metadata::MetadataMap;
 // TokenSpeed has a fully independent wire definition (see
 // ``proto/tokenspeed_scheduler.proto``) — distinct service, distinct
 // messages with intentionally trimmed field sets aimed at top-tier LLM
@@ -28,6 +27,7 @@ use tonic::metadata::MetadataMap;
 pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
 pub use trtllm_service::{proto as trtllm_proto, TrtllmServiceClient};
 pub use vllm_engine::{proto as vllm_proto, VllmEngineClient};
+use tonic::metadata::MetadataMap;
 
 /// Shared `get_tokenizer()` implementation for all engine clients.
 ///
--- a/crates/grpc_client/src/lib.rs
+++ b/crates/grpc_client/src/lib.rs
@@ -1,7 +1,8 @@
-//! gRPC clients for SGLang, vLLM, TensorRT-LLM, and MLX backends
+//! gRPC clients for SGLang, vLLM, TensorRT-LLM, MLX, and TokenSpeed backends.
 //!
 //! This crate provides gRPC client implementations for communicating with
-//! SGLang scheduler, vLLM engine, TensorRT-LLM engine, and MLX engine backends.
+//! the SGLang scheduler, vLLM engine, TensorRT-LLM engine, MLX engine, and
+//! TokenSpeed scheduler backends.
 
 pub mod common_proto {
     #![allow(clippy::all, clippy::absolute_paths, unused_qualifications)]
@@ -12,6 +13,7 @@ pub mod channel;
 pub mod mlx_engine;
 pub mod sglang_scheduler;
 pub mod tokenizer_bundle;
+pub mod tokenspeed_scheduler;
 pub mod trtllm_service;
 pub mod vllm_engine;
 
@@ -22,6 +24,7 @@ pub use abort_on_drop::{AbortOnDropClient, AbortOnDropStream};
 pub use channel::{connect_channel, normalize_grpc_endpoint};
 pub use mlx_engine::{proto as mlx_proto, MlxEngineClient};
 pub use sglang_scheduler::{proto as sglang_proto, SglangSchedulerClient};
+pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
 use tonic::metadata::MetadataMap;
 pub use trtllm_service::{proto as trtllm_proto, TrtllmServiceClient};
 pub use vllm_engine::{proto as vllm_proto, VllmEngineClient};
--- a/crates/grpc_client/src/tokenspeed_scheduler.rs
+++ b/crates/grpc_client/src/tokenspeed_scheduler.rs
@@ -33,7 +33,7 @@ use openai_protocol::{
     chat::ChatCompletionRequest, completion::CompletionRequest, generate::GenerateRequest,
     messages::CreateMessageRequest, responses::ResponsesRequest,
 };
-use tonic::{transport::Channel, Streaming, Request};
+use tonic::{transport::Channel, Request, Streaming};
 use tracing::{debug, warn};
 
 use crate::{
--- a/crates/grpc_client/src/tokenspeed_scheduler.rs
+++ b/crates/grpc_client/src/tokenspeed_scheduler.rs
@@ -1,54 +1,126 @@
 //! gRPC client for the TokenSpeed scheduler service.
 //!
-//! Wire-level identity (package ``tokenspeed.grpc.scheduler``, service
-//! ``TokenSpeedScheduler``) is distinct from SGLang — see
-//! ``proto/tokenspeed_scheduler.proto``. Message types are currently
-//! imported from the SGLang proto (the two backends happen to share their
-//! tokenized-request and sampling-param shapes today), so this wrapper
-//! reuses ``crate::sglang_scheduler::proto`` for all request / response
-//! types and only implements the thin layer that dispatches to the
-//! ``TokenSpeedScheduler`` service. When the two protocols diverge, the
-//! shared message types should be split into this module (or a
-//! ``tokenspeed_proto`` submodule) and the call sites updated.
+//! TokenSpeed has a fully independent wire definition (see
+//! ``proto/tokenspeed_scheduler.proto``) — distinct package
+//! (``tokenspeed.grpc.scheduler``), distinct service, distinct messages with
+//! intentionally trimmed field sets aimed at the top-tier LLM workloads
+//! (Kimi K2, MiniMax M2, Qwen 3, gpt-oss, DeepSeek V4). Anything SGLang has
+//! that doesn't apply here (PD-disaggregated serving, multimodal inputs,
+//! LoRA hot-swap, hidden-state forwarding, embeddings, classifier outputs,
+//! tokenizer streaming, KV-event subscription) is simply not on TokenSpeed's
+//! wire.
 //!
-//! Also reuses ``SglangSchedulerClient``'s request-builder helpers
-//! (``build_generate_request_from_chat``, ``build_grpc_sampling_params_*``,
-//! etc.) — those are pure conversions from openai-protocol types to the
-//! shared proto types, so duplicating them would be busy-work.
+//! Internally this client still leverages SGLang's
+//! ``build_grpc_sampling_params_from_*`` helpers because the source-of-truth
+//! is an OpenAI request and most fields map identically. We translate from
+//! the SGLang-shaped ``GenerateRequest`` into a TokenSpeed-shaped one at the
+//! wire boundary, and translate the streamed response back so the router's
+//! ``ProtoGenerateStreamChunk`` / ``ProtoGenerateComplete`` accessors can
+//! operate on a familiar shape. When TokenSpeed needs a field SGLang lacks,
+//! add it to the proto and extend the translator — not the router.
 
-use std::{sync::Arc, time::Duration};
+use std::{
+    pin::Pin,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    task::{Context, Poll},
+    time::Duration,
+};
 
 use openai_protocol::{
     chat::ChatCompletionRequest, completion::CompletionRequest, generate::GenerateRequest,
     messages::CreateMessageRequest, responses::ResponsesRequest,
 };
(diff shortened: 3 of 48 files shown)


yetone:
Review addressed — pushed `7aca1716` (force-push).

### Fixed (from CodeRabbit review)

- 🔴 **`model_specs.py` duplicate key** (line 55-59 + line 80-86). I added a second `Qwen/Qwen3-30B-A3B` entry instead of merging; Python kept the later (thinking/reasoning) one, silently dropping `function_calling` + `tool_choice` features. **This was also the `python-lint` / `pre-commit` CI failure** (ruff F601). Merged into the existing entry with the union of features.
- 🟠 **P1: Abort sweep for n>1 child rids**. Previously `Abort` only hit the original `request_id`; for `n>1` requests `Generate` minted `{rid}-n{i}` children that kept consuming GPU work on client-driven abort. `Abort` now sweeps the original and every `{rid}-n*` child.
- **P1: Cancel path also sweeps n>1 children**. Same class of bug in the `except asyncio.CancelledError` branch in `Generate` — fixed alongside a new `test_cancel_aborts_all_n_children` unit test.
- **🟡 `Embed` missing `CancelledError` / `finally` cleanup**. Now mirrors `Generate`'s cleanup pattern so a client disconnect during an embed call doesn't leak the rid in `rid_to_state`.
- **Stale docstrings** (`constants.py`, `worker.py`, `test_function_calling.py` class-level comment/docstring, "Llama-3.2-1B is flaky" comment).
- **Unused `SimpleNamespace` import** in `test_tokenspeed_health_servicer.py`.
- **Ruff `UP038` issues** (PEP 604 isinstance union style) in `servicer.py`.
- **Ruff format** applied to all touched files.

### Fixed (from `e2e-1gpu-chat` CI failures)

- **Model swap → OOM on 80 GB H100 CI**: reviewer correctly predicted `Qwen/Qwen3-30B-A3B` (~66 GB bf16 weights alone) wouldn't fit. Switched `TestOpenAIServerFunctionCalling` to `Qwen/Qwen3-4B` (dense, still `Qwen3ForCausalLM` — TokenSpeed-supported). This also unbreaks the `sglang` / `vllm` / `trtllm` variants of the same test class that the 30B swap had inadvertently broken.
- Kept `Qwen/Qwen3-30B-A3B` available in `model_specs.py` (for the thinking/reasoning test classes that already used it) — the duplicate was the only bug.

### Added — TokenSpeed in GitHub CI

Missing from the first push (my bad). Wired up end-to-end:

- `scripts/ci_install_tokenspeed.sh` — clones TokenSpeed, builds kernel (CUDA) + scheduler (C++/CMake/nanobind) + Python runtime, caches the built kernel wheel at `/tmp/tokenspeed-wheel-cache/` to skip the ~30-min CUDA compile on subsequent runs. Mirrors the install-from-source + wheel-cache pattern already used for `ci_install_trtllm.sh`.
- `.github/actions/setup-tokenspeed/action.yml` — composite action wiring the install script alongside the venv setup.
- `.github/workflows/e2e-gpu-job.yml` — new `Setup TokenSpeed backend` step gated on `inputs.engine == 'tokenspeed'`, plus a docstring update.
- `.github/workflows/pr-test-rust.yml`:
  - `e2e-1gpu-chat` matrix: added `tokenspeed` (60-min timeout, matches TRT-LLM's pattern for build-from-source engines).
  - `detect-changes` paths-filter: added `scripts/ci_install_tokenspeed.sh` to `common` and `chat-completions` filters so edits to the install script re-trigger the test.

### Servicer fallback dropped

Servicer / scheduler_launcher no longer dual-mode import. Target: only the post-Phase-D.3 `AsyncLLM` (standalone class, not the legacy `TokenizerManager` wrapper). This matches the e2e evidence from the prior asciinema (https://asciinema.org/a/3pl8PKYo92DnSA68) where `AsyncLLM.__bases__ = (EngineClient, TokenizerCommunicatorMixin)`.

### What I did NOT change (out of scope for this PR)

- `TestToolChoiceLlama` still uses `meta-llama/Llama-3.2-1B-Instruct` + `llama` parser — baseline Llama regression coverage preserved for sglang/vllm/trtllm. This class is NOT marked to run on `tokenspeed` because TokenSpeed's registry doesn't include `LlamaForCausalLM`; flip that marker back in a follow-up once TokenSpeed registers it.
- smg gateway's `tool_choice="required"` / `tool_choice="specific"` constraint translation — vLLM fails these same tests identically on the same model, so the failure is upstream of the engine and belongs in a separate smg-gateway PR.

yetone:
## Dug into the 4 remaining `tool_choice="auto"` failures — root-caused + fixed

### Reproduction

Full-stack e2e on a single H100: tokenspeed gRPC worker + smg gateway + openai client against `meta-llama/Llama-3.2-1B-Instruct`. Setup matches CI (`ci_install_tokenspeed.sh` + `TestOpenAIServerFunctionCalling`).

### Root cause

Instrumented the servicer to log what flows out:

```
[SRV-AFTER] completion=22 raw=27
first10=[128009, 128006, 78191, 128007, 271, 5018, 609, 794, 330, 723]
last5  =[65, 794, 330, 20, 32075]
+ trailing 128008 (<|eom_id|>)
```

TokenSpeed's AsyncLLM prepends the Llama-3 assistant-role header tokens — `[<|eot_id|>, <|start_header_id|>, "assistant", <|end_header_id|>, "\n\n"]` — to every `output_ids`, **plus** includes the matched stop token. When the smg gateway detokenizes with `skip_special_tokens=True`, the 128xxx control tokens get stripped, but `"assistant"` (78191) and `"\n\n"` (271) are **regular word tokens** so they survive. The decoded text arriving at the `llama` parser looks like:

```
assistant\n\n{"name": "add", "parameters": {"a": "3", "b": "5"}}
```

The `llama` parser's `parse_complete` does `text.trim_start().starts_with('{')` which returns **false** because of the `assistant\n\n` prefix, falls through to "no JSON structure found", and emits an empty tool-call list.

Why this didn't hit Qwen3-30B: the `qwen` parser uses a regex to extract `<tool_call>...</tool_call>` tags and is tolerant of both leading prefix and trailing special tokens.

Why this didn't hit sglang/vllm/trtllm on the same model: their respective gRPC servicers/backends emit `token_ids` containing only the model's generated content — no header echo, no stop token included. SGLang's on-wire shape is the contract the gateway was built for; this PR's tokenspeed servicer was forwarding `output_ids` verbatim, inheriting TokenSpeed-AsyncLLM-internal details.

### Fix

Added `_generated_output_ids(output, reason_dict)` to the tokenspeed servicer. It:

1. Slices `raw[-meta.completion_tokens:]` to drop the header prefix.
2. If the finish reason is `"stop"` and the trailing id matches `reason.matched`, drops that too.

`matched_token_id` still rides in the separate proto field. `meta_info.completion_tokens` passes through unchanged (SGLang's convention — it reports generation accounting, not wire-payload length).

Updated unit tests:
- `test_non_streaming_emits_complete` now asserts the stripped wire shape (`[10, 11]` with `matched_token_id=12`).
- New `test_strips_chat_template_prefix` — reproduces the real 27-token payload shape and asserts the servicer forwards only the 21 content tokens.

### Result

`TestOpenAIServerFunctionCalling` against `unsloth/Llama-3.2-1B-Instruct` + tokenspeed (ungated mirror of the official repo; identical weights; local test environment had no HF gated-repo token):

| Run | passed | failed | skipped |
|---|---:|---:|---:|
| **Before fix** | 6 | 10 | 2 |
| **After fix**  | 10 | **6** | 2 |

The 6 remaining failures are the pre-existing smg-gateway `tool_choice=required/specific` constraint-translation bug (same failure set on vLLM for the same model — unrelated to this PR). See the earlier reference runs in the PR description.

Commit: 77188437

yetone:
## Correction and dig deeper on the 6 tool_choice=required/specific failures

@yetone thank you for catching this — my earlier comment was wrong. You were right: vLLM passes all of `TestOpenAIServerFunctionCalling` on Llama-3.2-1B-Instruct in CI. My "vLLM fails the same way" claim was stale, remembered from a Qwen3-30B run. Sorry for the noise.

### What I found actually going on

Dug back into the 6 remaining failures (all are `tool_choice=required` / `tool_choice={specific fn}` × n variants × {openai,smg} clients). Two issues chained:

**(1) TokenSpeed defaults `grammar_backend=None`.**

The smg gateway translates `tool_choice=required/specific` into a `sampling_params.json_schema` constraint. TokenSpeed with no grammar backend **silently drops the constraint** — no warning, no error, just ignored. vLLM/SGLang/TRT-LLM all have their grammar backends enabled by default (TRT-LLM via `guided_decoding_backend: xgrammar` in the config file, which `_build_trtllm_cmd` sets). Fixed in `b6904f12` by adding `--grammar-backend xgrammar` to the tokenspeed worker launch.

**(2) xgrammar in tokenspeed isn't enforcing the smg gateway's schema.**

Even with the grammar backend on, the 3 tests still fail. I captured what the gateway actually sends as the constraint for a 2-tool `tool_choice=required` request:

```json
{
  "type": "array",
  "minItems": 1,
  "items": {
    "type": "object",
    "anyOf": [
      {"properties": {"name": {"enum": ["sub"]},      "parameters": {...}}, "required": ["name","parameters"]},
      {"properties": {"name": {"enum": ["get_weather"]}, "parameters": {...}}, "required": ["name","parameters"]}
    ]
  }
}
```

That schema is well-formed and unambiguous: force an **array** of `{name, parameters}` objects where `name` is one of the declared tool names. `parse_json_schema_response` in the Rust gateway then pops the array and wraps each item as a `ToolCall`.

What the model (Llama-3.2-1B through tokenspeed) actually emits, verified via live server log capture with my servicer-side `[SRV-CONSTRAINT]` / `content` probes:

```json
{"function": "get_weather", "parameters": {"properties": {"city": "Paris"}}}
```

- not an array (schema requires `type: array`)
- `"function"` key instead of `"name"`
- `"properties"` from the schema leaking in as a payload key — the word "properties" isn't in the prompt or tool definitions anywhere; it's a JSON Schema syntax keyword

That only happens if xgrammar compiled the `anyOf`+`enum` schema to a degraded/no-op grammar, and the model is free-running and cargo-culting schema-shaped tokens. vLLM/SGLang's grammar stacks handle this correctly, which is why those engines pass.

### Scope call

This second issue is tokenspeed-side (xgrammar's schema handling, not the gRPC servicer). Not something I can fix inside `grpc_servicer/` without either:

1. Rewriting the smg gateway's schema generator to emit tokenspeed-xgrammar-compatible shapes (bypassing `anyOf` with `enum`), **or**
2. Fixing xgrammar's handling upstream.

Both are out of scope for this PR — the PR's job is "make TokenSpeed routable through the SGLang gRPC path." With the two fixes I've landed (output-id normalization in `77188437`, grammar-backend in `b6904f12`), we get:

- 10 passed / 6 failed / 2 skipped on Llama-3.2-1B-Instruct (was 6 passed / 10 failed before those two commits)
- All 6 remaining failures are the same `tool_choice=required/specific` × n × client bucket
- Non-tool-choice-constrained tests pass

I'll open a tokenspeed issue to track (2) and link it here once the repo is in a state to accept it. If you'd rather not ship this PR until those 3 tests pass as well, I can also mark them with `pytest.mark.xfail(engine=="tokenspeed", strict=False)` so the test suite is green — let me know which you prefer.

yetone:
Filed lightseekorg/tokenspeed#361 with the grammar-pipeline writeup — it documents the architectural gap (`Req`-based scaffolding is dead code in the AsyncLLM path) and outlines the hook points the fix will need. The 3 xfailed tests in this PR reference that issue, so they'll self-unmute once the upstream wiring lands. No change here in the meantime.

yetone:
## Added: TokenSpeed as a distinct `RuntimeType` (commit `b37935c9`)

You were right — labeling TokenSpeed workers as `RuntimeType::Sglang` at the router would make them indistinguishable from real SGLang workers in metrics/traces/registry. Keeping the wire reuse (same proto, same client) but making identity independent at the router boundary:

| Change | Why |
|---|---|
| `RuntimeType::TokenSpeed` variant (`crates/protocols/src/worker.rs`) | The canonical enum already drives Display/FromStr used by metrics & worker registry. |
| `GrpcClient::connect`: `"sglang" \| "tokenspeed" => Self::Sglang(...)` (`client.rs`) | Wire format is identical, so we share the same client — just a different label. |
| Post-handshake `GetServerInfo` promotion in `detect_backend.rs` | Real SGLang schedulers never stamp `__ts_backend_runtime__` in `server_args`; presence of that key flips the detected runtime from `"sglang"` → `"tokenspeed"`. Failures fall back to `"sglang"` so a slow/flaky worker still works, just labeled less precisely. |
| `request_execution.rs` PD match adds `RuntimeType::TokenSpeed` to the "not supported" arm | TokenSpeed 0.1 doesn't ship PD disagg; matches how trtllm/mlx are handled. |
| `LOCAL_OWNED_BY` whitelist in `classify.rs` adds `"tokenspeed"` | Defensive — covers raw-HTTP TokenSpeed workers that don't go through the gRPC servicer. |

Python side stamps `BACKEND_RUNTIME_MARKER_KEY = "__ts_backend_runtime__"` into `GetServerInfo.server_args` via `servicer.py`. Constants on both sides are commented to stay in lockstep. New unit test `test_stamps_runtime_marker_in_server_args` fails loudly if the marker is ever dropped.

Verified `cargo check -p smg --bin smg` compiles clean locally after the changes.

key4ng:
## Cancellation behavior — verification

Verified that cancelling an in-flight `Generate` (Ctrl-C / channel close / explicit `call.cancel()`) cleanly stops scheduler-side work and the engine recovers for new requests. Live tests against `Qwen/Qwen3-4B-Instruct-2507` on a single H100, plus a code audit of tokenspeed's scheduler internals.

### TL;DR

| Scenario | Frontend (`active_requests`) | Scheduler-side (`num_running_reqs`, `num_waiting_reqs`) | Recovery |
|---|---|---|---|
| Streaming, n=1 | 1 → 0 | 1 → 0 | ✅ |
| Non-streaming, n=1 | 1 → 0 | 1 → 0 | ✅ |
| Streaming, n=3 (parallel sampling) | 3 → 0 | 3 → 0 | ✅ |
| Channel close (TCP drop) | 1 → 0 | 1 → 0 | ✅ |
| Cancel during prefill (16K-token prompt) | 1 → 0 | 1 → 0 | ✅ |

`active_requests` polled via `GetServerInfo`, `num_running_reqs`/`num_waiting_reqs` polled via `GetLoads`. **Both** drain to 0 immediately on cancel, including for `n>1` where the servicer expands the rid into `[{rid}-n{i}]` per-choice ids.

### Code paths

The cancellation chain end-to-end:

1. gRPC client `call.cancel()` → server-side `asyncio.CancelledError` raised inside `Generate`
2. Servicer's `except asyncio.CancelledError` block (`grpc_servicer/smg_grpc_servicer/tokenspeed/servicer.py:206-216`) calls `self.async_llm.abort_request(rid)` for each rid (single rid OR the expanded `{rid}-n{i}` list when `n>1`)
3. `AsyncLLM.abort_request` (`tokenspeed/runtime/engine/async_llm.py:536-541`) does `del rid_to_state[rid]` and sends `AbortReq` over zmq to the scheduler subprocess
4. Scheduler's `mark_abort` (`generation_output_processor.py:308`) flips `state.to_abort`; `check_finished` emits `FINISH_ABORT` at the next iteration boundary
5. `cache_finished_req` (`prefix_cache.py:326-386`) inserts the request's prefilled KV into the radix tree and decrements `lock_ref` (KV pages become evictable)

### Evidence — full results

**Streaming + non-streaming + n>1:**
```
=== STREAMING — cancel after 5 chunks ===
received 5 chunks, active_requests=1; calling call.cancel()
call ended with: StatusCode.CANCELLED
after cancel: active_requests=0

=== NON-STREAMING — cancel mid-generation ===
~1s in, active_requests=1; calling call.cancel()
call ended with: StatusCode.CANCELLED
after cancel: active_requests=0

=== n>1 (parallel sampling) cancel ===
got 6 chunks across choices [0, 1, 2], mid-active=3
call ended: StatusCode.CANCELLED
after cancel (n=3): active=0

=== Channel close mid-stream ===
3 chunks in: active=1, scheduler_running=0; closing channel ch2
drained at iter 0: active=0, scheduler_running=0, scheduler_waiting=0
```

**Cancel during prefill (3 timing windows on a 16,501-token prompt):**
```
Run A: cancel at +6ms  → 0 chunks, drained at iter 0: (0, 0, 0)
Run B: cancel at +50ms → 0 chunks, drained at iter 0: (0, 0, 0)
Run C: cancel at +200ms → 0 chunks, drained at iter 0: (0, 0, 0)
Recovery: 8 tokens, post-cancel metrics (0, 0, 0)
```

### A note on KV-page accounting

A leak detector test (42 cancels with **unique** prompts each, watching `num_used_tokens` from the real `GetLoads`) showed:
- `running` and `waiting` drain to 0 on every cancel ✅
- `used_tokens` grows by exactly +4096 per cancel (the prompt's prefill page count), reaching 38.8% of pool after 42 cancels with no eviction observed in the test window

This is **prefix-cache retention by design**, not a leak. `_get_load`'s `num_used_pages = total_pages - available_kv_pages`, where `available_kv_pages` reports `kv_allocator.available_size()` (free list only). Pages held by the radix cache for prefix reuse sit in `evictable_size_`/`protected_size_` (`prefix_cache.py:515-520`) — outside the free list but **evictable on demand**: `schedule_batch.py:227-243` calls `tree_cache.evict()` (`prefix_cache.py:456-482`) when `kv_allocator.alloc()` returns `None`, heap-popping `lock_ref==0` LRU leaves and feeding their pages back through `kv_allocator.free()`.

So the linear growth is the request's prefilled KV being inserted into the radix tree by `cache_finished_req` on the abort-driven Finish event — not retained slot state, not orphaned active work.

### Independent code-review audit

The above interpretation was confirmed by an independent code audit that traced:
- `event_loop.py:702-717` — `_get_load` reads scheduler-subprocess state, not frontend
- `data_parallel_controller.py:295` — confirms `EventLoop` runs in a separate process
- `prefix_cache.py:326-386` — `cache_finished_req` inserts aborted request's KV into radix tree
- `prefix_cache.py:456-482` — eviction is wired and on-demand under allocator pressure

Auditor verdict: *"Linear +4096/cancel is the request's prefill KV being inserted into the radix tree by `cache_finished_req` on the abort-driven Finish event, not retained slot state. Eviction is wired and on-demand. The `running=0`/`waiting=0` signal originates inside the scheduler subprocess, so it is the definitive cancel-correctness witness."*

### Side effect: real `GetLoads` (was a stub)

This audit surfaced that the previous `GetLoads` was a hardcoded stub returning zeros, which made router-side load balancing useless and any cancel-leak diagnostics tautological. Fixed in `4a39090f` — now bridges to `AsyncLLM.get_load()` and returns real scheduler-side counts, with timeouts and proper proto mapping.

key4ng:
## Cancellation behavior — re-verification

Re-ran the same matrix from [the previous comment](https://github.com/lightseekorg/smg/pull/1351#issuecomment-4354541873) against the live `smg_grpc_servicer.tokenspeed` worker on `Qwen/Qwen3-4B-Instruct-2507`, single H100. All scenarios still drain cleanly.

| Scenario | Frontend (`active_requests`) | Scheduler (`running`, `waiting`) | Recovery |
|---|---|---|---|
| Streaming, n=1 | 1 → 0 | 1 → 0 | ✅ |
| Non-streaming, n=1 | 1 → 0 | 1 → 0 | ✅ |
| Streaming, n=3 (parallel sampling) | 3 → 0 | 3 → 0 | ✅ |
| Channel close (TCP drop) | 1 → 0 | 1 → 0 | ✅ |
| Cancel during prefill, +6ms | 1 → 0 | 0 → 0 | ✅ |
| Cancel during prefill, +50ms | 1 → 0 | 1 → 0 | ✅ |
| Cancel during prefill, +200ms | 1 → 0 | 1 → 0 | ✅ |

`active_requests` from `GetServerInfo`; `running`/`waiting` from `GetLoads` (which now talks to the scheduler subprocess directly, post-`4a39090f`). After every cancel a tiny `Generate(max_new=8)` round-trips successfully, confirming the engine recovers.

### Raw output

```
=== S1: streaming, n=1 — cancel after 5 chunks ===
  mid: chunks=5 active=1 running=1 waiting=0 used=178304
  call ended: CANCELLED
  [S1] drained: active=0 running=0 waiting=0 (used_tokens=178240)
  [S1] recovery: complete=True

=== S2: non-streaming, n=1 — cancel mid-generation ===
  mid: active=1 running=1 waiting=0; cancelling
  call ended: CANCELLED
  [S2] drained: active=0 running=0 waiting=0 (used_tokens=178432)

=== S3: streaming, n=3 (parallel sampling) ===
  mid: chunks=6 choices_seen=[0, 1, 2] active=3 running=3 waiting=0
  call ended: CANCELLED
  [S3] drained: active=0 running=0 waiting=0 (used_tokens=178432)

=== S4: channel close (TCP drop) mid-stream ===
  mid: chunks=3 active=1 running=1 waiting=0; closing channel
  [S4] drained: active=0 running=0 waiting=0 (used_tokens=178432)

=== S5: cancel during prefill (16.5K tokens) ===
  +6ms:   call ended CANCELLED, drained at iter 0: (0, 0, 194880)
  +50ms:  call ended CANCELLED, drained at iter 0: (0, 0, 194880)
  +200ms: call ended CANCELLED, drained at iter 0: (0, 0, 194944)
```

### What changed since the last verification

- `4a39090f` made `GetLoads` real (was a stub) — confirmed scheduler-side `running`/`waiting` counters drain on every cancel.
- `64531a90` tightened the `Abort` child-rid match to `^{rid}-n\d+$` — n>1 still works (S3 above expanded to choices `[0,1,2]` and all three drained).
- `43075075` made warmup-driven `set_serving()` only flip on a real Complete frame; cancel path is unchanged.
- `6fd6316b` returned `Status::unimplemented` for TokenSpeed `get_tokenizer`; unrelated to cancel.

The same `cache_finished_req` → radix-tree retention behaviour holds for the 16K-prompt scenario (used_tokens grows by the prefill page count then plateaus on repeat-prompt cache hits), as analysed in the prior comment.

crates/protocols/src/worker.rs
```diff
@@ -197,6 +197,14 @@ pub enum RuntimeType {
     Trtllm,
     /// MLX runtime (Apple Silicon).
     Mlx,
+    /// TokenSpeed runtime. Speaks the dedicated
```
CatherineSue: Can we shorten the comments here?

crates/grpc_client/build.rs
```diff
@@ -13,14 +14,20 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .protoc_arg("--experimental_allow_proto3_optional")
         .compile_protos(&["proto/common.proto"], &["proto"])?;
 
-    // Pass 2: compile engine protos, referencing common types via extern_path
+    // Pass 2: compile engine protos, referencing common types via extern_path.
+    //
+    // Note: TokenSpeed defines its own message catalog (see
```
CatherineSue: Can we remove the comments here? I don't see its necessarity.

grpc_servicer/tests/conftest.py
```diff
@@ -0,0 +1,22 @@
+"""Pytest configuration for smg-grpc-servicer unit tests.
```
CatherineSue: Let's add tests in a follow up PR

model_gateway/src/workflow/steps/local/detect_backend.rs
```diff
@@ -1,8 +1,15 @@
 //! Backend runtime detection step.
 //!
-//! Detects the runtime type (sglang, vllm, trtllm, mlx) for both HTTP and gRPC workers.
-//! - HTTP: probes `/v1/models` (owned_by field), falls back to unique endpoints.
-//! - gRPC: tries sglang → vllm → trtllm → mlx health checks sequentially.
+//! Detects the runtime type for local workers.
```
CatherineSue: Same in this file, i think the previous comments is pretty concise and straightforward. We can simply just add `gRPC: tries ... -> tokenspeed`.

CatherineSue (commented):
Can we split the CI and the actual gRPC servicer into separate PRs?

e2e_test/chat_completions/test_function_calling.py
```diff
@@ -1510,14 +1509,20 @@ class TestToolChoiceLlama(_TestToolChoiceBase):
...
 @pytest.mark.gateway(extra_args=["--tool-call-parser", "qwen", "--history-backend", "memory"])
 @pytest.mark.parametrize("setup_backend", ["grpc"], indirect=True)
 @pytest.mark.parametrize("api_client", ["openai", "smg"], indirect=True)
 class TestToolChoiceQwen(_TestToolChoiceBase):
-    """Tests for tool_choice functionality with Qwen model."""
+    """Tests for tool_choice functionality with Qwen model.
+
+    Uses Qwen3-4B-Instruct (non-thinking variant) — same
+    `<tool_call>\\n{...}\\n</tool_call>` format as Qwen 2.5, so the gateway's
+    ``qwen`` parser applies, while the Qwen3 family is in TokenSpeed's model
```
CatherineSue: Let's remove the comment here. Comments like this should exist in PR descriptions, not in the docs I think.

crates/protocols/src/worker.rs
key4ng: Done — shortened to a single `/// TokenSpeed runtime.` line in 6b696e53 (now PR1 head).

crates/grpc_client/build.rs
key4ng: Done — dropped the verbose Note block in 6b696e53. Kept just the one-line `// Pass 2: compile engine protos, referencing common types via extern_path` summary.

grpc_servicer/tests/conftest.py
key4ng: Split this PR into a 3-PR stack so the servicer + tests are reviewable independently of the Rust router and the CI changes:

- PR1 (this, #1351): Rust gRPC + protocol only
- PR2 (#1464): Python servicer + unit tests
- PR3 (#1465): CI + e2e

I kept the unit tests with the servicer in PR2 (rather than as a separate follow-up) because they're tightly coupled to the code under test. If you'd still prefer them peeled into a 4th PR, happy to do that — let me know.

model_gateway/src/workflow/steps/local/detect_backend.rs
key4ng: Done — restored the original concise module doc, just adding `tokenspeed` to the existing list. Now reads:

```rust
//! Backend runtime detection step.
//!
//! Detects the runtime type (sglang, vllm, trtllm, tokenspeed, mlx) for both HTTP and gRPC workers.
//! - HTTP: probes `/v1/models` (owned_by field), falls back to unique endpoints.
//! - gRPC: tries sglang → vllm → trtllm → tokenspeed → mlx health checks sequentially.
```

In 6b696e53.

e2e_test/chat_completions/test_function_calling.py
key4ng: Done — moved the rationale to PR3 (#1465)'s description and dropped the docstring on `TestToolChoiceQwen`. Class doc is now just `Tests for tool_choice functionality with Qwen model.` (in 001cf780, on the PR3 branch).

key4ng:
## Reorganized into a 3-PR stack

@CatherineSue per your review, split this into three reviewable PRs:

| PR | Scope | Base |
|---|---|---|
| **#1351 (this)** | Rust gRPC + protocol — proto, generated stubs, `TokenSpeedSchedulerClient`, router wiring, `RuntimeType::TokenSpeed` | `main` |
| **#1464** | Python servicer + unit tests | `feat/grpc-servicer-tokenspeed` |
| **#1465** | CI workflows + e2e tests + install scripts | `feat/grpc-tokenspeed-servicer` |

PR #1464 and #1465 are open as **draft** until this PR is in shape to merge.

### What changed in this PR vs. the previous tip

- All servicer / CI / e2e content moved out — see #1464 and #1465
- All 5 of your inline asks addressed (replied inline with the new commit)
- Dropped the verbose comments per your feedback (worker.rs, build.rs, detect_backend.rs)
- Rebased onto current `main`
- Force-pushed (single commit `6b696e53`); old commit history is preserved in `feat/grpc-servicer-tokenspeed-v2` if needed

### Verification (this PR)

- `cargo +nightly fmt --all -- --check` ✅
- `cargo clippy -p smg-grpc-client --all-targets --all-features -- -D warnings` ✅
- `cargo clippy -p smg --all-targets --all-features -- -D warnings` ✅

### Other reviewer notes

- 🔴 critical bot finding (FakeAsyncLLM crashing on n>1 list rid) is fixed in PR #1464.
- The duplicate `Qwen/Qwen3-30B-A3B` dict-key flagged earlier is not present on the current branch (resolved in an earlier yetone commit; verified `grep -n` shows only one entry).
- Lower-priority bot findings (P1/P2 on warmup-then-SERVING ordering, `temperature=0` preservation, bootstrap_port=0 ambiguity, etc.) are deliberately out of scope for this restructuring; happy to address them in follow-up PRs once this stack lands.

cc @yetone @slin1237 @XinyueZhang369 @gongwei-130

model_gateway/src/routers/grpc/common/stages/helpers.rs
```diff
@@ -119,7 +119,10 @@ pub(crate) fn apply_sampling_defaults_to_generate_request(
     request_type: &RequestType,
     workers: Option<&WorkerSelection>,
 ) {
-    if matches!(request, ProtoGenerateRequest::Trtllm(_)) {
+    if matches!(
+        request,
+        ProtoGenerateRequest::Trtllm(_) | ProtoGenerateRequest::TokenSpeed(_)
```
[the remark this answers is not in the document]
key4ng: Fixed in `c478f1ac`. Dropped TokenSpeed from the early-return, added a
`ProtoGenerateRequest::TokenSpeed(req)` match arm calling a new
`apply_tokenspeed_sampling_defaults` helper. Defaults now flow
end-to-end: `translate::model_info` populates
`default_sampling_params_json` (sibling thread), discovery flattens it
into the label map, and the injector applies the masked fields with
`Some(value)` per TokenSpeed's `optional` scalar shape.

model_gateway/src/routers/grpc/common/stages/helpers.rs
```diff
@@ -119,7 +119,10 @@ pub(crate) fn apply_sampling_defaults_to_generate_request(
...
+    ) {
         return;
     }
```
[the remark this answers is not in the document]
key4ng: Fixed in `c478f1ac` rather than deferred — good catch on the
regression. `apply_sampling_defaults_to_generate_request` no longer
early-returns for TokenSpeed; there's a dedicated
`apply_tokenspeed_sampling_defaults` mirroring the SGLang/vLLM/MLX
helpers but writing `Some(value)` because every TokenSpeed
`SamplingParams` scalar is `optional` on the wire (the servicer's
`HasField()` checks rely on that set-vs-unset distinction).

Paired with the `translate::model_info` fix in the same commit so
`default_sampling_params_json` carries the JSON the discovery path
expects.

crates/grpc_client/src/tokenspeed_scheduler.rs
```diff
@@ -0,0 +1,542 @@
...
+            supports_vision: false,
+            model_type: r.model_type,
+            eos_token_ids: r.eos_token_ids,
+            pad_token_id: r.pad_token_id,
+            bos_token_id: r.bos_token_id,
+            max_req_input_len: r.max_req_input_len,
+            architectures: r.architectures,
+            id2label_json: String::new(),
+            num_labels: 0,
+            default_sampling_params_json: String::new(),
```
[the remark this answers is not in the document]
key4ng: Fixed in `c478f1ac`. `translate::model_info` now writes
`r.preferred_sampling_params` to **both** the legacy
`preferred_sampling_params` slot **and** `default_sampling_params_json`,
so the worker-discovery flatten path surfaces TokenSpeed-reported
defaults as a label the request stage can read. Paired with a
`TokenSpeed` arm in `apply_sampling_defaults_to_generate_request` so
the defaults actually get applied.

crates/grpc_client/src/lib.rs
```diff
@@ -18,6 +20,12 @@ use std::sync::Arc;
 
 pub use mlx_engine::{proto as mlx_proto, MlxEngineClient};
 pub use sglang_scheduler::{proto as sglang_proto, SglangSchedulerClient};
+// TokenSpeed has a fully independent wire definition (see
+// ``proto/tokenspeed_scheduler.proto``) — distinct service, distinct
+// messages with intentionally trimmed field sets aimed at top-tier LLM
+// workloads. The client wraps that wire and translates to/from SGLang-shaped
+// types at the boundary so the router's dispatch enums don't proliferate.
+pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
```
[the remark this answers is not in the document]
key4ng: Dropped in `e93dca99`. The 5-line rationale block is gone; the design
notes already live in `tokenspeed_scheduler.rs`'s module doc, so the
re-export line is now just:

```rust
pub use tokenspeed_scheduler::{tokenspeed_proto, TokenSpeedSchedulerClient};
```

crates/grpc_client/src/lib.rs
```diff
@@ -8,8 +8,10 @@ pub mod common_proto {
     tonic::include_proto!("smg.grpc.common");
 }
 pub mod mlx_engine;
+pub mod sampling_params;
 pub mod sglang_scheduler;
 pub mod tokenizer_bundle;
+pub mod tokenspeed_scheduler;
```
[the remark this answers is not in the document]
key4ng: Fixed in `e93dca99`. Collapsed the duplicate 2-line crate doc into a
single neutral line that covers all supported backends:

```rust
//! gRPC clients for the supported inference backends.
```

Same approach I used for `proto_wrapper.rs` earlier — avoids the
listing drift you flagged here.

PR #1352, opened 2026-04-23
feat(gateway): implement R6 image_generation tool integration

slin1237:
## Summary

Implements **audit §R6** — end-to-end runtime support for the `image_generation`
built-in tool built on top of T4's merged protocol types. Covers MCP routing,
streaming events, and request-time tool-argument override merging.

PR #1091 (still open) attempted the same thing with some scope creep. This PR
keeps the load-bearing behavior (compactor, overrides, transformer, streaming
events, MCP wiring) and intentionally drops #1091's defensive gRPC rejection
scaffolding and speculative field additions.

Refs: audit §R6; provenance: referenced PR #1091

### Protocols (`crates/protocols/src/event_types.rs`, +39 lines)
- `ImageGenerationCallEvent { InProgress, Generating, PartialImage, Completed }`
  with constant strings `response.image_generation_call.{in_progress|generating
  |partial_image|completed}`, `as_str()`, `Display`.
- `ItemType::ImageGenerationCall` + `IMAGE_GENERATION_CALL`. Included in
  `is_builtin_tool_call()`.

### MCP transform (`crates/mcp/src/transform/{types,transformer}.rs`, +431 lines)
- `ResponseFormat::ImageGenerationCall` variant.
- `ResponseFormat::compact_tool_output_for_model_context(&self, &Value) -> String`
  — for image_generation strips `result`/`data` (base64 payloads) from both flat
  and text-embedded content blocks, so subsequent model turns see
  `status`/`revised_prompt` instead of multi-MB base64. Other formats pass
  through unchanged.
- `ResponseTransformer::to_image_generation_call` emits
  `ResponseOutputItem::ImageGenerationCall { id, result, revised_prompt, status }`
  — uses **only** T4's four fields, no speculative additions. Tolerates three
  MCP payload shapes (text-wrapped JSON, flat, bare object) and marks result as
  `Failed` rather than smuggling error text into the base64 slot.
- 4 new transformer tests + 3 new compactor tests.

### MCP core config (`crates/mcp/src/core/config.rs`, +18 lines)
- `BuiltinToolType::ImageGeneration` + `ResponseFormatConfig::ImageGenerationCall`
  wired through `response_format()`, `Display`, and serde-roundtrip tests.
- `session.rs` already handled `ResponseOutputItem::ImageGenerationCall` (landed
  with T4), no change needed.

### Router common — `tool_overrides.rs` (NEW, 225 lines)
- `request_tool_overrides(response_format, request) -> Option<Value>` — for
[...]

PR #1355, opened 2026-04-23
feat(gateway): add image_generation infrastructure for hosted-tool MCP plumbing (R6.1)

slin1237:
## Summary

Audit item R6 of `.claude/_audit/responses-api-gap-audit.md` calls for
`image_generation` to be plumbed through the same hosted-tool MCP path
already wired for `web_search` / `code_interpreter` / `file_search`.

This PR is **1 of 4** — it adds only the **shared infrastructure** so
PRs **R6.2**, **R6.3**, and **R6.4** (router-specific event emission:
OpenAI router, harmony gRPC, regular gRPC) can be built in parallel on
top.

Supersedes PR #1352 (closed) and PR #1091; avoids their `unreachable!()`
stubs by using real neutral values that keep runtime safe until real
wiring lands.

## What's in this PR

Layered additions, each mirroring the existing shape of the other
three hosted tools:

### Protocol layer (`crates/protocols/src/event_types.rs`)
- New `ImageGenerationCallEvent { InProgress, Generating,
  PartialImage, Completed }` enum with
  `response.image_generation_call.*` constants, `as_str()`, and
  `Display`. Spec: OpenAI SDK v2.8.1
  `response_image_gen_call_*_event.py` +
  `.claude/_audit/openai-responses-api-spec.md`.
- `ItemType::ImageGenerationCall` + `IMAGE_GENERATION_CALL` constant;
  `is_builtin_tool_call()` extended.

### MCP config layer (`crates/mcp/src/core/config.rs`)
- `BuiltinToolType::ImageGeneration` (`Display = "image_generation"`,
  `response_format()` → `ResponseFormatConfig::ImageGenerationCall`).
- `ResponseFormatConfig::ImageGenerationCall` variant.
- Existing exhaustive-variant serde tests extended.

### MCP transform layer (`crates/mcp/src/transform/`)
- `ResponseFormat::ImageGenerationCall` variant + `From`
  `ResponseFormatConfig` arm + serde round-trip test.
- `ResponseTransformer::transform` dispatches to
  `to_image_generation_call(result, tool_call_id)` which maps MCP
  `CallToolResult` → `ResponseOutputItem::ImageGenerationCall` using
  **only** T4's actual fields: `id`, `result` (base64),
  `revised_prompt?`, `status`. No invented fields.
- Extractor probes direct object fields AND embedded
  `openai_response` text-block payloads (mirrors
  `to_web_search_call`).
- New `compact_image_generation_output(item)` helper strips the
  base64 `result` payload for stored multi-turn context (no-op for
  non-image items). Re-exported from `transform` and the crate root.
- Unit tests: direct fields, `b64_json` alias, embedded text blocks,
  missing-image fallback, compactor strip + no-op.

### Router shared streaming
(`model_gateway/src/routers/grpc/common/responses/streaming.rs`)
- `emit_tool_call_in_progress` / `emit_tool_call_searching` (emits
  `generating` as intermediate) / `emit_tool_call_completed` now
  handle `ImageGenerationCall`.
- New `emit_image_generation_partial_image(...)` helper for the
  image-specific `partial_image` event with
  `partial_image_index` + `partial_image_b64` payload. Gated on
  format; per-router wiring in R6.2/R6.3/R6.4 decides when to call
  it. Marked `#[expect(dead_code, reason="...")]` until then.
- `type_str_for_format` / `output_item_type_for_format` /
  `OutputItemType` variants + `allocate_output_index` id prefix
  `"ig"` all extended.

### Router common util
(`model_gateway/src/routers/common/mcp_utils.rs`)
- `collect_builtin_routing` and `extract_builtin_types` recognize
  `ResponseTool::ImageGeneration(_)` →
  `BuiltinToolType::ImageGeneration`.
- New test `test_collect_builtin_routing_image_generation`.

### Stub arms in exhaustive ResponseFormat matches
(`model_gateway/src/routers/openai/mcp/tool_loop.rs`)

Option X — `ResponseFormat` remains non-`#[non_exhaustive]`, so the
compiler still enforces exhaustiveness at every site. Stubs use real
neutral values (no `todo!()`/`unreachable!()`/`panic!()`):

| Site | Stub behavior |
|------|---------------|
| `send_tool_call_intermediate_event` | emit `ImageGenerationCallEvent::GENERATING` |
| `send_tool_call_completion_events` | map `ItemType::IMAGE_GENERATION_CALL` → `ImageGenerationCallEvent::COMPLETED` |
| `stable_streaming_tool_item_id` | prefix item ids with `"ig_"` |
| `non_streaming_tool_item_id_source` | join existing `strip_prefix("fc_"\|"call_")` arm |

Sites using `matches!(...)` / `_ =>` wildcards (harmony streaming,
regular gRPC streaming, OpenAI `responses::streaming`) do **not**
require stubs because the compiler does not enforce exhaustiveness
there. They preserve their existing default behavior; PRs
R6.2/R6.3/R6.4 own extending them.

## Why / How

- Mirrors the established pattern for the other three hosted tools
  rather than inventing a new one — reviewers only need to verify
  the pattern was applied consistently.
- Option X (exhaustive matches + stubs) was chosen over
  `#[non_exhaustive]` so the compiler keeps enforcing coverage at
  every site as new hosted tools are added in the future.
- The transformer's `result: String` falls back to an empty string
  when the MCP server returns no image data; per-router code in
  R6.2/R6.3/R6.4 owns surfacing a user-visible error status.
- `compact_image_generation_output` is needed so multi-turn storage
  doesn't balloon with base64 image payloads; the input-side
  `ResponseInputOutputItem::ImageGenerationCall` already allows
  `result` to be absent (id-only reference), so this is a
  spec-compatible storage optimization.

## Non-goals for R6.1

Explicitly left for the follow-up PRs so this PR stays small and
reviewable:

- No real event emission in per-router streaming paths beyond what
  the shared emitter provides. Harmony and regular gRPC streaming
  paths are untouched.
- No MCP server-config additions beyond the new `BuiltinToolType`
  variant — operators register their image-gen MCP server through
  existing config mechanisms.
- No changes to `crates/protocols/src/responses.rs` (T4 owns those
  types).
- No gRPC BUILTIN_TOOLS / harmony tool-registration changes — that
  is being handled separately in PR #1353.
- No `model_gateway/src/routers/common/tool_overrides.rs` —
  reviewed and deemed out-of-scope for R6.1 infrastructure; if
  caller-pinned `ResponseTool::ImageGeneration(...)` config merging
  is needed, it'll land as a follow-up PR or inside R6.2/R6.3/R6.4.

## Test plan

- [x] `cargo check -p openai-protocol -p smg-mcp -p smg --lib --tests --benches`
- [x] `cargo test -p openai-protocol -p smg-mcp -p smg --lib` — 855
      tests pass, 8 new image_generation tests green:
      - `test_image_generation_transform_direct_fields`
      - `test_image_generation_transform_b64_json_alias`
      - `test_image_generation_transform_embedded_openai_response`
      - `test_image_generation_transform_missing_image_data`
      - `test_compact_image_generation_output_strips_base64`
      - `test_compact_image_generation_output_is_noop_for_other_items`
      - `test_collect_builtin_routing_image_generation`
      - serde round-trips for the three new enum variants
- [x] `cargo fmt --all`
- [x] `cargo clippy -p openai-protocol -p smg-mcp -p smg --lib --tests --benches -- -D warnings`

<details>
<summary>Checklist</summary>

- [x] Tests added
- [x] Commit signed-off (DCO)

</details>

Refs: R6 in `.claude/_audit/responses-api-gap-audit.md`

## Summary by CodeRabbit

* **New Features**
  * Image generation added as a built-in tool with full streaming lifecycle: in‑progress, generating, partial‑image updates, and completion events.
  * Partial image updates are emitted during generation so previews can be displayed progressively.
  * Image results are compacted to reduce payload size across conversation turns.
  * Stable, predictable IDs used for image-generation outputs for reliable tracking.

### DIFF
--- a/crates/mcp/src/core/config.rs
+++ b/crates/mcp/src/core/config.rs
@@ -302,6 +302,8 @@ pub enum BuiltinToolType {
     CodeInterpreter,
     /// File search tool (OpenAI: file_search)
     FileSearch,
+    /// Image generation tool (OpenAI: image_generation)
+    ImageGeneration,
 }
 
 impl BuiltinToolType {
@@ -311,6 +313,7 @@ impl BuiltinToolType {
             BuiltinToolType::WebSearchPreview => ResponseFormatConfig::WebSearchCall,
             BuiltinToolType::CodeInterpreter => ResponseFormatConfig::CodeInterpreterCall,
             BuiltinToolType::FileSearch => ResponseFormatConfig::FileSearchCall,
+            BuiltinToolType::ImageGeneration => ResponseFormatConfig::ImageGenerationCall,
         }
     }
 }
@@ -321,6 +324,7 @@ impl fmt::Display for BuiltinToolType {
             BuiltinToolType::WebSearchPreview => write!(f, "web_search_preview"),
             BuiltinToolType::CodeInterpreter => write!(f, "code_interpreter"),
             BuiltinToolType::FileSearch => write!(f, "file_search"),
+            BuiltinToolType::ImageGeneration => write!(f, "image_generation"),
         }
     }
 }
@@ -350,6 +354,7 @@ pub enum ResponseFormatConfig {
     WebSearchCall,
     CodeInterpreterCall,
     FileSearchCall,
+    ImageGenerationCall,
 }
 
 /// Argument mapping configuration for tool aliases.
@@ -1025,6 +1030,10 @@ tools:
                 "\"code_interpreter_call\"",
             ),
             (ResponseFormatConfig::FileSearchCall, "\"file_search_call\""),
+            (
+                ResponseFormatConfig::ImageGenerationCall,
+                "\"image_generation_call\"",
+            ),
         ];
 
         for (format, expected) in formats {
@@ -1191,6 +1200,7 @@ policy:
             (BuiltinToolType::WebSearchPreview, "\"web_search_preview\""),
             (BuiltinToolType::CodeInterpreter, "\"code_interpreter\""),
             (BuiltinToolType::FileSearch, "\"file_search\""),
+            (BuiltinToolType::ImageGeneration, "\"image_generation\""),
         ];
 
         for (builtin_type, expected) in types {
@@ -1215,6 +1225,10 @@ policy:
         assert_eq!(
             BuiltinToolType::FileSearch.response_format(),
             ResponseFormatConfig::FileSearchCall
+        );
+        assert_eq!(
+            BuiltinToolType::ImageGeneration.response_format(),
+            ResponseFormatConfig::ImageGenerationCall
         );
     }
 
--- a/crates/mcp/src/lib.rs
+++ b/crates/mcp/src/lib.rs
@@ -57,5 +57,6 @@ pub use responses_bridge::{
 pub use tenant::{SessionId, TenantContext, TenantId};
 // Re-export from transform
 pub use transform::{
-    extract_embedded_openai_responses, mcp_response_item_id, ResponseFormat, ResponseTransformer,
+    compact_image_generation_output, extract_embedded_openai_responses, mcp_response_item_id,
+    ResponseFormat, ResponseTransformer,
 };
--- a/crates/mcp/src/transform/mod.rs
+++ b/crates/mcp/src/transform/mod.rs
@@ -23,6 +23,7 @@ mod transformer;
 mod types;
 
 pub use transformer::{
-    extract_embedded_openai_responses, mcp_response_item_id, ResponseTransformer,
+    compact_image_generation_output, extract_embedded_openai_responses, mcp_response_item_id,
+    ResponseTransformer,
 };
 pub use types::ResponseFormat;
--- a/crates/mcp/src/transform/transformer.rs
+++ b/crates/mcp/src/transform/transformer.rs
@@ -516,7 +516,12 @@ impl ResponseTransformer {
 /// For any non-image item this is a no-op.
 pub fn compact_image_generation_output(item: &mut ResponseOutputItem) {
     if let ResponseOutputItem::ImageGenerationCall { result, .. } = item {
-        result.clear();
+        // `result.clear()` zeros the length but keeps the heap buffer
+        // allocated — for base64 image bytes that can be several MB of
+        // wasted capacity per stored item, defeating the compaction goal.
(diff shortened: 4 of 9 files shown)


slin1237:
Thanks for the review! Addressed both points in commit 05e4ea0c:

1. **Logic duplication** — refactored into an `update_fields` closure; direct-object and embedded-text-block paths now share one probe.
2. **Data loss when fields are split across blocks** — switched from "early-return on first hit" to "accumulate across all sources". Added `test_image_generation_transform_fields_distributed_across_blocks` as regression guard.

First occurrence still wins for each slot; direct object lookup runs before embedded blocks so its values win when both are present.

slin1237:
Thanks for the catch! Addressed in commit 508c78da — switched to `*result = String::new()` so the base64 backing buffer is actually released. Existing unit tests still cover the strip + no-op behavior.

slin1237:
@codex re: [P1 on `type_str_for_format`](https://github.com/lightseekorg/smg/pull/1355#discussion_r...) — this is the explicit non-goal of R6.1 called out in the PR description: "No real event emission in per-router streaming paths beyond what the shared emitter provides. PRs R6.2/R6.3/R6.4 wire router-specific sites."

Concretely, R6.3 (harmony gRPC) and R6.4 (regular gRPC) are both responsible for replacing the generic `json!({name, arguments, status})` item builders at their call sites (regular gRPC L653-739, harmony L842-L970) with the `image_generation_call`-shaped builder (`{id, result, revised_prompt?, status}`). The shared transformer's `to_image_generation_call` in `crates/mcp/src/transform/transformer.rs` is the canonical helper they'll reach for — it already produces the correctly-shaped `ResponseOutputItem::ImageGenerationCall`.

The failure mode you describe (malformed `image_generation_call` items dropped by `finalize()`) is gated on an operator configuring `BuiltinToolType::ImageGeneration` in their MCP config, which is itself net-new in this PR. No existing flows can reach it on `main`; by the time R6.3/R6.4 land, the per-router shaping will be in place before anyone has cause to enable the config.

R6.2 (OpenAI router) is unaffected because `tool_loop.rs` builds its items via `ResponseTransformer::transform` (see `build_transformed_mcp_call_item` callers) rather than via `type_str_for_format`.

PR #1370, opened 2026-04-24
feat(mcp): add hosted-tool overrides helpers and wire image_generation dispatch (R6.6)

slin1237:
### Problem
R6.1–R6.4 (#1355, #1356, #1358, #1359) added image_generation plumbing end-to-end: `ResponseFormat::ImageGenerationCall`, `BuiltinToolType::ImageGeneration`, streaming event emitters per router, and the MCP transformer. The MCP dispatch path still **drops caller-declared hosted-tool configuration**. A request declaring

\`\`\`json
{ "tools": [{ "type": "image_generation", "size": "512x512", "quality": "high" }] }
\`\`\`

paired with a model that emits `{prompt}` reaches the MCP server as just `{prompt}` — `size` and `quality` are silently lost. R6.5's e2e probe on #1365 (`test_image_generation_tool_overrides_size`) confirmed the gap: the mock MCP server never saw `size=512x512`.

### Solution
Two pure helpers in `smg-mcp` + per-router wiring at every MCP dispatch site.

- Helpers live in `smg-mcp` because the crate already depends on `openai-protocol` and already contains the inverse direction (`ResponseTransformer::to_image_generation_call`: MCP result → `ResponseOutputItem`). The extract-and-merge path is the symmetric complement.
- Pure functions (not `McpToolSession` methods) because the merge needs no session state: `&[ResponseTool] + BuiltinToolType → Option<Value>`, then in-place merge into dispatch args. This keeps `call_tool`'s contract clean and makes the helpers unit-testable in isolation.
- Router wiring is narrow: each call site maps its per-tool `ResponseFormat` back to `BuiltinToolType` via a new `ResponseFormat::to_builtin_tool_type()` and calls the two helpers before `session.execute_tool[s]`.

### `smg-mcp` helpers (commit 1)

- **`crates/mcp/src/transform/overrides.rs`** (new):
  - `pub fn extract_hosted_tool_overrides(tools: &[ResponseTool], kind: BuiltinToolType) -> Option<Value>` — finds the first matching declaration, serializes the inner struct, drops null fields, drops the `type` discriminator, returns `None` if empty. Image_generation is the only kind with caller-configurable fields today; the other arms stay in place so they pick up new overridable fields automatically as those tool structs grow.
  - `pub fn apply_hosted_tool_overrides(arguments: &mut Value, overrides: &Value)` — in-place merge; caller-declared keys win over model-emitted keys; no-op for non-object inputs.
  - 11 unit tests: null-drop, empty→None, missing-declaration→None, first-match-after-non-matching, caller-wins, args extension, non-object args/overrides no-ops, empty-overrides no-op, and coverage of the other three hosted kinds.
- **`crates/mcp/src/transform/types.rs`**: `ResponseFormat::to_builtin_tool_type() -> Option<BuiltinToolType>` as the inverse of `BuiltinToolType::response_format()`, plus a round-trip test.
- **`crates/mcp/src/transform/mod.rs`** and **`crates/mcp/src/lib.rs`**: top-level re-exports following the existing transform re-export convention.

### Router wiring (commit 2)

All dispatch loci now merge caller-declared overrides before `session.execute_tool[s]`:

- **`openai/mcp/tool_loop.rs`** — streaming (`execute_streaming_tool_calls` gains a `request_tools: &[ResponseTool]` parameter) and non-streaming (`execute_tool_loop` reads `original_body.tools`).
- **`openai/responses/streaming.rs`** — threads `original_request.tools` into `execute_streaming_tool_calls`.
- **`grpc/harmony/responses/execution.rs`** — `execute_mcp_tools` gains `request_tools: &[ResponseTool]` and merges inside the `ToolExecutionInput` build loop.
- **`grpc/harmony/responses/{streaming,non_streaming}.rs`** — call sites pass `original_request.tools` / `original_tools` respectively (both preserved before MCP tool injection).
- **`grpc/regular/responses/non_streaming.rs`** — tool-loop `ToolExecutionInput` build loop merges from `original_request.tools`.
- **`grpc/regular/responses/streaming.rs`** — single-tool `execute_tool` path merges from `original_request.tools`.

Non-hosted dispatch is unaffected: `ResponseFormat::Passthrough` maps to `None` and the merge short-circuits. User-registered (non-builtin) MCP tool dispatch paths are guarded by the same check — no forced cascade.

## Non-goals

- No compactor changes (R6.1 already covers).
- No new event types (R6.1 covers).
- No fix for the gRPC harmony gpt-oss dispatch gap R6.5 found (separate — belongs in R6.7).
- No fix for R6.2's event-ordering / empty-\`result\` bugs R6.5 found (separate R6.x follow-up).

## Test Plan

- [x] \`cargo check -p openai-protocol -p smg-mcp -p smg --lib --tests --benches\`
- [x] \`cargo test -p smg-mcp --lib\` — 195 tests pass (12 new in transform::overrides + transform::types).
- [x] \`cargo test -p smg --lib\` — 633 tests pass.
- [x] \`cargo fmt --all\`
- [x] \`cargo clippy -p openai-protocol -p smg-mcp -p smg --lib --tests --benches -- -D warnings\` — clean.
- [ ] R6.5 e2e (#1365) — once this lands and #1365 rebases, \`test_image_generation_tool_overrides_size\` should pass (the mock MCP server will see \`size=512x512\` as an effective arg).

Refs: R6 §tool-overrides (\`.claude/_audit/responses-api-gap-audit.md\`); R6.5 #1365

## Summary by CodeRabbit

* **New Features**
  * Tool execution now merges caller-declared hosted-tool configuration overrides into dispatch arguments across streaming and non-streaming flows and protocols, ensuring caller-specified tool settings are applied when tools run.
  * Tool argument payloads are normalized so overrides consistently take effect, and logged/approved payloads reflect the effective (post-merge) arguments.

### DIFF
--- a/crates/mcp/src/lib.rs
+++ b/crates/mcp/src/lib.rs
@@ -57,6 +57,7 @@ pub use responses_bridge::{
 pub use tenant::{SessionId, TenantContext, TenantId};
 // Re-export from transform
 pub use transform::{
-    compact_image_generation_output, extract_embedded_openai_responses, mcp_response_item_id,
+    apply_hosted_tool_overrides, compact_image_generation_output,
+    extract_embedded_openai_responses, extract_hosted_tool_overrides, mcp_response_item_id,
     ResponseFormat, ResponseTransformer,
 };
--- a/crates/mcp/src/transform/mod.rs
+++ b/crates/mcp/src/transform/mod.rs
@@ -19,9 +19,11 @@
 //! );
 //! ```
 
+mod overrides;
 mod transformer;
 mod types;
 
+pub use overrides::{apply_hosted_tool_overrides, extract_hosted_tool_overrides};
 pub use transformer::{
     compact_image_generation_output, extract_embedded_openai_responses, mcp_response_item_id,
     ResponseTransformer,
--- a/crates/mcp/src/transform/overrides.rs
+++ b/crates/mcp/src/transform/overrides.rs
@@ -1,0 +1,283 @@
+//! Pure-function helpers that merge caller-declared hosted-tool configuration
+//! into model-emitted tool-call arguments before MCP dispatch.
+//!
+//! # Motivation
+//!
+//! OpenAI's Responses API lets callers declare hosted-tool configuration on the
+//! request, e.g.:
+//!
+//! ```json
+//! { "tools": [{ "type": "image_generation", "size": "512x512", "quality": "high" }] }
+//! ```
+//!
+//! When the model emits the hosted-tool call, the arguments field often
+//! contains only the dynamic subset (e.g. `{"prompt": "a cat"}`). The remaining
+//! caller-declared knobs (`size`, `quality`, `model`, ...) live on the
+//! `tools` declaration and must be merged into the dispatch payload so the
+//! MCP server receives the effective configuration.
+//!
+//! These helpers are intentionally:
+//! - **pure**: no session state, no I/O; testable in isolation.
+//! - **symmetric** with [`crate::transform::transformer`], which handles the
+//!   inverse direction (MCP result → `ResponseOutputItem`).
+//!
+//! # Contract
+//!
+//! [`extract_hosted_tool_overrides`] walks `tools` for the first declaration of
+//! the requested kind, serializes it to `Value::Object`, drops null entries
+//! (so absent fields don't overwrite model-supplied values), strips the `type`
+//! discriminator (the kind is already known to the caller), and returns the
+//! remaining object — or `None` if no configurable fields remain.
+//!
+//! [`apply_hosted_tool_overrides`] mutates the dispatch args in-place, with
+//! overrides winning over model args for overlapping keys. Non-object or empty
+//! inputs are no-ops.
+
+use openai_protocol::responses::ResponseTool;
+use serde_json::Value;
+
+use crate::core::config::BuiltinToolType;
+
+/// Extract caller-declared configuration for the given hosted-tool kind from
+/// the request's `tools` declarations.
+///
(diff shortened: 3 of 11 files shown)


slin1237:
Review feedback addressed in c592c4e8:

**@claude** — fixed in both streaming + non-streaming paths. `debug!` now logs the effective post-merge `arguments` Value, and the non-streaming path serializes merged args once into `effective_arguments` for both the log and the approval payload.

**@coderabbitai** — adopted the defensive coercion at every hosted-tool dispatch site (openai streaming/non-streaming, gRPC harmony `execute_mcp_tools`, gRPC regular streaming + non-streaming). Non-object model payloads now coerce to `{}` before override merge, so a malformed LLM emission never silently drops caller-declared hosted-tool config.

**@chatgpt-codex-connector — approval payload (openai tool_loop:916)** — fixed. `build_mcp_approval_request_item` now receives the serialized post-merge `effective_arguments` string, so users approve the payload that will actually dispatch rather than the pre-merge string.

**@chatgpt-codex-connector — ResponseFormat::Passthrough → None (types.rs:34)** — this is by design and safe in the current architecture. A config that sets `builtin_type: image_generation` on a server always derives a non-Passthrough `ResponseFormatConfig` via `BuiltinToolType::response_format()`, so dispatch for a built-in tool is never `Passthrough`. The only way the described scenario manifests is if a user explicitly overrides `response_format: passthrough` in `ToolConfig` for a server that is also flagged `builtin_type: image_generation` — which is a misconfiguration outside R6.6's scope and would require a separate decision about whether the explicit override wins or the implicit builtin mapping wins. Flagging for a follow-up if we decide to forbid that combo in `McpServerConfig::validate`.

PR #1386, opened 2026-04-24
fix(mcp): response_format runtime override + user forwarding + regression e2e (R7)

slin1237:
### Problem

An external reviewer tested our `image_generation` MCP integration at
commit `cb84bbb8` (R6.5) and reported three real spec violations. All
three are gateway-side bugs, not user-config issues.

#### Bug 1 — `user` not forwarded to MCP dispatch

`ResponsesRequest.user` is mirrored into `safety_identifier` for OpenAI
cloud forwarding (`model_gateway/src/routers/openai/responses/utils.rs:91`)
but never reaches the MCP server. MCP proxies downstream of SMG cannot
attribute usage or enforce per-user quotas.

Repro: send a Responses request with `user: "u1"` and a hosted
`image_generation` tool declaration. Observed: the MCP tool call
arguments contain `prompt`/`size`/etc. but no `user`.

#### Bug 2+3 — Response item shape is wrong when the MCP server lacks `builtin_type`

When an MCP server exposes an `image_generation` tool but its YAML
config does NOT tag it with `builtin_type: image_generation`, the
emitted output item is shaped as a generic `mcp_call`:

```json
{"id": "...", "type": "image_generation_call", "output": "<nested json with base64>", "arguments": "<metadata json>", ...}
```

Correct spec shape (`openai-responses-api-spec.md §ImageGenerationCall`):

```json
{"id": "ig_...", "type": "image_generation_call", "result": "<base64>", "status": "completed",
 "size": "1024x1024", "quality": "...", "background": "...", "output_format": "...", "action": "...",
 "revised_prompt": "..."}
```

Repro: point the gateway at an MCP server without `builtin_type` in its
YAML config, send a Responses request with `tools:
[{"type": "image_generation", ...}]`, inspect the response output —
item type is present but with the `mcp_call` field layout (`output` +
`arguments`) instead of the spec-correct `result` + `status` + metadata
fields.

Root cause: `McpToolSession::tool_response_format(tool_name)` at
`crates/mcp/src/core/session.rs:461` resolves the format purely from
the session binding, which is built from the server-side `builtin_type`
config. If the server wasn't tagged, the format falls back to
`Passthrough` and `ResponseTransformer::transform` emits `mcp_call`
output. Per spec, the CLIENT's `tools` declaration determines the
output item shape; the server's `builtin_type` is an SMG-specific
routing knob and should NOT be the sole gate on the output shape.

The same class of bug affects all four hosted tools (image_generation,
web_search_preview, code_interpreter, file_search). This PR fixes the
general case; e2e coverage is scoped to image_generation because
that's what the reviewer exercised.

### Why the R6.5 test suite didn't catch these

The R6.5 `gateway_with_mock_mcp_cloud` fixture tagged the mock MCP
server with `builtin_type: image_generation` AND the client declared
the same hosted tool on the request. Both paths matched, so:
1. The session-side response format was already `ImageGenerationCall`
   at dispatch time — the runtime override path was never exercised.
2. There was no assertion on `user` reaching the MCP server, just that
   the tool ran.

This PR adds a parallel fixture and test class that breaks the
pre-matched setup, plus a new user-forwarding assertion mixed into
both the old and new fixtures.

#### Fix A — Runtime response_format override (one pure function, one source of truth)

* New pure function `resolve_response_format(session_format,
  request_tools, tool_name)` in `crates/mcp/src/core/session.rs`:
  * If the session already classifies the tool as a hosted format
    (server was explicitly tagged), keep it — explicit config wins.
  * Otherwise, if the request `tools` contains a hosted-tool
    declaration AND the dispatched tool name matches that hosted
    tool's canonical name (`image_generation`,
    `web_search_preview`/`web_search`, `code_interpreter`,
    `file_search`), return the matching hosted format.
  * Otherwise, return the session format unchanged.

* New `ToolExecutionInput.response_format_override:
  Option<ResponseFormat>` field (defaults to `None`, so existing
  call sites compile unchanged). When set,
  `execute_tool_resolved_result` replaces the tool entry's
  session-derived format on the resulting `ToolExecutionOutput` /
  `PendingToolExecution`. `ToolExecutionOutput::to_response_item`
  therefore emits the correct shape without per-site transformer
  rewrites.

* Wired at all five router execution sites:
  1. `grpc/harmony/responses/execution.rs::execute_mcp_tools`
  2. `grpc/regular/responses/non_streaming.rs` tool-loop batch
  3. `grpc/regular/responses/streaming.rs` per-call dispatch
  4. `openai/mcp/tool_loop.rs::execute_streaming_tool_calls`
  5. `openai/mcp/tool_loop.rs::execute_tool_loop` (non-streaming)

* Also updated `openai/responses/streaming.rs::apply_event_transformations_inplace`
  (cloud SSE rewrite layer) so the `fc_*` → hosted-item rewrite uses
  the resolved format — otherwise streaming events for a plain-MCP
  server would still emit `mcp_call` types.

#### Fix B — `user` forwarding into MCP dispatch args (one pure function, hosted-tools only)

OpenAI's spec does NOT list `user` as a field on `ImageGenerationTool`,
so we can't ride the existing `tool_overrides` (R6.6) path. Instead:

* New pure function `inject_user_into_hosted_tool_args(arguments,
  resolved_format, user)` in `crates/mcp/src/core/session.rs`:
  * No-op for `ResponseFormat::Passthrough` (plain MCP function
    tools are left alone — safer than potentially breaking an MCP
    tool schema that doesn't expect a `user` key).
  * No-op for empty/None `user`.
  * If `arguments` already has a `user` key, preserve the
    model-supplied value and log at `debug!` — don't overwrite.
  * Otherwise inject `"user": "<value>"` into the args object.

* Called at each of the five router sites after
  `apply_hosted_tool_overrides`.

### Out of scope (noted for follow-up)

* `grpc/harmony/streaming.rs` (the chat-completion-level streaming
  path) still uses the session-only format at its mid-tool event
  emission site. The function doesn't currently have access to
  `ResponsesRequest`; R7's cloud regression test doesn't exercise
  it (cloud routes through `openai/responses/streaming.rs`).
  Threading `request_tools` through the Harmony chat pipeline
  is a follow-up for when the plain-MCP scenario gets e2e coverage on
  the gRPC lanes.

* Parallel plain-MCP + user-forwarding e2e coverage for
  `web_search_preview`, `code_interpreter`, `file_search`. The
  Fix A / Fix B logic already handles all four hosted-tool kinds
  uniformly — only the e2e test coverage is deferred to when each
  tool gets its own R6.x-style mock tool + test suite.

### Rust / MCP crate

* `crates/mcp/src/core/session.rs` — adds `resolve_response_format`
  and `inject_user_into_hosted_tool_args` pure functions, plus 14
  unit tests covering every branch (session-wins, passthrough
  promotion, tool-name mismatch, four hosted kinds, user injection
  happy path, skipped-for-passthrough, skipped-for-empty,
  preserve-existing-key, non-object no-op).
* `crates/mcp/src/core/orchestrator.rs` — adds
  `ToolExecutionInput.response_format_override` and the override
  plumbing inside `execute_tool_resolved_result`.
* `crates/mcp/src/core/mod.rs` + `crates/mcp/src/lib.rs` —
  re-exports for both new pure functions.
* `crates/mcp/src/core/session.rs` — propagates `response_format_override`
  through `execute_tool_result` for the resolved-name hop.

### Router execution sites

* `model_gateway/src/routers/grpc/harmony/responses/execution.rs`
  + `non_streaming.rs` + `streaming.rs`: thread `request_tools` +
  `request_user` through; call `resolve_response_format` and
  `inject_user_into_hosted_tool_args`; attach override to
  `ToolExecutionInput`.
* `model_gateway/src/routers/grpc/regular/responses/non_streaming.rs`
  + `streaming.rs`: same.
* `model_gateway/src/routers/openai/mcp/tool_loop.rs`: same at both
  streaming and non-streaming dispatch sites. Adds `request_user`
  param to `execute_streaming_tool_calls`.
* `model_gateway/src/routers/openai/responses/streaming.rs`:
  resolves format for the SSE item-type rewrite.
* `model_gateway/src/routers/anthropic/mcp.rs`: `response_format_override: None`
  for the existing `ToolExecutionInput` construction (no behavior
  change on the Anthropic path).

### E2E tests

* `e2e_test/responses/conftest.py`: new `_plain_mcp_config`,
  `mock_mcp_config_file_plain`, `gateway_with_mock_mcp_plain_cloud`
  fixture — same mock server, no `builtin_type` tagging.
* `e2e_test/responses/test_image_generation.py`:
  * New `TestImageGenerationClientDeclaredServerPlain` class that
    reuses the full `_ImageGenerationAssertions` body (4 tests) +
    the new user-forwarding assertion.
  * New `_UserForwardingAssertions` mix-in with
    `test_image_generation_forwards_user_to_mcp`, mixed into both
    `TestImageGenerationCloud` and the new plain-server class (so
    we exercise `user` on both server configs).
* `e2e_test/infra/mock_mcp_server.py`: `image_generation` tool
  accepts `user: str | None` and records it in `call_log`.

## Test plan

* `cargo check --workspace --all-targets` — green.
* `cargo clippy --workspace --all-targets -- -D warnings` — green.
* `cargo test -p smg-mcp --lib` — 215 tests pass (14 new).
* `cargo test -p smg` — all existing tests pass.
* `python3.11 -m ruff check e2e_test/responses/ e2e_test/infra/` — clean.
* `python3.11 -m ruff format --check e2e_test/responses/ e2e_test/infra/` — clean.
* New e2e test classes collect successfully:
  * `TestImageGenerationCloud` — 5 tests (4 existing + 1 user-forwarding).
  * `TestImageGenerationClientDeclaredServerPlain` — 5 tests (4 happy-path + 1 user-forwarding) against the plain-MCP fixture.
  * `TestImageGenerationGrpcSglang` / `TestImageGenerationGrpcVllm` — 4 tests each (unchanged).
* Full cloud e2e runs require `OPENAI_API_KEY`; CI will exercise both
  `TestImageGenerationCloud` and `TestImageGenerationClientDeclaredServerPlain`
  on the cloud runner.

Refs: R7

<details>
<summary>Checklist</summary>

- [x] Documentation updated (inline doc comments for both pure functions)

</details>

## Summary by CodeRabbit

* **New Features**
  * Per-call response format resolution for tool calls so outputs match session and client-declared tool intent across server types.
  * Forwarding of caller user context into hosted-tool dispatch arguments to enable per-user attribution.

* **Tests**
  * Added end-to-end tests verifying user forwarding and response-format behavior with mock MCP server configurations.

### DIFF
--- a/crates/mcp/src/core/mod.rs
+++ b/crates/mcp/src/core/mod.rs
@@ -27,4 +27,7 @@ pub use orchestrator::{
 };
 pub use pool::{McpConnectionPool, PoolKey};
 pub use reconnect::ReconnectionManager;
-pub use session::{McpServerBinding, McpToolSession, DEFAULT_SERVER_LABEL};
+pub use session::{
+    inject_user_into_hosted_tool_args, resolve_response_format, McpServerBinding, McpToolSession,
+    DEFAULT_SERVER_LABEL,
+};
--- a/crates/mcp/src/core/orchestrator.rs
+++ b/crates/mcp/src/core/orchestrator.rs
@@ -204,7 +204,7 @@ impl ToolCallResult {
 ///
 /// This is a simplified input format that allows routers to batch
 /// multiple tool calls without worrying about the underlying execution details.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct ToolExecutionInput {
     /// Unique identifier for this tool call (from LLM response).
     pub call_id: String,
@@ -212,6 +212,16 @@ pub struct ToolExecutionInput {
     pub tool_name: String,
     /// Tool arguments as JSON.
     pub arguments: Value,
+    /// Optional runtime override for the response format applied to the
+    /// output item transformation.
+    ///
+    /// When `Some`, this replaces the session-resolved `ResponseFormat` on the
+    /// resulting [`ToolExecutionOutput`]. The canonical producer is
+    /// [`crate::core::session::resolve_response_format`], which combines the
+    /// session's configured format with hosted-tool declarations from the
+    /// caller's `tools` list. Leaving this as `None` preserves the legacy
+    /// behavior (session state is authoritative).
+    pub response_format_override: Option<ResponseFormat>,
 }
 
 /// Output from batch tool execution.
@@ -1017,6 +1027,7 @@ impl McpOrchestrator {
 
         let qualified = QualifiedToolName::new(server_key, &input.tool_name);
         let entry = self.tool_inventory.get_entry(server_key, &input.tool_name);
+        let format_override = input.response_format_override;
 
         match entry {
             Some(entry) => match self
@@ -1029,6 +1040,9 @@ impl McpOrchestrator {
                     output.server_key = server_key.to_string();
                     output.server_label = server_label.to_string();
                     output.arguments_str = arguments_str;
+                    if let Some(override_format) = format_override {
+                        output.response_format = override_format;
+                    }
                     ToolExecutionResult::Executed(output)
                 }
                 ToolExecutionResult::PendingApproval(mut pending) => {
@@ -1037,6 +1051,9 @@ impl McpOrchestrator {
                     pending.server_key = server_key.to_string();
                     pending.server_label = server_label.to_string();
                     pending.arguments_str = arguments_str;
+                    if let Some(override_format) = format_override {
+                        pending.response_format = override_format;
+                    }
                     ToolExecutionResult::PendingApproval(pending)
                 }
             },
@@ -1054,7 +1071,7 @@ impl McpOrchestrator {
                     output: serde_json::json!({ "error": &err }),
                     is_error: true,
                     error_message: Some(err),
-                    response_format: ResponseFormat::Passthrough,
+                    response_format: format_override.unwrap_or(ResponseFormat::Passthrough),
                     duration: start.elapsed(),
                 })
             }
--- a/crates/mcp/src/core/session.rs
+++ b/crates/mcp/src/core/session.rs
@@ -896,15 +896,10 @@ impl<'a> McpToolSession<'a> {
     }
 }
 
-/// Canonical hosted-tool name that identifies a dispatched `image_generation` call.
-///
-/// R7: callers declare `tools: [{"type": "image_generation", ...}]`, but the
-/// MCP server may expose the tool under an alias or a synonym. This constant
-/// pins the set of aliases we treat as "this call is an image_generation
-/// hosted-tool call" for the purpose of inferring the response format when
-/// the server has no `builtin_type` config. Keep this list narrow — broader
-/// matching belongs in the MCP config's alias/tool_config mechanism, not
-/// here.
+/// Aliases we treat as an `image_generation` hosted-tool call for the
+/// purpose of inferring the response format when the MCP server has no
+/// `builtin_type` config. Keep this list narrow — broader matching belongs
+/// in the MCP config's alias/tool_config mechanism, not here.
 const HOSTED_IMAGE_GENERATION_NAMES: &[&str] = &["image_generation"];
 
 /// Canonical hosted-tool name for `web_search_preview` dispatches.
@@ -1946,7 +1941,7 @@ mod tests {
     }
 
     // ---------------------------------------------------------------
-    // R7: resolve_response_format + inject_user_into_hosted_tool_args
+    // resolve_response_format + inject_user_into_hosted_tool_args
     // ---------------------------------------------------------------
 
     fn default_image_generation_tool() -> ResponseTool {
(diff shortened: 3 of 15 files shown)


slin1237:
Closing — architecture is wrong. R7 invented a runtime response_format override + a cross-cutting user injection, which changes public API (McpToolSession, ToolExecutionInput) and contradicts the existing 'server config is authoritative' design used by web_search_preview / file_search / code_interpreter. The actual reviewer-reported bugs (output/arguments in wrong field) are caused by an MCP server misconfiguration — the server needed builtin_type: ImageGeneration + builtin_tool_name: "image_generation", same pattern web_search already uses. 'user' forwarding to MCP is a separate feature request orthogonal to this bug. Replaced by #1387 (ground-truth integration test against live OpenAI) which will surface any genuine protocol drift before we touch the MCP architecture.

PR #1390, opened 2026-04-25
test(e2e): MCP-builtin coverage for web_search_preview + code_interpreter

slin1237:
## Summary

Extends the R6.5 image_generation MCP-builtin e2e pattern (#1365) to two more hosted tools: `web_search_preview` and `code_interpreter`. The same in-process `MockMcpServer` from `e2e_test/infra/mock_mcp_server.py` is registered with `builtin_type: <tool>` + `builtin_tool_name: <tool>`, the gateway is launched with the corresponding `--mcp-config-path`, and the test asserts the response item shape matches the production OpenAI Responses API capture under `/tmp/openai_ground_truth/results/`.

The expected wire shapes are taken verbatim from the ground-truth captures referenced in `/tmp/openai_ground_truth/results/SUMMARY.md`:

| Tool | Output item fields | Streaming envelope |
|---|---|---|
| `web_search_call` | `id` (`ws_*`), `action {type: "search", query, queries, sources}`, `status`, `type` | `in_progress` -> `searching` -> `completed` |
| `code_interpreter_call` | `id` (`ci_*`), `code`, `container_id`, `outputs`, `status`, `type` | `in_progress` -> `interpreting` -> `completed` |

`file_search` is deferred to a follow-up — the OpenAI built-in maps to a vector store on the upstream side, and the gateway's MCP-builtin lane needs that provisioning hook before a mock can be exercised end-to-end.

### `e2e_test/infra/mock_mcp_server.py`
- Adds deterministic `web_search` and `code_interpreter` tool implementations with module-level `WEB_SEARCH_RESULTS` / `CODE_INTERPRETER_CODE` / `CODE_INTERPRETER_OUTPUTS` constants. Each tool records its arguments via the existing `_record_call` so tests can assert what the gateway dispatched.
- Extends the existing `image_generation` tool signature with an optional `user` kwarg so a forward-compatible gateway change that propagates the request-level `user` field into MCP dispatch args will not be rejected by FastMCP schema validation.
- Drops the stale TODO comment block; replaces it with a single note explaining that `file_search` is intentionally deferred.

### `e2e_test/infra/__init__.py` and `e2e_test/infra/mock_mcp.py`
- Re-export the three new constants so tests can import them via `from infra import ...`.

### `e2e_test/responses/conftest.py`
- Adds `_web_search_mcp_config` and `_code_interpreter_mcp_config` builders that produce per-builtin YAML config dicts.
- Adds session-scoped `mock_mcp_config_file_web_search` and `mock_mcp_config_file_code_interpreter` fixtures, plus a shared `_write_mcp_config_yaml` helper.
- Adds class-scoped `gateway_with_mock_mcp_web_search_cloud`, `gateway_with_mock_mcp_code_interpreter_cloud`, and the matching `_grpc_sglang` variants. Cloud lanes share a `_start_cloud_gateway_with_mcp` helper that skips when `OPENAI_API_KEY` is absent.

### `e2e_test/responses/test_web_search_preview.py` (new)
- `_WebSearchPreviewAssertions` mix-in with non-streaming + streaming test bodies. Forces `tool_choice {"type": "web_search_preview"}`. Asserts the documented `web_search_call` shape (`id` `ws_*`, `status` `completed`, `action.type` `search`, `action.query` non-empty, no `arguments` / `output` fields) plus a non-vacuity guard on the mock's `call_log`.
- `TestWebSearchPreviewCloud` (vendor `openai`, `gpu(0)`).
- `TestWebSearchPreviewGrpcSglang` (engine `sglang`, `gpu(2)`, `openai/gpt-oss-20b`). The harmony parser routes `browser`-recipient messages into reasoning rather than function calls today, so the inherited tests are overridden with `pytest.skip` documenting the gap (see `model_gateway/src/routers/grpc/harmony/parser.rs`).
- `TestWebSearchNonPreview` documents that `{"type": "web_search"}` (non-preview) is not routed through MCP-builtin in `extract_builtin_types` (see `model_gateway/src/routers/common/mcp_utils.rs`) and skips with a clear reason.

### `e2e_test/responses/test_code_interpreter.py` (new)
- `_CodeInterpreterAssertions` mix-in with non-streaming + streaming tests. Asserts unconditional invariants (`id` `ci_*`, `status` `completed`, `container_id` non-empty string, no `arguments` field). Per-field equality on `container_id` / `code` / `outputs` is gated on a `pytest.skip` when the transformer surfaces the degraded shape it produces today: `to_code_interpreter_call` only reads `result.as_object()` and the FastMCP-wrapped MCP content is an array of text blocks, so the gateway emits `container_id="unknown"`, `code=None`, `outputs=None`. Once the transformer walks text blocks (matching `to_image_generation_call`), the skip becomes unreachable.
- `TestCodeInterpreterCloud` and `TestCodeInterpreterGrpcSglang` follow the same pattern as web_search_preview, with the same harmony skip rationale (`python` / `container` recipient channels).

### `e2e_test/responses/test_image_generation.py`
- Adds `test_image_generation_user_forwarded_to_mcp` on the shared mix-in: posts a request with `user="..."` and asserts the value reaches the mock under `last_call_args["arguments"]["user"]`. The current gateway maps `user` -> `safety_identifier` on the upstream payload (see `model_gateway/src/routers/openai/responses/utils.rs`) but does not propagate it onto MCP dispatch args; the test skips with a clear reason today and starts asserting once that propagation lands. Coordinates with the parallel user-forwarding work — once that PR merges, the skip path becomes unreachable.

## Why

The R6.5 PR (#1365) validated the MCP-as-builtin pattern for `image_generation` end-to-end against both the OpenAI cloud lane and the sglang harmony lane. The other hosted-tool builtins routable through MCP had no equivalent e2e coverage, so any regression on dispatch / transformer shape would only surface in manual smoke testing.

## How

Mirrors the R6.5 fixture and test-class structure so reviewers see the same shape. Each cloud test forces `tool_choice` to the matching hosted tool so the model deterministically invokes it, then asserts spec-required invariants on the resulting output item plus a non-vacuity guard on the mock's `call_log`. Streaming variants assert the documented event sequence per `crates/protocols/src/event_types.rs` and the ground-truth captures.

Where a current-state gap prevents an assertion from holding today, the test skips with a clear explanation rather than xfailing or asserting weaker invariants:

| Gap | Where | Trigger |
|---|---|---|
| `web_search` (non-preview) is not in `extract_builtin_types` | `model_gateway/src/routers/common/mcp_utils.rs` | `TestWebSearchNonPreview` |
| Harmony parser routes `browser` / `python` / `container` recipients to reasoning, not function calls | `model_gateway/src/routers/grpc/harmony/parser.rs` | grpc-sglang classes |
| `to_code_interpreter_call` only reads `result.as_object()`, not text blocks | `crates/mcp/src/transform/transformer.rs` | per-field equality check inside `_assert_code_interpreter_call_item` |
| Request-level `user` not propagated to MCP dispatch args | `model_gateway/src/routers/openai/responses/utils.rs` | `test_image_generation_user_forwarded_to_mcp` |

No router or protocol code is modified — this PR is tests + fixtures + mock-server extensions only.

## Test plan

- [x] `ruff check e2e_test/` and `ruff format --check e2e_test/` pass locally.
- [x] All new modules import cleanly (`python -c 'import responses.test_web_search_preview, responses.test_code_interpreter'`).
- [x] Manual `MockMcpServer.start()` listing verifies the three tools register with the expected input schemas (`image_generation`: `prompt/size/quality/moderation/output_format/user`; `web_search`: `query/count`; `code_interpreter`: `code`).
- [x] Manual tool calls against the mock verify the deterministic payloads serialize correctly through FastMCP's `content` + `structuredContent` envelope.
- [ ] Cloud lanes (`TestWebSearchPreviewCloud`, `TestCodeInterpreterCloud`) will be exercised in CI with `OPENAI_API_KEY` set; absent the key they skip cleanly with a clear message rather than failing. Run locally with `OPENAI_API_KEY=$(cat /tmp/openai_ground_truth/key.txt) pytest e2e_test/responses/test_web_search_preview.py e2e_test/responses/test_code_interpreter.py -v` to verify against real cloud.
- [ ] gpu-2 sglang lane CI (`e2e-2gpu-responses`) skips both grpc-sglang tests cleanly with a clear harmony-routing reason.

<details>
<summary>Checklist</summary>

- [x] Tests added
- [x] No source code modified
- [x] Skips are explicit and explained
- [x] Mock server reused (no second mock written)

</details>

## Summary by CodeRabbit

* **Tests**
  * Added E2E tests for web search preview and code interpreter (streaming and non‑streaming) and a test verifying image generation forwards the user field; covers cloud and local gateway variants with appropriate fixtures and targeted skips.
* **Chores**
  * Enhanced test infrastructure and mock tooling with deterministic mock payloads, new hosted/local gateway fixtures, and reusable MCP config/launcher helpers for hosted and local test runs.

### DIFF
--- a/e2e_test/infra/__init__.py
+++ b/e2e_test/infra/__init__.py
@@ -40,7 +40,14 @@ from .constants import (  # Enums; Convenience sets; Fixture parameters; Default
 from .gateway import Gateway, WorkerInfo, launch_cloud_gateway
 from .gpu_monitor import GPUMonitor
 from .gpu_monitor import should_monitor as should_monitor_gpu
-from .mock_mcp import IMAGE_GENERATION_PNG_BASE64, MockMcpServer, mock_mcp_server
+from .mock_mcp import (
+    CODE_INTERPRETER_CODE,
+    CODE_INTERPRETER_OUTPUTS,
+    IMAGE_GENERATION_PNG_BASE64,
+    WEB_SEARCH_RESULTS,
+    MockMcpServer,
+    mock_mcp_server,
+)
 from .model_specs import (  # Default model paths; Model groups
     CHAT_MODELS,
     DEFAULT_EMBEDDING_MODEL_PATH,
@@ -136,6 +143,9 @@ __all__ = [
     "MockMcpServer",
     "mock_mcp_server",
     "IMAGE_GENERATION_PNG_BASE64",
+    "WEB_SEARCH_RESULTS",
+    "CODE_INTERPRETER_CODE",
+    "CODE_INTERPRETER_OUTPUTS",
     # Default model paths
     "DEFAULT_MODEL_PATH",
     "DEFAULT_SMALL_MODEL_PATH",
--- a/e2e_test/infra/mock_mcp.py
+++ b/e2e_test/infra/mock_mcp.py
@@ -32,7 +32,13 @@ from collections.abc import Iterator
 
 import pytest
 
-from .mock_mcp_server import IMAGE_GENERATION_PNG_BASE64, MockMcpServer
+from .mock_mcp_server import (
+    CODE_INTERPRETER_CODE,
+    CODE_INTERPRETER_OUTPUTS,
+    IMAGE_GENERATION_PNG_BASE64,
+    WEB_SEARCH_RESULTS,
+    MockMcpServer,
+)
 
 
 @pytest.fixture(scope="session")
@@ -47,7 +53,10 @@ def mock_mcp_server() -> Iterator[MockMcpServer]:
 
 
 __all__ = [
+    "CODE_INTERPRETER_CODE",
+    "CODE_INTERPRETER_OUTPUTS",
     "IMAGE_GENERATION_PNG_BASE64",
+    "WEB_SEARCH_RESULTS",
     "MockMcpServer",
     "mock_mcp_server",
 ]
--- a/e2e_test/infra/mock_mcp_server.py
+++ b/e2e_test/infra/mock_mcp_server.py
@@ -77,6 +77,36 @@ IMAGE_GENERATION_PNG_BASE64 = (
 )
 
 
+# Deterministic ``web_search`` results. Shape mirrors the embedded
+# ``openai_response`` payload that ``ResponseTransformer::extract_web_sources``
+# walks through to populate the ``web_search_call.action.sources`` field.
+# Hard-coded so tests can byte-compare URLs / titles without depending on a
+# live search backend.
+WEB_SEARCH_RESULTS = [
+    {
+        "title": "Mock weather report",
+        "url": "https://example.com/mock-weather",
+        "snippet": "Deterministic mock snippet for weather.",
+    },
+    {
+        "title": "Mock secondary result",
+        "url": "https://example.com/mock-secondary",
+        "snippet": "Deterministic mock snippet for follow-up.",
+    },
+]
+
+
+# Deterministic ``code_interpreter`` payload. ``code`` is a tiny self-contained
+# expression so a non-zero string survives the gateway's transformer (it
+# unconditionally forwards ``code`` from the MCP result onto the
+# ``code_interpreter_call.code`` field). ``outputs`` uses the typed
+# ``{type: "logs", logs: ...}`` shape that ``extract_code_outputs`` recognises.
+CODE_INTERPRETER_CODE = "print('hi')"
+CODE_INTERPRETER_OUTPUTS = [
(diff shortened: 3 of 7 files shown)


slin1237:
Closing — scope was wrong. The intent was a small extension to e2e_test/responses/test_image_generation.py adding negative coverage for the regular-MCP-server-no-builtin_type path. Web_search_preview / code_interpreter / file_search are already covered by e2e_test/responses/test_builtin_tools.py and don't need new test files. Replacing with a small targeted PR.

PR #1460, opened 2026-05-07
fix(protocols): add continuous_usage_stats to StreamOptions

Abhishek8108:
## Motivation

`continuous_usage_stats` is missing from the `StreamOptions` struct in `crates/protocols/src/common.rs`. When a client sends `stream_options: { "include_usage": true, "continuous_usage_stats": true }`, serde silently drops `continuous_usage_stats` on deserialisation because the field is not declared on the struct. The re-serialised payload forwarded downstream therefore never contains it — the backend falls back to emitting a usage chunk only at the end of the stream, ignoring the caller's intent.

Reported upstream: sgl-project/sglang#24608

## Modifications

- Added `continuous_usage_stats: Option<bool>` to `StreamOptions` in `crates/protocols/src/common.rs`, following the same `#[serde(skip_serializing_if = "Option::is_none")]` pattern as the existing fields.
- Added `test_stream_options_continuous_usage_stats_valid` in `model_gateway/tests/spec/chat_completion.rs` to assert the field is accepted by validation and round-trips correctly through serde.

## Non-breaking

All existing `StreamOptions { ... }` construction sites use `..StreamOptions::default()`, so adding a new `Option<bool>` field (which defaults to `None` and is not serialised when `None`) does not affect any existing code or wire format.

## Accuracy Tests

Not applicable — no model output or inference logic is affected.

## Speed Tests and Profiling

Not applicable.

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations). *(no new public API surface; existing `StreamOptions` doc comment updated inline)*
- [ ] Provide accuracy and speed benchmark results. *(not applicable)*
- [x] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Summary by CodeRabbit

* **New Features**
  * Added optional continuous usage statistics for streaming responses. Usage metrics are now emitted with each streamed delta in real-time instead of only in the final response, providing improved visibility into token consumption during active streaming operations.

### DIFF
--- a/crates/protocols/src/common.rs
+++ b/crates/protocols/src/common.rs
@@ -240,6 +240,11 @@ pub struct StreamOptions {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub include_usage: Option<bool>,
 
+    /// Chat Completions / Completions: emit a usage chunk with every streamed
+    /// delta instead of only in the final chunk.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub continuous_usage_stats: Option<bool>,
+
     /// Responses API: add random chars on `obfuscation` field of delta events
     /// to normalize payload sizes. Defaults to `true` upstream when absent.
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/model_gateway/src/routers/grpc/regular/responses/conversions.rs
+++ b/model_gateway/src/routers/grpc/regular/responses/conversions.rs
@@ -499,6 +499,7 @@ mod tests {
             stream_options: Some(StreamOptions {
                 include_usage: None,
                 include_obfuscation: Some(false),
+                ..StreamOptions::default()
             }),
             ..Default::default()
         };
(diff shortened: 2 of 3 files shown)


Abhishek8108:
The P2 note about streaming paths not acting on `continuous_usage_stats` is correct — `grpc/regular/streaming.rs` and `grpc/harmony/streaming.rs` currently only gate on `include_usage`. This PR's scope is just wiring the field through deserialization so it stops being silently dropped. Actually emitting per-delta usage chunks based on the flag is a follow-on change that touches the streaming logic and warrants its own PR.

PR #1515, opened 2026-05-21
feat(grpc): add TokenSpeed multimodal (VLM) input support

chenht2022:
### Problem

TokenSpeed is wired into the SMG gateway for text generation, but its gRPC path carries no multimodal inputs: an image request routed to a TokenSpeed worker is rejected at the router ("TokenSpeed backend does not support multimodal inputs").

### Solution

Extend the TokenSpeed gRPC path to carry the gateway's already-preprocessed multimodal tensors. The gateway performs per-model image preprocessing in `crates/multimodal`; the servicer reconstructs the tensors and sets `precomputed_multimodal_inputs` so the engine consumes them directly and skips its own preprocessing.

## Changes

- `crates/grpc_client/proto/tokenspeed_scheduler.proto`: add `MultimodalInputs` / `TensorData` / `PlaceholderRange`, `GenerateRequest.mm_inputs`, and `GetModelInfoResponse.supports_vision`.
- `crates/grpc_client/src/tokenspeed_scheduler.rs`, `model_gateway/src/routers/grpc/{client,multimodal,proto_wrapper}.rs`: assemble and forward TokenSpeed multimodal requests via a new `MultimodalData::TokenSpeed` variant.
- `grpc_servicer/smg_grpc_servicer/tokenspeed/{servicer,server}.py`: decode the precomputed multimodal payload into the engine's inputs; `TOKENSPEED_GRPC_MAX_MESSAGE_BYTES` lifts the gRPC ceiling for large VLM `pixel_values`.
- `scripts/ci_install_tokenspeed.sh`: simplify install (3 editable `uv pip install -e` lines) and bump `TOKENSPEED_REF` past lightseekorg/tokenspeed#247.

## Test Plan

- `cargo +nightly fmt --all` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` pass; Python `ruff format` / `ruff check` are clean. (`cargo test` runs in CI.)
- The Python proto stubs regenerate cleanly and the servicer imports and round-trips the new `MultimodalInputs` message.


grpc_servicer/smg_grpc_servicer/tokenspeed/servicer.py
```diff
@@ -687,6 +700,81 @@ def _sampling_params_from_proto(
...
+        model_dtype: torch.dtype | None = None,
+    ):
+        """Reconstruct the engine's ``MultimodalInputs`` from the precomputed proto.
+
+        The gateway already preprocessed, so the engine skips its own preprocessing
+        (``precomputed_multimodal_inputs`` is set); this just boxes the tensors and
+        placeholder offsets into the engine's data class.
+        """
+        from tokenspeed.runtime.multimodal.inputs import (
+            Modality,
```
CatherineSue: Please put imports at the top of the file

grpc_servicer/smg_grpc_servicer/tokenspeed/servicer.py
```diff
@@ -29,8 +29,9 @@
 from smg_grpc_servicer.tokenspeed.health_servicer import TokenSpeedHealthServicer
 
 if TYPE_CHECKING:
-    # Type-only imports — not resolved at module load so the servicer is
-    # importable in test environments that stub AsyncLLM / ServerArgs.
+    # Type-only imports, not resolved at module load, so the servicer stays
+    # importable where the engine / torch are stubbed.
+    import torch
```
CatherineSue: Same here

grpc_servicer/smg_grpc_servicer/tokenspeed/servicer.py
```diff
@@ -687,6 +700,81 @@ def _sampling_params_from_proto(
...
+        tensor_data: tokenspeed_scheduler_pb2.TensorData,
+        cast_to: torch.dtype | None = None,
+    ):
+        """Reconstruct a torch.Tensor from a proto TensorData.
+
+        numpy/torch are imported lazily so the module loads where the engine is
+        stubbed. Floats are cast to ``cast_to``, fused into the decode; the buffer
+        is copied so it never aliases the transient proto bytes.
+        """
+        import numpy as np
```
CatherineSue: Same here.

grpc_servicer/smg_grpc_servicer/tokenspeed/server.py
```diff
@@ -33,8 +33,9 @@ async def serve_grpc(server_args: ServerArgs) -> None:
     server = grpc.aio.server(
         futures.ThreadPoolExecutor(max_workers=10),
         options=[
-            ("grpc.max_send_message_length", 1024 * 1024 * 256),
-            ("grpc.max_receive_message_length", 1024 * 1024 * 256),
+            # VLM pixel_values can be large; raise to gRPC's max (2 GiB - 1).
+            ("grpc.max_send_message_length", (2 << 30) - 1),
+            ("grpc.max_receive_message_length", (2 << 30) - 1),
```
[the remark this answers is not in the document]
CatherineSue: This comment is valid.

zhyncs:
@codex review

grpc_servicer/smg_grpc_servicer/tokenspeed/servicer.py
```diff
@@ -21,16 +21,23 @@
...
+import torch
 from google.protobuf.struct_pb2 import Struct
 from google.protobuf.timestamp_pb2 import Timestamp
 from smg_grpc_proto import tokenspeed_scheduler_pb2_grpc
 from smg_grpc_proto.generated import tokenspeed_scheduler_pb2
+from tokenspeed.runtime.multimodal.inputs import (
+    Modality,
+    MultimodalDataItem,
+    MultimodalInputs,
+)
```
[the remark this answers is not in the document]
zhyncs: @CatherineSue any thoughts?

crates/grpc_client/proto/tokenspeed_scheduler.proto
```diff
@@ -85,13 +87,46 @@ message GenerateRequest {
...
+// Where a multimodal item's tokens sit inside input_ids.
+message PlaceholderRange {
+  uint32 offset = 1;
+  uint32 length = 2;
+}
+
+// Multimodal inputs for vision/audio models. Tensors are produced by the
+// gateway's per-model preprocessor (crates/multimodal); the servicer
+// only reconstructs them and hands them to the engine — no preprocess.
+message MultimodalInputs {
```
yechank-nvidia, line 116: Current structure assumes 1) the primary tensor is image-centric 2) the whole request is single modality. I think we need to expand mixed-modality.

```suggestion
message MultimodalInputs {
    repeated MultimodalItem items = 1;
}

message MultimodalItem {
    Modality modality = 1;          // IMAGE | AUDIO | VIDEO | ...
    bytes content_hash = 2;          // gateway-computed
    map<string, TensorData> tensors = 3;  // HF processor's key
    PlaceholderRange placeholder = 4; 
}

enum Modality {
    MODALITY_UNSPECIFIED = 0;
    IMAGE = 1;
    AUDIO = 2;
    VIDEO = 3;
}
```

crates/grpc_client/proto/tokenspeed_scheduler.proto
```diff
@@ -85,13 +87,46 @@ message GenerateRequest {
...
+  MultimodalInputs mm_inputs = 9;
 }
 
 message TokenizedInput {
   repeated uint32 input_ids = 1;
   string original_text = 2; // cosmetic, for worker logs
 }
 
+// A typed tensor: raw little-endian bytes + shape + dtype.
+message TensorData {
```
yechank-nvidia, line 101: For a 100 MB bf16 `pixel_values` tensor, the request path involves 2-4 CPU memcpys (producer-side serialize, two kernel-side TCP copies, possibly a worker-side proto decode copy depending on the protobuf backend) plus `pin_memory()` and H2D DMA.

For image-sized payloads this is largely hidden by downstream compute (vision tower + LLM prefill). The concerns are scaling:

  1. **Tensor size grows non-linearly** - 1-min video is 100-500 MB, so per-request memcpy + wire time crosses into the tens-to-hundreds-of-ms range and stops being hidden by compute.
  2. **Aggregate bandwidth at concurrency** - large tensors x N concurrent requests contend with other workloads (KV cache, etc.) on shared memory bandwidth.
  3. **2 GiB gRPC ceiling is a hard wall** - longer / higher-res video will hit it again; raising the limit each time isn't a sustainable answer.
  4. **CPU-bound deserialize at scale** - protobuf decode + numpy/torch conversion becomes the bottleneck before model compute under high QPS due to the GIL.

None of these block the image use case today. The point is that the wire schema locks the design into a single transport, so when video lands we'll either raise the gRPC limit again or break the wire format. Both are avoidable by reserving the `oneof payload` shape now.

```suggestion
message TensorData {
    repeated uint32 shape = 1;
    string dtype = 2;
    oneof payload {
      bytes inline = 3;             // current
      ShmHandle shm = 4;            // same host
      CudaIpcHandle cuda_ipc = 5;   // same GPU node
      NixlDescriptor nixl = 6;      // RDMA
    }
  }
```

crates/grpc_client/proto/tokenspeed_scheduler.proto
```diff
@@ -191,6 +226,11 @@ message GetModelInfoResponse {
   // ``{"temperature": 0.6, "top_p": 0.9}``). Empty = no overrides. Surfaced
   // to the router via the ``default_sampling_params_json`` worker label.
   string default_sampling_params_json = 13;
+
+  // True when the model can consume image/video inputs. Drives the
+  // router's mm_inputs handling (router rejects mm requests for
+  // non-vision workers before tokenization).
+  bool supports_vision = 14;
```
yechank-nvidia, line 233:
```suggestion
  bool support_multimodal = 14;
```

PR #1525, opened 2026-05-24
fix(observability): include addr in metrics bind failure

zhyncs:
`tokio::net::TcpListener::bind` errors don't carry the address that was attempted, so a port-conflict / EACCES panic at `metrics_server.rs:59` dropped the operator's configured host:port on the floor. Surface `addr` in the failure message so a bad startup points at the conflicting endpoint.

The bind is pulled into a small `bind_metrics_listener` helper returning `Result<TcpListener, String>` so the panic site stays minimal and the message construction is unit-testable. The clippy guard on `start_metrics_server` flips from `expect_used` to `panic`.

## Test plan

- New unit test `bind_metrics_listener_error_includes_addr`: pre-binds a loopback port and asserts the helper's error carries both `addr.to_string()` and the `failed to bind metrics server` prefix.
- Verified locally: `cargo test bind_metrics_listener_error_includes_addr` ✅, `cargo clippy -- -D warnings` ✅, `cargo +nightly fmt -- --check` ✅.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined metrics server startup to centralize listener binding and error handling.

* **Bug Fixes**
  * Improved bind-failure messages to include the target address and clearer context for startup errors.

* **Tests**
  * Added a unit test to validate bind-failure messaging includes address and bind-failure context.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightseekorg/smg/pull/1525?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1528, opened 2026-05-24
fix(observability): use expect for metrics bind failure

zhyncs:
## Summary
- replace the metrics listener bind failure panic path with Result::expect
- keep the existing #[expect(clippy::expect_used)] lint expectation accurate

## Testing
- git diff --check
- not run: cargo fmt --check / cargo test -p smg bind_metrics_listener_error_includes_addr (cargo is not installed in this environment)

Refs https://github.com/lightseekorg/smg/pull/1525#discussion_r3293978154

### DIFF
--- a/model_gateway/src/observability/metrics_server.rs
+++ b/model_gateway/src/observability/metrics_server.rs
@@ -62,7 +62,7 @@ pub async fn start_metrics_server(
 
     let listener = bind_metrics_listener(addr)
         .await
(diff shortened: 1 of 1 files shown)


zhyncs:
```
  running 1 test
  bind metrics listener error: failed to bind metrics server on 127.0.0.1:41221: Address already in use (os error 98)
  test observability::metrics_server::tests::bind_metrics_listener_error_includes_addr ... ok
```

PR #1557, opened 2026-05-27
fix(scheduler): recompute per-class queue limit against live capacity

CatherineSue:
### Problem

`queue_size_per_slot` is a per-class multiplier that's supposed to scale the queue depth limit with the live backend capacity:

```
effective_queue_limit = max(class.queue_size, ceil(class.queue_size_per_slot * capacity))
```

The intent is "if my fleet auto-scales 10×, the queue limit auto-scales 10× too, no retune." But the scheduler called `queue_for(settings, class, capacity)` *once* in `PriorityScheduler::new` and baked the result into `FifoClassQueue::max` at construction. When `apply_new_capacity` updated the slot pool's ceiling via the `WorkerCapacity::watch()` channel, the per-class queue limit stayed frozen at the startup value — the multiplier did nothing after t=0.

Caught during M2b review/discussion; deferred from #1550 to keep the PR focused.

### Solution

Move the depth check from the queue's internal state to a per-call parameter, and recompute the limit against live capacity on every admission.

## Changes

One commit:

- **`ClassQueue::try_enqueue` takes `max: usize`** per call. `FifoClassQueue` drops its `max` field and its `new(max)` constructor, replaced with `new()`.
- **`ClassRuntimeConfig` gains `queue_size` + `queue_size_per_slot`** plus a `queue_limit_for(capacity) -> usize` helper that performs the `max(queue_size, ceil(queue_size_per_slot * capacity))` recompute.
- **`PriorityScheduler::admit`** reads `self.slot_pool.capacity()` and the per-class runtime config on every enqueue and passes the result to `try_enqueue`. Removes the now-obsolete `queue_for` helper.
- **`PriorityScheduler::new`** constructs `FifoClassQueue::new()` (no baked-in limit) for each class.

## Test Plan

1 new unit test for the queue (`test_try_enqueue_max_can_grow_between_calls`), 3 new unit tests for `ClassRuntimeConfig::queue_limit_for` (floor wins / multiplier wins / multiplier disabled), and 1 end-to-end integration test that exercises the autoscaling path:

```
build scheduler with capacity=10, queue_size=4, queue_size_per_slot=1.0
  → startup limit = max(4, ceil(1.0 * 10)) = 10
fill queue to 10 (each enqueue under startup limit succeeds)
set_capacity(20) via the slot pool
  → live limit = max(4, ceil(1.0 * 20)) = 20
try_enqueue(eleventh-waiter, live_limit=20) → succeeds
```

The previous code would have rejected the 11th waiter with `QueueFull` because `FifoClassQueue::max` was still 10.

```bash
$ cargo test --package smg --lib middleware::scheduler
... 91 passed; 0 failed
```

```bash
$ cargo clippy --package smg --all-targets --all-features -- -D warnings
... no warnings
```

<details>
<summary>Checklist</summary>

- [x] `cargo +nightly fmt` passes (enforced by pre-commit hook)
- [x] (Optional) Documentation updated — design doc §3 already specified live recompute; code now matches

</details>

## Summary by CodeRabbit

* **Improvements**
  * Queue scheduling now dynamically adjusts admission limits based on current backend capacity, allowing additional requests to be accepted when capacity increases
  * Scheduler behavior adapts to live capacity changes instead of using fixed startup limits

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightseekorg/smg/pull/1557?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/model_gateway/src/middleware/scheduler/engine.rs
+++ b/model_gateway/src/middleware/scheduler/engine.rs
@@ -72,6 +72,12 @@ pub struct PriorityScheduler {
     /// exits cleanly.
     pub(super) release_notify: Arc<Notify>,
     class_config: [ClassRuntimeConfig; 4],
+    /// `(queue_size, queue_size_per_slot)` for each class, indexed by
+    /// `class as usize`. Read on every admit to recompute the queue
+    /// depth limit against the *current* slot pool capacity so the
+    /// `queue_size_per_slot` multiplier actually tracks live capacity
+    /// changes (it'd be frozen at startup otherwise).
+    class_queue_limit_inputs: [(u32, f32); 4],
 }
 
 impl Drop for PriorityScheduler {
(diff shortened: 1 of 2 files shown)


CatherineSue:
Closing in favor of a simpler approach — removing queue_size_per_slot entirely instead of trying to make the multiplier live. The feature was never wired to track capacity changes after startup, and the fixes to make it actually live were heavier than the autoscaling benefit justifies. Follow-up PR removes the field.

PR #1565, opened 2026-05-27
fix(cache-aware): gate hash_index hot-path writes behind explicit flag

ekzhang:
Add an AtomicBool populate_hash_index field to CacheAwarePolicy (default false) and a set_populate_hash_index() setter. Gate the four hot-path inserts on the flag; mesh wiring is expected to call set_populate_hash_index(true) when attaching the policy to a TreeSyncAdapter. The cold-start apply_repair_page writes are not gated since they only run when mesh is actually applying remote pages.

Note: there is currently no production code in server.rs that wires the v2 mesh adapters to the policy (per the existing comment about v1->v2 migration landing in a follow-up). Until that wiring lands and opts in, the index stays empty and the memory leak is closed.

### Problem

The hash_index (DashMap<model_id, PerModelHashIndex>) is written from four select_worker_* request-hot-path sites on every request, but its only readers are mesh-only methods on the TreeHandle trait (apply_known_remote_insert reads, apply_repair_page also writes during cold-start sync). When no mesh adapter is attached to the policy these entries accumulate with no consumer, resulting in OOM crashes every ~15 minutes in production.

### Solution

Gate writing on `hash_index` to only when mesh mode is enabled. This way, the disabled path (default) doesn't run into this issue.

## Changes

We needed to add an `AtomicBool` field to make this work due to recent refactors that removed `mesh_sync`.

## Test Plan

We've been running this in production for a while.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added a runtime toggle to enable or disable hot-path population of per-model cache hash indexes, letting operators control when request paths write index entries.

* **Bug Fixes**
  * Gates cache-index writes to prevent unintended updates during inactive sync or fallback paths, reducing overhead and improving stability.

* **Tests**
  * Updated tests to exercise the new toggle and validate hot-path index behavior.

### DIFF
--- a/model_gateway/src/policies/cache_aware.rs
+++ b/model_gateway/src/policies/cache_aware.rs
@@ -443,7 +443,11 @@ impl CacheAwarePolicy {
                 // returns the prior shared prefix (~50-200 chars).
                 let matched_prefix = if self.should_populate_hash_index() {
                     let result = tree.match_prefix_with_counts(text);
-                    Some(text.chars().take(result.matched_char_count).collect::<String>())
+                    Some(
+                        text.chars()
+                            .take(result.matched_char_count)
+                            .collect::<String>(),
+                    )
                 } else {
                     None
                 };
--- a/model_gateway/src/policies/cache_aware.rs
+++ b/model_gateway/src/policies/cache_aware.rs
@@ -41,7 +41,10 @@
     block_size:              Backend KV cache block size for event-driven routing
 */
 
-use std::sync::Arc;
+use std::sync::{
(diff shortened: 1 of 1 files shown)


ekzhang:
Hey @slin1237 , could I get a review on this PR? I think it's pretty important as without this PR the gateway leaks memory at a very high rate, over 1 GB/s increase in memory resident set size.

I fixed the rustfmt failure (sorry about that) and configured my editor so that it stops being an issue

PR #1585, opened 2026-06-01
test(scheduler): integration tests for rejection, clamp, preemption, starvation

slin1237:
### Problem
The merged priority scheduler had only the 4 basic integration tests from M6. Rejection codes, tenant clamping, end-to-end preemption, slot release, and starvation promotion were untested.

### Solution
Add 6 deterministic HTTP-level integration tests (mock-worker harness) covering those paths, plus a controllable-hold mock-worker helper so timing-sensitive scenarios are deterministic (no racing sleeps).

## Changes
- `tests/scheduler_admission_test.rs` (new): 429 queue-full, 408 queue-timeout, preemption → **503 + `X-SMG-Preempted`**, tenant clamp (header can't self-promote past the tenant cap), slot-release-after-drain, starvation promotion.
- `tests/common/mock_worker.rs`: additive `HoldGate` + per-port `SchedulerControls` (opt-in; `MockWorkerConfig` unchanged — absent ⇒ prior behavior).

## Test Plan
- `cargo test --test scheduler_admission_test` → **6/6 green, run twice** (non-flaky).
- Full integration suite compiles with the harness change; an existing merged mock-worker test (`inflight_tracker_test`, 12/12) stays green — the change is additive/opt-in.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (test target)

</details>

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive HTTP-level integration tests for the priority-admission scheduler covering queue-full rejections, queue timeouts, priority-based preemption, tenant class enforcement, and starvation-driven promotion.
  * Enhanced test infrastructure with deterministic, per-worker scheduler controls (including controllable request gating and capacity overrides) to reliably simulate and validate scheduler behaviors.

### DIFF
--- a/model_gateway/tests/common/mock_worker.rs
+++ b/model_gateway/tests/common/mock_worker.rs
@@ -4,7 +4,10 @@
 use std::{
     collections::{HashMap, HashSet},
     convert::Infallible,
-    sync::{Arc, Mutex, OnceLock},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, Mutex, OnceLock,
+    },
     time::{SystemTime, UNIX_EPOCH},
 };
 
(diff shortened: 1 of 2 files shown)


slin1237:
Addressed review feedback:
- **(coderabbit, valid)** `SCHEDULER_CONTROLS` is now pruned on `MockWorker::Drop` (the bound port is cached so the sync `Drop` can remove it), so a recycled port can't inherit stale `max_running_requests` / `hold_gate`.
- **(gemini, respectfully skipping)** dynamic gateway port / removing `#[serial]`: these tests exercise the gateway **in-process via `tower::oneshot`** — `create_app()` returns an `axum::Router` and `AppTestContext` never binds `config.port` (the default helper has used a hardcoded `3002` suite-wide). So the gateway port is inert config and can't cause address-in-use, and `#[serial]` is still required because the tests share a process-global metrics recorder and the `SCHEDULER_CONTROLS` table. Changing it would add an unused `pick_unused_port()` without removing any real risk.

Verified after the change: `scheduler_admission_test` 15/15 and `inflight_tracker_test` 12/12 green.

PR #1601, opened 2026-06-04
fix(responses): align function_call serde

shenoyvvarun:
## Summary
As per the openai specs.
- Preserve required empty Responses API arrays for `output_text.annotations` and `reasoning.summary`.
- Make Responses `function_call.id` optional on input/output items and omit absent `function_call_output.id`.
- Update gRPC Responses call sites to wrap generated IDs and fall back to `call_id` when replaying prior function calls.
- Add serde regressions for empty annotations, empty reasoning summaries, missing `function_call.id`, and missing `function_call_output.id`.

## Verification

- `cargo fmt`
- `cargo test -p openai-protocol --test responses`
- `cargo check -p smg`
- Launched SMG locally against the gpt-oss HTTP worker on `127.0.0.1:9091` and sent a Responses request with missing `function_call.id` and missing `function_call_output.id`; request returned HTTP 200 and response preserved `summary: []` and `annotations: []`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Function tool call identifiers are now optional and omitted when absent, improving payload fidelity.
  * Tool-call identifiers are consistently populated and preserved across history, conversions, and streaming to avoid mismatched or empty IDs.
  * Serialization now preserves empty arrays for reasoning summaries and text annotations to maintain wire-shape.

* **Tests**
  * Added roundtrip tests validating preservation of empty arrays and missing optional fields during serialize/deserialize.


key4ng:
It has grpc router related changes. @CatherineSue could you have a look when you have time

slin1237 (approved):

key4ng:
Hi @shenoyvvarun — thanks for tightening up the Responses serde, I validated all the wire-shape changes against the live OpenAI API and they line up perfectly. One adjacent thing I'd like your take on before this lands:

**Regular streaming accumulator never populates `call_id`** — `model_gateway/src/routers/grpc/regular/responses/streaming.rs:312-341`

The accumulator initializes `call_id: String::new()` and only ever accumulates into `id`:

```rust
self.tool_calls.push(ResponseOutputItem::FunctionToolCall {
    id: None,
    call_id: String::new(),   // <- never filled in
    ...
});
...
if let ResponseOutputItem::FunctionToolCall { id, name, arguments, .. } = &mut self.tool_calls[index] {
    if let Some(delta_id) = &delta.id {
        id.get_or_insert_with(String::new).push_str(delta_id);  // call_id left empty
    }
```

The non-streaming path (`conversions.rs:381`) sets `call_id: tool_call.id.clone()`, so the two paths now diverge. Because this PR switches `responses_to_chat` to read **`call_id`** instead of `id`, a streaming-accumulated function call that gets persisted (`finalize()` -> `persist_response_if_needed`) and later replayed as input history would yield an **empty `tool_call_id`** — whereas before this PR it used the populated `id`. So this is a pre-existing latent gap that the `id -> call_id` switch makes observable.

Would you be open to also populating `call_id` in the accumulator to match the non-streaming path? Something like binding `call_id` in the same `if let` and accumulating `delta_id` into it as well. Happy to be wrong if there's a downstream step that backfills `call_id` that I missed.

shenoyvvarun:
Addressed in latest push 0f589f6f: the regular streaming accumulator now populates call_id from streamed tool-call delta.id alongside id, and includes a regression test for that path. The branch is also rebased onto current main bb6373ba.

PR #1621, opened 2026-06-10
feat(cache-aware): KV-aware imbalance triggers (spread + overload) via load monitor

slin1237:
### Problem

Cache-aware routing decides imbalance — whether to abandon cache affinity and fall back to shortest-queue — solely from in-flight **request-count** spread (`w.load()`). Two failure modes:

1. **Long-context KV saturation** (e.g. Kimi K2.5): a few requests saturate one engine's KV without ever crossing the count-based `balance_abs_threshold` (64) / `balance_rel_threshold` (1.5) gate. Queue time piles onto the hot engine while others sit idle.
2. **Replica-count drift**: `w.load()` is *per-gateway* in-flight count. With N router replicas each sees ~1/N of the true load, so the effective imbalance threshold silently scales with replica count.

### Solution

Consume the backend **`token_usage`** signal (0.0–1.0 KV utilization, already polled by `WorkerMonitor` and used by `power_of_two`) via the shared `subscribe()` watch channel, wired into `CacheAwarePolicy` through `PolicyRegistry` like the existing `kv_event_monitor` injection. Imbalance detection becomes a **3-term OR** that adds KV-awareness *without replacing* the existing request-count logic.

## How cache-aware routing works — and when it falls back

For every request, `select_worker` runs in this order:

```
select_worker(request):
  1. is_imbalanced(pool)?  ──yes──▶  select_worker_min_load()   // abandon affinity → shortest queue
  2. else: score prefix overlap (radix tree) per worker
       best match_rate > cache_threshold ?  ──yes──▶  route to best-matching worker  // cache hit
                                             ──no───▶  select_worker_min_load()        // no useful cache → shortest queue
```

**Normal path (balanced pool).** The policy scores how much of the request's prefix each worker already holds in its KV cache — via one of three indexers (event-driven KV-events, approximate token tree, or approximate string tree) — and routes to the best-matching worker when the overlap clears `cache_threshold`. This is what maximizes cache hits.

**Fallback path.** The policy abandons cache affinity and routes to the **shortest queue** (`select_worker_min_load` → fewest in-flight requests) in two cases:
- the pool is **imbalanced** (the guard below), or
- **no worker** has a useful cached prefix (`match_rate ≤ cache_threshold`).

### The imbalance guard `is_imbalanced` is a 3-term OR

This PR's core change. The pool is imbalanced if **any** of these fire:

| # | Trigger | Condition | Knob (default) |
|---|---|---|---|
| a | **overload** | hottest engine KV `>` ceiling | `overload_token_usage_threshold` (`1.0` = off) |
| b | **KV spread** | (hottest − coldest) KV `>` threshold | `balance_token_usage_threshold` (`1.0` = off) |
| c | **count spread** | (max − min) count `> abs` **and** max `> rel·min` | `balance_abs_threshold` (64) / `balance_rel_threshold` (1.5) |

- (a) and (b) require a backend KV snapshot; both default to `1.0`, which disables them (utilization and spread are both ≤ 1.0). With defaults, **only (c) runs → behavior is byte-for-byte unchanged.**
- The three are **independent and OR'd**: the count check (c) is *always* evaluated and independently tunable, and the KV signal is *additive*, never a replacement. (This is @ekzhang's review suggestion — tune `--balance-abs-threshold` on its own, and recover "KV-only" behavior by setting it very high.)
- Why **spread**, not a bare max, for (b): `max_kv > t` is a *saturation* signal, not a *balance* one — it can't distinguish "one hot engine, idle neighbors" (rebalance) from "all engines uniformly hot" (keep affinity; spilling just destroys cache hits when the cluster is busiest). The overload ceiling (a) is kept separately as an explicit safety valve.

### Fallback target

When imbalanced, `select_worker_min_load` picks the **shortest queue by request count**. Making that spill toward the **coldest-KV** engine (so a KV-triggered imbalance is relieved on the KV axis) is a separate, coherent follow-up.

## Changes

- `policies/cache_aware.rs`: `is_imbalanced` → 3-term OR; `backend_token_usage_bounds` returns `(min, max)` KV over healthy workers; the count fold runs over `healthy_indices`. 5 unit tests pin the distinctions (uniform-high vs one-hot-rest-idle vs overload vs high-count/low-KV vs default-disabled).
- `policies/registry.rs`: inject the monitor's `subscribe()` receiver (mirrors `kv_event_monitor`).
- `config/*`, `main.rs`, `policies/{mod,factory}.rs`, `app_context.rs`: thread `balance_token_usage_threshold` (spread) + `overload_token_usage_threshold` (ceiling), both default `1.0`.
- `bindings/python` (`lib.rs` + `router_args.py`): expose `--balance-token-usage-threshold` and `--overload-token-usage-threshold`.

## Test Plan

Local gate (`maturin` unavailable, so the binding is verified with `cargo build`):

| Check | Result |
| --- | --- |
| `cargo +nightly fmt --all` | clean |
| `cargo clippy --all-targets --all-features -- -D warnings` | `Finished` — zero |
| `cargo test -p smg --lib cache_aware` | `32 passed` |
| `cargo test -p smg --test routing_tests` | `92 passed` |
| `cargo build -p smg-python` | `Finished` |

Rebased onto `main` @ `63b532a7` (incl. #1629 `least_load`) — no conflicts.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added two configurable token-usage thresholds for cache-aware routing (disabled by default).
  * Exposed the thresholds in the CLI and Python constructor with sensible defaults.
  * Runtime wiring enables cache-aware policies to consume per-worker load snapshots for token-usage-based imbalance detection.
  * Validation enforces valid threshold values and registry propagation ensures policies receive the load snapshot stream.


model_gateway/src/policies/cache_aware.rs
```diff
@@ -236,12 +250,67 @@ impl CacheAwarePolicy {
...
+    fn is_imbalanced(&self, workers: &[Arc<dyn Worker>], healthy_indices: &[usize]) -> bool {
+        // Token-usage mode: configured (< 1.0) AND a snapshot exists → use the
+        // KV signal alone, ignoring request counts. A missing snapshot falls
+        // through to the count check below (never left blind).
+        let kv_threshold = self.config.balance_token_usage_threshold;
+        if (0.0..1.0).contains(&kv_threshold) {
+            if let Some(max_usage) = self.max_backend_token_usage(workers, healthy_indices) {
+                return max_usage > f64::from(kv_threshold);
+            }
+        }
```
ekzhang: Could it be more flexible to check based on _either_ the token usage threshold, or the existing `--balance-abs-threshold`? Reason being that if it's implemented this way, then people can also set the balance abs threshold independently, and you can recover the "sole" token usage behavior by setting `--balance-abs-threshold 10000000`


```suggestion
        // Token-usage mode: configured (< 1.0) AND a snapshot exists → use the
        // KV signal additionally to check for imbalance.
        let kv_threshold = self.config.balance_token_usage_threshold;
        if (0.0..1.0).contains(&kv_threshold) {
            if let Some(max_usage) = self.max_backend_token_usage(workers, healthy_indices) {
                if max_usage > f64::from(kv_threshold) {
                    return true;
                }
            }
        }
```

ekzhang (commented):
Thank you so much!! Very excited for this, just a minor suggestion since I believe some of our workloads also benefit from the batch-size / concurrent requests balancing :D

PR #1642, opened 2026-06-10
fix(reasoning): use fresh parser on non-streaming path to avoid shared mutex

slin1237:
### Problem

Non-streaming reasoning extraction served all requests for a given model from one **pooled** parser instance (`Arc<Mutex<Box<dyn ReasoningParser>>>`). Every non-streaming request would `lock().await` that single mutex, `reset()` the parser, run `detect_and_parse_reasoning`, then drop the state — so concurrent requests for the same model **serialized** on one mutex despite needing no shared state. The streaming path already avoids this by creating a fresh parser per request/index for state isolation.

From the codebase audit: crates/reasoning_parser/src/factory.rs:64 — Non-streaming requests serialize on one shared parser mutex per model.

### Solution

Switch the non-streaming paths to a fresh, owned per-request parser via `create_parser` / `create_reasoning_parser` (the same approach the streaming path already uses). A fresh parser is inherently clean, so the explicit `reset()` and the `lock().await` are no longer needed, removing the per-model contention entirely. The now-unused pooled helper `get_reasoning_parser` is dropped.

## Changes

- `model_gateway/src/routers/grpc/regular/processor.rs`: both non-streaming sites (chat completions + Anthropic messages) now take a fresh parser via `utils::create_reasoning_parser(...)` instead of the pooled parser + `lock().await` + `reset()`.
- `model_gateway/src/routers/parse/handlers.rs`: the `/separate_reasoning` endpoint uses `registry().create_parser(...)` instead of `get_pooled_parser(...).lock().await`.
- `model_gateway/src/routers/grpc/utils/parsers.rs`: removed the now-unused `get_reasoning_parser` pooled helper (and its unused `PooledParser` import); refreshed the `create_reasoning_parser` doc to note it serves both streaming and non-streaming. Added unit tests.
- `model_gateway/src/routers/grpc/utils/mod.rs`: dropped the `get_reasoning_parser` re-export.

The tool-parser pooled path is intentionally left unchanged (separate concern/finding).

## Test Plan

Added focused unit tests in `parsers.rs`:
- `create_reasoning_parser_returns_independent_instances` — two parsers created for the same model are independent: `mark_reasoning_started()` on one does not flip the other's `is_in_reasoning()`, and the untouched instance still parses a full `<think>…</think>` document correctly. This is exactly the isolation the shared pooled parser violated.
- `create_reasoning_parser_honors_configured_parser` — the configured-parser path resolves to the requested parser.

Gate (sccache disabled, scoped to `smg`):

```
$ cargo +nightly fmt --all
(clean, exit 0)

$ RUSTC_WRAPPER="" cargo clippy -p smg --all-targets -- -D warnings
    Checking smg v1.4.1
    Finished `dev` profile [unoptimized + debuginfo] target(s)
(no warnings, exit 0)

$ RUSTC_WRAPPER="" cargo test -p smg
running 2 tests
test routers::grpc::utils::parsers::tests::create_reasoning_parser_honors_configured_parser ... ok
test routers::grpc::utils::parsers::tests::create_reasoning_parser_returns_independent_instances ... ok
...
test result: ok. 1005 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out
     Running tests/api_tests.rs
test result: ok. 105 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
     Running tests/routing_tests.rs
test result: ok. 92 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
     Running tests/spec_test.rs
test result: ok. 95 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
     Running tests/security_tests.rs
test result: ok. 48 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
     (all remaining integration test binaries: ok, 0 failed)
```

<details>
<summary>Checklist</summary>

</details>

## Refactor
* Improved reasoning parser handling by creating independent parser instances per request instead of using shared pooled parsers, enhancing reliability and reducing contention.

## Tests
* Added tests to verify parser isolation and correct parser selection based on configuration.

### DIFF
--- a/model_gateway/src/routers/grpc/regular/processor.rs
+++ b/model_gateway/src/routers/grpc/regular/processor.rs
@@ -97,37 +97,35 @@ impl ResponseProcessor {
         let mut processed_text = final_text;
 
         if original_request.separate_reasoning && reasoning_parser_available {
-            let pooled_parser = utils::get_reasoning_parser(
+            // Fresh parser per request: non-streaming extraction keeps no state
+            // across requests, so avoid serializing on the shared pooled mutex.
+            if let Some(mut parser) = utils::create_reasoning_parser(
                 &self.reasoning_parser_factory,
                 self.configured_reasoning_parser.as_deref(),
                 &original_request.model,
-            );
-
-            let mut parser = pooled_parser.lock().await;
-            // Reset pooled parser to clean state before each request
-            parser.reset();
-
-            // If the template injected `<think>` in the prefill (thinking toggle
-            // is supported and effectively ON), start in reasoning mode.
-            if utils::should_mark_reasoning_started(
-                utils::extract_thinking_from_kwargs(
-                    original_request.chat_template_kwargs.as_ref(),
-                    tokenizer.as_ref(),
-                ),
-                tokenizer.as_ref(),
             ) {
-                parser.mark_reasoning_started();
-            }
-
-            match parser.detect_and_parse_reasoning(&processed_text) {
-                Ok(result) => {
-                    if !result.reasoning_text.is_empty() {
-                        reasoning_text = Some(result.reasoning_text);
-                    }
-                    processed_text = result.normal_text;
+                // If the template injected `<think>` in the prefill (thinking toggle
+                // is supported and effectively ON), start in reasoning mode.
+                if utils::should_mark_reasoning_started(
+                    utils::extract_thinking_from_kwargs(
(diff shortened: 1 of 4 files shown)


CatherineSue (approved):

CatherineSue:
Maybe we can remove the shared mutex after this PR? Right now I don't see the usage of get_parser.

PR #1669, opened 2026-06-11
fix(gateway): DP-aware rank suffix

ekzhang:
### Problem

`--remove-unhealthy-workers` did not work with DP-aware ranks, they would try to remove a URL ending with `@`, which wasn't detected.

### Solution

Use `base_url()` instead of `url()` when deciding which workers to remove in the `--dp-aware` mode.

## Changes

- Fixed bug in `--dp-aware --remove-unhealthy-workers`
- Ran `cargo +nightly fmt` and `cargo clippy --all-targets`

## Test Plan

- Added test for regression

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed worker removal job handling to consistently use base URLs for differential privacy-configured workers, ensuring proper prefix-matching and removal workflow reliability.

* **Tests**
  * Added regression test to validate correct URL stripping behavior for DP-configured workers.

### DIFF
--- a/model_gateway/src/worker/manager.rs
+++ b/model_gateway/src/worker/manager.rs
@@ -403,7 +403,7 @@ fn queue_due_probes(
             if config.remove_unhealthy {
                 removals.push(RemovalCandidate {
                     worker_id: worker_id.clone(),
-                    url: worker.url().to_string(),
+                    url: worker.base_url().to_string(),
                     expected_revision,
                 });
             }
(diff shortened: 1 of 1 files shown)


CatherineSue (approved):

ekzhang:
Hm I'm not sure what the test failure is caused by!

PR #1670, opened 2026-06-11
fix(messages): Anthropic messages API, preserve system "text" and unknown fields

ekzhang:
### Problem

When smg proxies /v1/messages requests, it deserializes then re-serializes the body. Two issues caused the re-serialized payload to diverge from the original:

1. System content blocks lost their `type: "text"` discriminator because `SystemContent::Blocks` used `Vec<TextBlock>` directly and `TextBlock` has no `type` field. Backends (e.g. sglang) that validate the discriminator rejected the forwarded request. Fix: wrap `TextBlock` in a new `SystemContentBlock` tagged enum so `type` is emitted on serialization.

2. Unknown/beta Anthropic API fields (`context_management`, `output_config`, etc.) were silently dropped because `CreateMessageRequest` lacked a `#[serde(flatten)]` catch-all. Fix: add `other: Map<String, Value>` with flatten, matching the existing pattern in `ChatCompletionRequest`.

### Solution

This fixes the round-trip for those fields and adds tests.

## Changes

- Added `text` to round-trip of system content block so the type field is not dropped.
- Added a catch-all for unknown fields that aren't handled yet, so they can be forwarded safely to the HTTP backend.
- Ran `cargo +nightly fmt` and `cargo clippy --all-targets`

## Test Plan

- Added tests for both behaviors (regression tests)

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling and preservation of system content blocks during message processing and serialization round-trips.
  * Enhanced system prompt extraction logic to properly support typed content blocks.

* **Tests**
  * Expanded test coverage to verify preservation of content type information across serialization cycles and proper handling of message fields.

### DIFF
--- a/crates/protocols/src/messages.rs
+++ b/crates/protocols/src/messages.rs
@@ -6,7 +6,7 @@
 use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
+use serde_json::{Map, Value};
 use validator::Validate;
 
 use crate::{common::GenerationRequest, validated::Normalizable};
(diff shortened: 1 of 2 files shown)


ekzhang:
@knlb-tml wrote this diff, I'm just upstreaming it! (verified tests pass, clippy works, reformatted, etc.)

PR #1688, opened 2026-06-11
refactor(service-discovery): migrate k8s watcher to informer (reflector + Store)

key4ng:
### Problem

SMG's Kubernetes service discovery (`model_gateway/src/service_discovery.rs`) is built directly on a raw `kube::runtime::watcher()` consumed through `.applied_objects()`. Because that gives a bare event stream with **no cached state** and, critically, **no delete events**, a layer of hand-written workarounds had accreted to paper over the gaps a raw watcher leaves behind:

1. **Periodic full-LIST reconciliation** (`reconcile_pods`, every `check_interval` = 60s): a full `pods.list()` against the API server, diffed against an in-memory tracked set to catch events the watcher missed (force-deleted pods, pods created while the watcher was restarting).
2. **Manual watcher restart loop with exponential backoff** (1s → 300s): the stream wrapped in an infinite `loop` that re-establishes the watch on error or stream completion.
3. **Reconciliation supervisor** that catches panics in the reconcile task and restarts it.
4. **Delete detection via `deletion_timestamp`**: since `.applied_objects()` drops delete events, deletions were inferred from a `deletion_timestamp` on a *modified* object — which only fires during the graceful-termination window and **misses hard/force deletes entirely**. The router/mesh watcher has no reconcile safety net, so hard-deleted router pods were only ever detected via mesh SWIM timeout.

(1)–(3) are emulating, by hand, exactly what a Kubernetes **informer** provides natively: an always-consistent in-memory cache (initial LIST → watch → automatic re-LIST on desync) plus self-healing reconnect.

### Solution

Migrate both watchers (worker discovery and the optional router/mesh discovery) to the kube-rs **informer** pattern: `reflector` + `Store` + `WatchStreamExt::default_backoff()`, consuming real `watcher::Event<Pod>`.

- `watcher(api, cfg).default_backoff().reflect(writer)` maintains an in-memory `Store<Pod>`. The stream **never terminates** — transient watch errors trigger an internal exponential-backoff reconnect, and a desync (410 Gone / "too old resource version") triggers an automatic re-LIST.
- Events are dispatched directly: `Apply`/`InitApply` → add (idempotent), `Delete` → remove (**real delete events**, incl. hard deletes), `InitDone` → reconcile the tracked set against the `Store` snapshot. The `InitDone` reconcile fires on initial sync **and** on every re-LIST after a desync — exactly when drift can occur — replacing the timer-driven full-LIST reconcile with **zero** API calls.

This removes the watcher-gap workarounds while leaving the downstream worker-registration pipeline (`PodInfo → WorkerSpec → Job::AddWorker → JobQueue → create_worker`), UID-restart eviction, idempotent dedup, and `(name, uid)` identity equality unchanged.

Considered and rejected: the full `kube::runtime::Controller`/`applier` (level-triggered reconciler). It duplicates the existing `Job`/`JobQueue` actuation layer (two work queues) and its owner/finalizer model is overkill for plain Pod discovery.

## Changes

Single file: `model_gateway/src/service_discovery.rs` (+341 / −287).

**Removed (all watcher-gap workarounds):**

| Removed | Replaced by |
| --- | --- |
| Periodic full `pods.list()` reconcile (`reconcile_pods` timer loop) | `reconcile_from_store` on `Event::InitDone` (in-memory `Store` snapshot) |
| Manual `loop { watcher … }` restart + exp. backoff (worker **and** router) | `.default_backoff()` |
| Reconcile supervisor (panic-restart wrapper, `interval_at`) | gone with the timer |
| `.applied_objects()` + `deletion_timestamp` delete inference | native `Event::Delete` |

**Added / rewritten:**

- `dispatch_worker_event(event, store, …)` — routes `Event<Pod>`: `Apply`/`InitApply` → `handle_pod_event`, `Delete` → `handle_pod_deletion`, `InitDone` → `reconcile_from_store`, `Init` → relist marker.
- `reconcile_from_store(store, …)` — replaces `reconcile_pods`; identical diff/act logic (reuses `build_live_pod_set` + `compute_reconciliation_diff`), but the live set comes from `store.state()` instead of an API `list()`.
- Worker loop in `start_service_discovery` — now `watcher().default_backoff().reflect(writer)` + an event-match loop; restart loop and reconcile supervisor deleted.
- `apply_router_event(event, …)` + rewritten `start_router_discovery` — reacts to real `Event::Delete` (**hard-deleted router pods now marked `Down` promptly** instead of only via mesh SWIM timeout), preserves the graceful-termination `deletion_timestamp → Down` path, uses `default_backoff`, no `Store` (mesh CRDT/SWIM handles convergence).

**Kept unchanged (legitimate, not watcher-gap hacks):** UID-restart eviction in `handle_pod_event`, tracked-set idempotent dedup, `(name, uid)` identity equality, the entire downstream registration pipeline, all metric names.

**Backward compatibility:** `ServiceDiscoveryConfig` and all CLI flags are unchanged. `check_interval` is retained for config/CLI compatibility but is now a **deprecated no-op** (the informer self-heals); a one-time `warn!` is logged at startup if it is set to a non-default value.

### Unit tests (`cargo test -p smg --lib service_discovery`)

`79 passed`. Existing tests for `PodInfo`, `ModelIdSource`, `build_live_pod_set`, `compute_reconciliation_diff`, `build_watcher_config`, and UID-restart eviction are retained (logic reused). New tests added:

- `test_dispatch_worker_event_apply_then_delete` — `Apply` adds, `Delete` removes (the core delete-event fix).
- `test_dispatch_worker_event_init_apply_adds` — `InitApply` adds.
- `test_dispatch_worker_event_init_done_adds_missing_from_store` — `InitDone` reconciles missing pods from a populated `Store` snapshot.
- `test_apply_router_event_delete_marks_down` — router `Apply`→`Alive`, `Delete`→`Down`.

### Real-cluster validation (before/after behavior)

The **actual `smg` binary built from this branch** was run against a fresh `kind` cluster (k8s v1.36.1), watching `--selector app=fakeworker` in namespace `sd-test`, with `RUST_LOG=info,smg::service_discovery=debug`. "Workers" were `registry.k8s.io/pause:3.10` pods (discovery is under test, not worker health).

```
smg launch --service-discovery --selector app=fakeworker \
  --service-discovery-namespace sd-test --service-discovery-port 8000 \
  --port 30080 --prometheus-port 29555
```

**1. Apply → add.** Create pod `w1`:
```
INFO smg::service_discovery: Adding pod: w1 | type: Some(Regular) | url: 10.244.0.5:8000
```

**2. Hard delete → real `Event::Delete` → immediate remove (the headline fix).** `kubectl delete pod w1 --grace-period=0 --force`:
```
INFO smg::service_discovery: Removing pod: w1 | type: Some(Regular) | url: 10.244.0.5:8000
```
Fired **immediately**. *Before* this PR, a force-delete emits no usable event and is only caught up to `check_interval` (60s) later by the periodic full-LIST reconcile. *After*, it is caught instantly via `Event::Delete`.

**3. Startup initial-sync (Init → InitApply → InitDone) rediscovers pre-existing pods.** With `w2`, `w3` already running, `smg` restarted:
```
INFO  smg::service_discovery: Starting K8s worker watcher | selector: 'app=fakeworker'
DEBUG smg::service_discovery: K8s worker watcher (re)list starting         # Event::Init
INFO  smg::service_discovery: Adding pod: w2 | type: Some(Regular) | url: 10.244.0.6:8000
INFO  smg::service_discovery: Adding pod: w3 | type: Some(Regular) | url: 10.244.0.7:8000
```

**4. Self-heal on apiserver blip (no manual restart loop).** `docker restart <kind-control-plane>` drops the apiserver + watch connection. Same `smg` pid throughout:
```
ERROR smg::service_discovery: K8s worker watcher error (auto-retrying with backoff): watch stream failed: error reading a body from connection
ERROR smg::service_discovery: K8s worker watcher error (auto-retrying with backoff): failed to start watching object: client error (Connect)
ERROR smg::service_discovery: K8s worker watcher error (auto-retrying with backoff): error returned by apiserver during watch: too old resource version: 1880 (1890): Expired
DEBUG smg::service_discovery: K8s worker watcher (re)list starting          # automatic re-LIST after desync
INFO  smg::service_discovery: Adding pod: w4 | type: Some(Regular) | url: 10.244.0.7:8000
```
The `Expired` (410-equivalent) desync triggers the watcher's automatic re-LIST; discovery resumes with **no manual restart** — exactly what `default_backoff` + reflector replaced the hand-written restart loop and periodic reconcile with.

**Not exercised live (covered by unit tests):** router/mesh `Event::Delete → Down` (same event mechanism as the worker delete above; `test_apply_router_event_delete_marks_down`), and StatefulSet same-name/new-uid eviction (unchanged logic; `test_handle_pod_event_evicts_old_uid_on_restart`).

<details>
<summary>Checklist</summary>

- [x] (Optional) Documentation updated — `check_interval` field doc-comment marks it deprecated/no-op

</details>

## Summary by CodeRabbit

* **Infrastructure & Reliability**
  * Switched service discovery to event-driven, self-healing watchers for workers and routers.
  * Improved pod lifecycle handling and router state transitions for more accurate, real-time status.
  * More efficient reconciliation that reduces unnecessary API calls and improves performance.
  * Removed explicit service-discovery "check interval" setting; system defaults are now used.

* **Tests**
  * Added and updated unit tests to validate the new discovery behavior.

### DIFF
--- a/model_gateway/src/service_discovery.rs
+++ b/model_gateway/src/service_discovery.rs
@@ -4,7 +4,6 @@ use std::{
     // (HashSet insert/remove/contains) and never cross .await boundaries.
     // See: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use
     sync::{Arc, Mutex},
-    time::Duration,
 };
 
 use futures::StreamExt;
@@ -90,10 +89,6 @@ impl ModelIdSource {
 pub struct ServiceDiscoveryConfig {
     pub enabled: bool,
     pub selector: HashMap<String, String>,
-    /// Deprecated: no longer used. The informer self-heals via watch + backoff
-    /// and reconciles against its in-memory `Store` on every re-LIST. Retained
-    /// for config/CLI backward compatibility.
-    pub check_interval: Duration,
     pub port: u16,
     pub namespace: Option<String>,
     // PD mode specific configuration
@@ -167,7 +162,6 @@ impl Default for ServiceDiscoveryConfig {
         ServiceDiscoveryConfig {
             enabled: false,
             selector: HashMap::new(),
-            check_interval: Duration::from_secs(60),
             port: 8000,
             namespace: None,
             pd_mode: false,
@@ -410,14 +404,6 @@ pub async fn start_service_discovery(
     let _ = ring::default_provider().install_default();
 
     let client = Client::try_default().await?;
-
-    if config.check_interval != ServiceDiscoveryConfig::default().check_interval {
-        warn!(
-            "check_interval ({}s) is deprecated and no longer used: the informer self-heals via \
-             watch + backoff and reconciles against its in-memory Store on every re-LIST",
-            config.check_interval.as_secs()
-        );
-    }
 
     // Log the appropriate selectors based on mode
     if config.pd_mode {
@@ -1233,7 +1219,6 @@ mod tests {
         ServiceDiscoveryConfig {
             enabled: true,
             selector: HashMap::new(),
-            check_interval: Duration::from_secs(60),
             port: 8080,
             namespace: None,
             pd_mode: true,
@@ -1360,8 +1345,6 @@ mod tests {
 
     #[test]
     fn test_apply_router_event_delete_marks_down() {
-        use std::collections::BTreeMap;
-
         let mut router_selector = HashMap::new();
         router_selector.insert("role".to_string(), "router".to_string());
         let config = ServiceDiscoveryConfig {
@@ -1370,10 +1353,11 @@ mod tests {
             ..ServiceDiscoveryConfig::default()
         };
 
-        let cluster_state: ClusterState = Arc::new(parking_lot::RwLock::new(BTreeMap::new()));
+        let cluster_state: ClusterState =
+            Arc::new(parking_lot::RwLock::new(std::collections::BTreeMap::new()));
 
         // Build a router pod and pre-seed it as Alive via an Apply event.
-        let mut labels = BTreeMap::new();
+        let mut labels = std::collections::BTreeMap::new();
         labels.insert("role".to_string(), "router".to_string());
         let mut pod = create_k8s_pod(
             Some("r1"),
@@ -1427,7 +1411,6 @@ mod tests {
         let config = ServiceDiscoveryConfig::default();
         assert!(!config.enabled);
         assert!(config.selector.is_empty());
-        assert_eq!(config.check_interval, Duration::from_secs(60));
         assert_eq!(config.port, 8000);
         assert!(config.namespace.is_none());
         assert!(!config.pd_mode);
--- a/model_gateway/src/service_discovery.rs
+++ b/model_gateway/src/service_discovery.rs
@@ -7,12 +7,13 @@ use std::{
     time::Duration,
 };
 
-use futures::{StreamExt, TryStreamExt};
+use futures::StreamExt;
 use k8s_openapi::api::core::v1::Pod;
 use kube::{
-    api::{Api, ListParams},
+    api::Api,
     runtime::{
-        watcher::{watcher, Config},
+        reflector::{self, Store},
+        watcher::{watcher, Config, Event},
         WatchStreamExt,
     },
     Client,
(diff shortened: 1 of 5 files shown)


model_gateway/src/service_discovery.rs
```diff
@@ -89,6 +90,9 @@ impl ModelIdSource {
 pub struct ServiceDiscoveryConfig {
     pub enabled: bool,
     pub selector: HashMap<String, String>,
+    /// Deprecated: no longer used. The informer self-heals via watch + backoff
+    /// and reconciles against its in-memory `Store` on every re-LIST. Retained
+    /// for config/CLI backward compatibility.
     pub check_interval: Duration,
```
CatherineSue: Should we remove this? I feel not many users actively use this field.

slin1237: all original hacks and a internal check should be removed

Issue #1694, opened 2026-06-12
[Perf] /health can be starved by the request runtime; /readiness is O(workers)


slin1237:
### Area

Liveness/readiness can be starved by the request runtime; `/readiness` is O(workers).

### Problem

`/health` → `liveness()` is trivial — it returns `"OK"` (`server.rs:101-103`, `/health` at `:170`). But it is served from the same tokio runtime and middleware stack as everything else, so when worker threads are saturated (CPU-bound work + O(workers) routing) the runtime can't schedule even this trivial future within the probe timeout → k8s kills the pod → restart loop. Separately, `/readiness` (`server.rs:106`) calls `get_all()` and filters all workers **per probe** (O(workers)).

### Evidence (mock-fleet baseline)

`/health` latency, sampled directly on the gateway:

| Workers | `/health` (typical) | `/health` (spike) |
|---|---|---|
| 500 | 0.0007 s | 0.0007 s |
| 2000 | 0.0009 s | **3.32 s** |
| 4000 | 0.0005 s | **2.30 s** |

The trivial liveness handler blocks for **2–3.3 s** at 2000–4000 workers under registration/load bursts (fine at 500). This is the reporter's failing-`/health` mechanism, reproduced.

### Direction

- Serve `/health`/`/liveness` from a dedicated OS thread + tiny runtime (or a separate listener) so liveness cannot be starved by the request runtime — the standard fix for this failure mode.
- Make `/readiness` O(1) (maintain a cached healthy/total count; don't `get_all()` per probe).

Refs #1685.

ekzhang:
Hmm this totally makes sense -- I also just wonder though why handlers on this thread pool would block for 2-3 seconds and if we can avoid that, since it seems that would add 2-3 seconds of latency to `/generate` requests as well (not ideal)

I tried allocating more CPU cores in Kubernetes to the gateway pods, it's now at 12 cores (in the prev report), and it never really gets all the way there. (One blind guess, that we might try: it could be memory allocator locking and might help to to switch to jemalloc/mimalloc?)

Issue #1685, opened 2026-06-11
[Performance]: High CPU usage in `spawn_worker_collector`


ekzhang:
### Performance Issue Description

Seeing a bit of higher CPU usage - the below trace was snapshotted over 20s on a `sgl-model-gateway` server in Kubernetes, while having 4500 workers (registered in IGW mode) and using 2 CPU cores. 40% of the samples in the trace (~0.8 CPU cores) was spent on the `metrics_ws` collector for workers.

<img width="491" height="296" alt="Image" src="https://github.com/user-attachments/assets/540d25d9-8f7d-457a-9e4d-0ba8cdbb848a" />

My understanding from reading the code is that we could improve this by reducing `worker_checkpoint_interval` which is set to 3s right now -- is that right? 3s seems quite often and it is a fallback for non-subscribed updates.

### Issue Type

High CPU usage

### Performance Metrics

CPU: spiked between 2-4 cores, the flamegraph was taken around the right end of the interval in this graph for 20s.

<img width="758" height="238" alt="Image" src="https://github.com/user-attachments/assets/832fa178-d1d6-465b-8e62-6f2586390576" />

### Expected Performance

Just hoping to reduce CPU usage a bit ahead of scaling up more! Anecdotally, even though we allocate 12 cores for the gateway deployment and set tokio threads to 32, when CPU usage goes above 2-3 cores the gateway starts being a bit unstable and failing /health probes.

### Load Profile

During measured period
 
- Concurrent connections: about 1M
- Request rate: about 3k /s
- Worker count: 4000

### Routing Policy

cache_aware

### Connection Mode

HTTP

### Configuration

```yaml

```

### Environment

Linux 6.12
Ubuntu 24.04

### Deployment Environment

Kubernetes

### Backend Workers

SGLang

### Pre-submission Checklist

- [x] I have verified this is not a backend worker issue
- [x] I have collected metrics during the performance issue
- [x] I can reproduce this issue

ekzhang:
Thank you @slin1237 and @CatherineSue  ! CPU usage is looking much better now

<img width="782" height="242" alt="Image" src="https://github.com/user-attachments/assets/f734bc23-a6c1-4423-9d73-c00494f7249e" />

PR #1711, opened 2026-06-13
perf(server): O(1) readiness + optional dedicated probe listener (--health-check-port), drain-aware probes

slin1237:
## Summary
Kubernetes probes become O(1) and optionally unstarvable, and draining finally flips readiness. Rebased on current main.

The incident mechanism: `/health` is trivial but lives on the request runtime behind the full middleware stack — when worker threads are saturated it takes 2–3.3s and kubelet kills the pod; `/readiness` additionally `get_all()`-scans the whole fleet **per probe** (O(workers): Arc-clones + per-worker tokenizer lookups).

## H1 — O(1) readiness
`ProbeState` keeps a `ReadinessSnapshot` behind an `ArcSwap`; the handler is now one drain-flag load + one pointer load. A maintainer task recomputes the snapshot on registry events (bursts coalesced) plus a 1s checkpoint.

**Why recompute + checkpoint rather than pure incremental counters:** the registry event stream is provably incomplete — the *primary become-ready path* (`ActivateWorkersStep`) calls `set_status(Ready)` directly without emitting `StatusChanged`, as do the mesh health refresh and Go FFI; tokenizer registrations emit no worker event at all. Event-emitting paths update the snapshot immediately; bypass paths converge within ≤1s (one ~µs registry scan/sec vs one per probe per kubelet today). `recompute_with` is a verbatim relocation of the old inline `/readiness` logic — same `enable_igw` short-circuit, same PD prefill+decode requirement, same tokenizer-autoload gating, byte-identical JSON bodies.

## H2 — dedicated probe listener via `--health-check-port`
A new CLI flag `--health-check-port` (and Python `health_check_port=`), unset = off. When set, `/liveness`, `/readiness`, `/health` are *also* served on that port by a minimal router with **zero middleware**, from a 1-worker runtime on its own OS thread — so probes cannot be starved by the request runtime, and keep answering through the entire drain. Main-listener routes are unchanged; point k8s probes at the dedicated port to opt in.

Plumbed as a first-class config value mirroring the main listener `--port` exactly: CLI arg (`main.rs`), `RouterConfig`/`ServerConfig` fields with `#[serde(default, skip_serializing_if)]`, **both** `to_router_config()` and `to_server_config()`, `Default`, and the Python bindings (`bindings/python/src/lib.rs` + `router_args.py` argparse). The Python flag is registered outside the `exclude_host_port` guard so embedded-server deployments can still enable probe isolation. (Replaces the `SMG_PROBE_PORT` env var from the first revision of this PR.)

## G2 — drain flips readiness
The `draining` flag was write-only. `readiness_response()` now returns `503 {"reason":"draining"}` on both listeners while liveness stays 200, so k8s stops routing new connections during the propagation window. No signal-path changes.

## Tests
11 probe tests (readiness state machine: add/ready/not-ready/remove, PD mode, IGW mode, tokenizer gating + HTTP exemption; drain flip vs liveness; probe router via tower oneshot; real ephemeral-socket test of the dedicated listener + drain over HTTP) plus config-plumbing tests proving the flag reaches both configs, defaults to `None`, and is range-rejected at parse time. Workspace clippy `-D warnings` clean, 3616 tests pass.

## Notes
- `metrics_ws/` deliberately untouched.
- Naming: `--health-check-port` per request. SMG already has unrelated `--health-check-timeout-secs` / `--health-check-endpoint` flags that configure *worker* health checks (not this gateway probe-listener port) — no collision, but the shared `health-check` prefix could read as related; `--probe-port` would be less ambiguous if you'd prefer to rename.
- Follow-ups (unchanged from review): the maintainer runs on the request runtime, so under total starvation readiness serves stale-but-fast (the intended failure mode); routing the `set_status` bypass paths through registry events would make readiness purely event-driven.

Refs #1694 #1685.

## Release Notes

* **New Features**
  * Added optional `--health-check-port <port>` to serve `/liveness`, `/readiness`, and `/health` on a dedicated port for more stable probe latency (also exposed via the Python Router API).
* **Bug Fixes**
  * Updated readiness/drain semantics: `/readiness` returns `503` with reason `"draining"` during drain, and can return `503` when no healthy workers remain; `/health` and `/liveness` stay `200`.
* **Documentation**
  * Refreshed graceful-shutdown guidance to match the updated readiness behavior.
* **Configuration**
  * Rejects `--health-check-port 0` as invalid.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -366,6 +366,7 @@ impl PyPostgresConfig {
 struct Router {
     host: String,
     port: u16,
+    health_check_port: Option<u16>,
     worker_urls: Vec<String>,
     policy: PolicyType,
     worker_startup_timeout_secs: u64,
@@ -685,6 +686,7 @@ impl Router {
             .policy(policy)
             .host(&self.host)
             .port(self.port)
+            .health_check_port(self.health_check_port)
             .connection_mode(self.connection_mode)
             .max_payload_size(self.max_payload_size)
             .request_timeout_secs(self.request_timeout_secs)
@@ -777,6 +779,7 @@ impl Router {
         policy = PolicyType::RoundRobin,
         host = String::from("0.0.0.0"),
         port = 3001,
+        health_check_port = None,
         worker_startup_timeout_secs = 600,
         worker_startup_check_interval = 30,
         load_monitor_interval = 10,
@@ -893,6 +896,7 @@ impl Router {
         policy: PolicyType,
         host: String,
         port: u16,
+        health_check_port: Option<u16>,
         worker_startup_timeout_secs: u64,
         worker_startup_check_interval: u64,
         load_monitor_interval: u64,
@@ -1016,6 +1020,7 @@ impl Router {
         Ok(Router {
             host,
             port,
+            health_check_port,
             worker_urls,
             policy,
             worker_startup_timeout_secs,
@@ -1187,6 +1192,7 @@ impl Router {
             Box::pin(server::startup(server::ServerConfig {
                 host: self.host.clone(),
                 port: self.port,
+                health_check_port: self.health_check_port,
                 router_config,
                 max_payload_size: self.max_payload_size,
                 log_dir: self.log_dir.clone(),
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -779,7 +779,6 @@ impl Router {
         policy = PolicyType::RoundRobin,
         host = String::from("0.0.0.0"),
         port = 3001,
-        health_check_port = None,
         worker_startup_timeout_secs = 600,
         worker_startup_check_interval = 30,
         load_monitor_interval = 10,
@@ -885,6 +884,10 @@ impl Router {
         mesh_advertise_host = None,
         drain_settle_secs = 5,
         enable_wasm = false,
+        // Appended last (not inserted mid-list) so every pre-existing
+        // positional argument keeps its index for callers that construct
+        // `_Router(...)` positionally. See the struct-field note above.
+        health_check_port = None,
     ))]
     #[expect(clippy::too_many_arguments)]
     #[expect(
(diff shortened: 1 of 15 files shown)


slin1237:
Addressed the review comments in `e02427b4` (on top of the `--health-check-port` rework and the move of the module out of `observability/` to crate-root `health.rs`):

- **Probe host bind (coderabbit / claude / codex):** now parses `host:port` as a `SocketAddr` exactly like the main listener — bracketed IPv6 literals (`[::1]`, `[::]`) work, and a bad host is a hard startup error instead of a silent rebind to `0.0.0.0`. (`start_probe_listener` still accepts port `0` for the ephemeral-port unit tests.)
- **Maintainer lifetime / reference cycle (gemini):** the task now holds `Weak` references to the worker/tokenizer registries and `upgrade()`s each iteration, breaking when the owner drops them — so the task and registries deallocate on teardown. New test `maintainer_exits_when_registry_dropped`. (Noted: the existing `metrics_ws` worker collector shares the same strong-Arc pattern — pre-existing, out of scope here.)
- **Python positional arg (codex):** `health_check_port` is appended **last** in the `Router` `#[new]` signature, so existing positional callers keep their argument positions.
- **Reject `--health-check-port 0` (coderabbit / codex):** rejected at the config-validation layer (matching the existing `--port 0` check) with test `test_reject_health_check_port_zero`; the listener itself still accepts `0` for tests.
- **Checkpoint missed-tick (coderabbit):** `MissedTickBehavior::Skip` — a starved maintainer takes one catch-up tick, not a burst (recompute is idempotent).
- **Probe runtime (claude):** `new_current_thread()` instead of `new_multi_thread().worker_threads(1)` — one fewer OS thread, identical isolation.

Skipped (with reason): the `wasm_test.rs` "ProbeState without maintainer" nit (those tests don't hit `/readiness`) and "logs started before serving" (the synchronous bind already catches the real failure — a port conflict — before the serving thread spawns).

Separately: the root cause behind the readiness checkpoint — `set_status()` mutating worker status without emitting `WorkerEvent::StatusChanged` (so `ActivateWorkersStep` and the mesh refresh bypass the broadcast) — will be fixed in a dedicated follow-up that consolidates status mutation onto the registry's emitting path. That also benefits the `metrics_ws` collector, which carries the same workaround. The checkpoint stays as a slow safety net afterward (tokenizer registration emits no event regardless).

clippy `-D warnings` clean; workspace tests 3618 passed / 0 failed.

PR #1714, opened 2026-06-13
fix(cache-aware): fix load imbalance when decode is faster than prefill

SYChen123:
### Problem

In some extreme scenarios, default cache_aware policy will lead to severe load imbalance problem.

For example, 32k input and output only 1 token. In such case, all the requests will be routed to the same prefill worker and others are always left unused.

The reason is that in the select_worker of CacheAwarePolicy, only workers[idx].load() is used for sorting. After the worker is selected, worker.load() will only be updated till create_streaming_response is called (after stream response is returned from decode worker to the router). If prefill is very slow and decode is very fast, then at most of the time, worker.load() will be the same for all prefill workers. Then pd-router will always select the first prefill worker.

## Changes

1. Create WorkerLoadGuard before request is sent to the worker for inference.
2. In select_worker, sort the workers according to (load, processed_requests, idx), instead of solely depending on load.

## Test Plan

1. Unit test test_streaming_load_tracking Passed
`cd ./model-gateway && TMPDIR=/private/tmp cargo test --lib test_streaming_load_tracking`

image

2. bench_serving
benching random dataset with 32k input and 1 token output, all the prefill workers have requests received.
image

<details>
<summary>Checklist</summary>

</details>

## Summary of changes

- **Performance Improvements**
  - Enhanced cache-aware worker selection by improving tie-breaking with additional workload metrics for more balanced routing.

- **Bug Fixes**
  - Fixed streaming load tracking to consistently apply to both streaming success and streaming decode-error responses.

- **Refactor**
  - Streamlined streaming response creation by passing a pre-built load-guard set through the streaming paths rather than constructing it internally.

- **Tests**
  - Updated token-tree routing coverage to use the full page-sized token sequence in the empty-indexer scenario.

### DIFF
--- a/model_gateway/src/policies/cache_aware.rs
+++ b/model_gateway/src/policies/cache_aware.rs
@@ -483,7 +483,7 @@ impl CacheAwarePolicy {
         // Use shortest queue when imbalanced
         let min_load_idx = healthy_indices
             .iter()
-            .min_by_key(|&&idx| workers[idx].load())
+            .min_by_key(|&&idx| (workers[idx].load(), workers[idx].processed_requests(), idx))
             .copied()?;
 
         let worker_url = workers[min_load_idx].url();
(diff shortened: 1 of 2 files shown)


slin1237 (approved):
lgtm
pending ci

SYChen123:
@slin1237 Thanks for your review and merge. Could you please merge the PR with the same modification in sglang as well? So that this fix can take effect in next version of sglang. Thanks a lot!
https://github.com/sgl-project/sglang/pull/27547

PR #1758, opened 2026-06-16
perf(cache-aware): gather worker routing state in one pass

slin1237:
### Problem

`cache_aware` worker selection made several O(workers) passes per request — the healthy filter (`get_healthy_worker_indices`), the `is_imbalanced` request-count fold, and the cache-hit tenant→worker / cache-miss min-load scans. Each per-worker access went through an `ArcSwap` guard: `is_healthy()`→`status()` and `load()` both load the worker's `runtime` guard, and `circuit_breaker_can_execute()` loads a second (circuit-breaker) guard. So a healthy worker was read ~3 times per request via separate guarded virtual calls.

In a no-GPU benchmark (4 reactor threads, saturating shared-prefix load, mock fleet) this per-worker guard traffic is what makes cache_aware's per-request CPU grow with worker count. gdb sampling at 2048 workers attributed the cost to `arc_swap` guard load/drop + per-worker dynamic dispatch in the selection loops — **not** the radix-tree insert, which is cheap and worker-count-independent.

### Solution

Gather everything the balanced and imbalanced paths need in a **single pass** over the workers, reading each worker once via a new `Worker::routing_state()` that shares the `runtime` guard for `status` + `load` + `processed_requests` (the circuit breaker is a separate `ArcSwap`, so it keeps its own guard). The pass collects the healthy indices, the load min/max for the imbalance count-spread, and the min-load index — with the `(load, processed_requests, idx)` tie-break from #1714, free here since `processed` rides the same guard. `is_imbalanced` and the min-load fallback consume the gathered values instead of re-scanning; a cache hit resolves its tenant→worker with a hash-free linear scan over the gathered healthy indices (`url()` is a cheap field read — no per-request map).

Net: ~3 guarded virtual calls/worker across ~3 passes → 2 guards + 1 virtual call/worker in a single pass. Worker selection is unchanged **except** that a cache hit no longer routes onto a `Ready`-but-circuit-broken worker — it now falls through to min-load, consistent with the rest of selection.

## Changes

- **`worker/worker.rs`**: add `RoutingState { healthy, can_execute, load, processed }` and `Worker::routing_state()`. The default impl composes the existing accessors; `BasicWorker` overrides it to read `status`/`load`/`processed` under one `runtime` guard.
- **`policies/cache_aware.rs`**: `select_worker` performs the single-pass gather; `is_imbalanced` takes the precomputed load bounds; the token-tree, string-tree, event-driven and min-load paths consume the gathered min-load index; cache-hit tenant→worker resolution is a hash-free scan over the gathered healthy indices.

## Test Plan

No-GPU simulation: 4 reactor threads (saturating), mock fleet (`--engine realistic`), 16 shared prefixes at 0.8 fraction, 8k offered req/s, `cache_aware` with the imbalance valve held open so the balanced (tree-route) path is exercised. Metric is gateway `cpu_ms/req` (CPU-time per completed request; lower is better), reported as the **marginal over `round_robin` at the same worker count** so tokenization/proxy overhead cancels and the policy cost is isolated. Single-run, so figures carry measurement noise (~±0.07 at these windows); the marginal is the robust signal.

Routing marginal over round_robin, HTTP path (string tree — not masked by tokenization, so routing cost is visible). The shipped (hash-free scan) and the earlier (per-request map) revisions were A/B'd at 2048 workers back-to-back under identical conditions:

| @2048 workers (same conditions) | cpu_ms/req | marginal vs round_robin |
|---|---|---|
| round_robin (baseline) | 1.684 | — |
| cache_aware — per-request `url→index` map (earlier revision) | 1.865 | +0.181 |
| **cache_aware — hash-free scan (shipped)** | **1.708** | **+0.024** |

The shipped version's cache_aware routing is **within noise of round_robin at 2048 workers (+0.024 cpu_ms/req)**, down from +0.181 with the per-request map and ~+0.40 unoptimized — the per-worker `ArcSwap` guard traffic that scaled with worker count is gone. At 64 workers the marginal is within measurement noise for all variants. gRPC path (token tree) is unchanged — per-request tokenization dominates there either way. (Simulator numbers; absolute magnitudes need a live A/B.)

Gates on this branch (rebased on `main`):

```
cargo +nightly fmt --check                          clean
cargo clippy -p smg --all-targets -- -D warnings    clean (rc=0)
cargo test -p smg --lib policies                     107 passed
cargo test -p smg --lib worker::                     207 passed
```

Full `cargo test -p smg --lib` shows one unrelated failure, `middleware::metrics::tests::distinct_ids_on_matched_route_do_not_grow_interner`, which passes in isolation (a shared-interner test sensitive to parallel execution) and is untouched by this change. `--all-features` is not run locally (it pulls in the `opencv-video` feature absent in this environment, per CONTRIBUTING); CI runs the full matrix.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Refactor**
  * Optimized load-balancing algorithm to reduce computational overhead during request routing and worker selection.
  * Improved cache-affinity routing logic to better align worker selection with cached content placement.
  * Enhanced worker state management for more consistent and efficient request distribution across available resources.

### DIFF
--- a/model_gateway/src/policies/cache_aware.rs
+++ b/model_gateway/src/policies/cache_aware.rs
@@ -59,8 +59,8 @@ use tokio::sync::watch;
 use tracing::{debug, warn};
 
 use super::{
-    get_healthy_worker_indices, normalize_model_key, utils::PeriodicTask, CacheAwareConfig,
-    LoadBalancingPolicy, SelectWorkerInfo,
+    normalize_model_key, utils::PeriodicTask, CacheAwareConfig, LoadBalancingPolicy,
+    SelectWorkerInfo,
 };
 use crate::{
     mesh::adapters::tree_sync::{RepairEntry, TreeRepairPage},
@@ -275,7 +275,18 @@ impl CacheAwarePolicy {
     /// - **count spread**: request-count dispersion (abs AND rel) over healthy
     ///   workers. Always evaluated, so high-count / low-KV imbalance is still
     ///   caught when KV looks even.
-    fn is_imbalanced(&self, workers: &[Arc<dyn Worker>], healthy_indices: &[usize]) -> bool {
+    /// Whether to abandon cache affinity for shortest-queue because the pool is
+    /// imbalanced — by backend KV usage (overload ceiling or hot-vs-cool spread)
+    /// or by request-count spread. `min_load`/`max_load` are the request-count
+    /// bounds over the healthy workers, which `select_worker` gathers in its
+    /// single worker pass (tests use the `imbalanced` helper to fold them).
+    fn is_imbalanced(
+        &self,
+        workers: &[Arc<dyn Worker>],
+        healthy_indices: &[usize],
+        min_load: usize,
+        max_load: usize,
+    ) -> bool {
         // KV-based triggers — need a load snapshot; both default 1.0 = disabled.
         if let Some((min_usage, max_usage)) =
             self.backend_token_usage_bounds(workers, healthy_indices)
@@ -291,14 +302,6 @@ impl CacheAwarePolicy {
         }
 
         // Count spread (abs AND rel) over healthy workers.
-        let (min_load, max_load) =
-            healthy_indices
-                .iter()
-                .fold((usize::MAX, 0usize), |(min, max), &idx| {
-                    let load = workers[idx].load();
-                    (min.min(load), max.max(load))
-                });
-        let min_load = if min_load == usize::MAX { 0 } else { min_load };
         max_load.saturating_sub(min_load) > self.config.balance_abs_threshold
             && (max_load as f32) > (min_load as f32 * self.config.balance_rel_threshold)
     }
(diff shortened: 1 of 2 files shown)


slin1237:
Addressed the review feedback (gemini-code-assist, chatgpt-codex, claude): dropped the per-request `url→index` HashMap — it allocated + hashed all worker URLs every request (incl. miss / imbalanced / event-driven paths that never used it). Cache-hit tenant resolution is now a hash-free linear scan over the already-gathered healthy indices (`url()` is a cheap field read). Also corrected the comment/PR note: a cache hit now (correctly) excludes circuit-broken workers and falls through to min-load, rather than chasing affinity into a CB-tripped worker. Force-pushed.

slin1237:
Re-measured the shipped (scan) version vs the earlier (map) version back-to-back at 2048 workers, same conditions:

| @2048w | cpu_ms/req | marginal vs round_robin |
|---|---|---|
| round_robin | 1.684 | — |
| cache_aware (per-request map) | 1.865 | +0.181 |
| cache_aware (hash-free scan, shipped) | 1.708 | **+0.024** |

Dropping the map cut the routing marginal +0.181 → +0.024 — cache_aware routing is now within noise of round_robin at 2048 workers (was ~+0.40 unoptimized). Test Plan updated.

PR #1793, opened 2026-06-19
feat(discovery): discover local HTTP model ids from /v1/models

SpencerGarnets:
## Summary

- classify OpenAI-compatible local frontends with `/v1/models.data[0].owned_by: nvidia` as local workers
- map `owned_by: nvidia` to the SGLang-compatible local HTTP metadata path
- use `/v1/models.data[0].id` as `served_model_name` and `.root` as `model_path` when stronger metadata did not provide them
- strip default SMG/SGLang-only request fields before generic HTTP forwarding, including true-by-default `separate_reasoning` and `stream_reasoning`
- add focused unit coverage for NVIDIA-owned local model discovery, backend detection, metadata identity, and default SGLang-field stripping

## Why

A Dynamo Frontend service can expose a healthy OpenAI-compatible `/v1/models` response with `owned_by: nvidia` while not populating SGLang-specific metadata endpoints. Before this change, SMG could classify/detect the worker incorrectly or register the model as `unknown`, so SMG `/v1/models` did not expose the real model id and real-model chat requests failed even though direct backend chat succeeded.

Dynamo Frontend also rejects SMG/SGLang-only request fields on its strict OpenAI-compatible HTTP surface. The pass-through HTTP router now removes unset/default SGLang-only fields before forwarding so strict local frontends accept the request, while preserving non-default fields such as `no_stop_trim: true`.

## Verification

On Linux VM, upstream-main PR worktree:

- `cargo fmt --all -- --check`
- `git diff --check`
- `timeout 900s cargo test -p smg --lib probe_models_owned_by_accepts_nvidia_as_local`
- `timeout 900s cargo test -p smg --lib detect_via_models_endpoint_maps_nvidia_to_sglang`
- `timeout 900s cargo test -p smg --lib strip_default_sglang_fields_removes_false_and_null_values`

Production compatibility was also validated on the v1.4.1-compatible branch with A4 Dynamo/SMG: SMG `/health` returned OK, `/v1/models` exposed `BlaiseAI/DeepSeek-V3.2-REAP-345B-SpinQuant-ActKV-NVFP4-NextN-Graft`, and `/v1/chat/completions` through SMG returned HTTP 200 `chat.completion` for that real model.

### DIFF
--- a/model_gateway/src/routers/http/router.rs
+++ b/model_gateway/src/routers/http/router.rs
@@ -54,6 +54,7 @@ use crate::{
         },
         error::{self, extract_error_code_from_response},
         grpc::utils::{error_type_from_status, route_to_endpoint},
+        openai::strip_default_sglang_fields,
         RouterTrait,
     },
     worker::{AttachedBody, ConnectionMode, Worker, WorkerLoadGuard, WorkerRegistry, WorkerType},
@@ -899,7 +900,7 @@ impl Router {
             }
         };
 
-        let json_val = match worker.prepare_request(json_val) {
+        let mut json_val = match worker.prepare_request(json_val) {
             Ok(prepared) => prepared,
             Err(e) => {
                 return error::bad_request(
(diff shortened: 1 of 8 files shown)


slin1237:
overall LGTM 
Can you fix the ci failure. Thanks for the contribution

PR #1818, opened 2026-06-22
feat(multimodal): configurable tensor transport + vLLM SHM and video

slin1237:
### Problem

The multimodal tensor transport was TokenSpeed-only and tuned exclusively via
`SMG_TOKENSPEED_MM_*` environment variables. There was no first-class
(CLI/config) way to select it, no per-worker control, and vLLM workers always
received large preprocessed tensors inline over gRPC. vLLM also rejected video
inputs entirely — which is exactly the case where avoiding the gRPC copy matters
most.

### Solution

Make the transport engine-agnostic and configurable, extend the shared-memory
(`/dev/shm`) transport to vLLM, and wire up vLLM video. The multimodal subsystem
receives resolved values and never depends on `RouterConfig`.

## Changes

**Transport config**
- New `--multimodal-tensor-transport` (`inline` | `shm` | `auto`) and
  `--multimodal-shm-min-bytes` CLI flags + `RouterConfig` fields, plus per-worker
  `WorkerSpec` overrides (`multimodal_tensor_transport`,
  `multimodal_shm_min_bytes`).
- Precedence: **worker spec → router config → `SMG_MM_*` env → built-in
  default**. Legacy `SMG_TOKENSPEED_MM_*` names are honored as fallback aliases.

**vLLM shared memory**
- Hoist `ShmHandle` / `RemoteTensorHandle` into `common.proto` and migrate
  TokenSpeed to them; add the transport `oneof` to vLLM `TensorData` (inline
  stays field 1 for wire compatibility).
- Gateway writes large tensors to `/dev/shm` via shared, engine-neutral
  `mm_shm` I/O (`smg-mm-` prefix, orphan sweep), with build- and send-path
  cleanup mirroring TokenSpeed.
- vLLM servicer reads payloads via a shared `mm_shm` helper and advertises
  `shm_namespace_id` (`GetServerInfo`) so `auto` can verify `/dev/shm`
  co-location.

**vLLM video**
- Assemble video inputs for vLLM (`is_video`) instead of rejecting them; the
  servicer routes the encoder tensor to `pixel_values_videos` with `video` field
  configs. (sglang/TRT remain image-only.)

**Docs**
- `docs/reference/configuration.md` updated for the flags, precedence, per-worker
  override, and vLLM support.

## Test Plan

Ran locally (macOS):
- `cargo +nightly fmt --all` — clean
- `cargo clippy -p smg -p smg-grpc-client -p openai-protocol --all-targets -- -D warnings` — clean (default features; `--all-features` pulls opencv)
- `cargo test -p smg` — green (the pre-existing `local_shm_namespace_id_resolves_on_linux` is Linux-only and skipped on macOS), `cargo test -p openai-protocol` — green
- `cargo check --workspace --tests`, `cargo build -p smg-python -p smg-golang` — clean
- Python: `mm_shm` unit tests (`/dev/shm` round-trip), servicer `py_compile` + `ruff`, proto regen field assertions

New tests:
- Rust: vLLM `into_proto` inline / below-threshold / image vs video modality
- Python: `grpc_servicer/tests/test_mm_shm.py` (shared SHM reader)
- E2E (`e2e_test/chat_completions/test_multimodal_shm.py`): Qwen3-VL image **and**
  video over vLLM gRPC with `--multimodal-tensor-transport shm`, asserting correct
  output **and** that `smg_mm_tensors_total{path="shm",runtime="vllm"}` increased
  (so a silent inline fallback fails the test). Requires a GPU runner; runs in CI.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets -- -D warnings` passes (default features)
- [x] Documentation updated

</details>

## Summary by CodeRabbit

* **New Features**
  * Added engine-agnostic multimodal tensor transport (`inline`, `shm`, `auto`) plus `--multimodal-shm-min-bytes` to control SHM eligibility.
  * Added multimodal routing for both images and videos, including SHM namespace identity reporting (`shm_namespace_id`) and unified SHM lifecycle cleanup behavior.
  * Introduced gateway-level, per-worker, and environment/CLI precedence with legacy environment alias support.
* **Bug Fixes**
  * Improved multimodal SHM cleanup on request/build failures and ensured the decode step strips multimodal inputs.
* **Tests**
  * Added end-to-end SHM verification for single-image and single-video requests on vLLM.
* **Documentation**
  * Updated configuration reference for both TokenSpeed and vLLM, including precedence and SHM-related environment variables.


slin1237:
Addressed all review feedback and rebased on `main`.

**Root cause of the vLLM e2e failure** (`launch_router.py: unrecognized arguments: --multimodal-tensor-transport`): the new flags weren't wired through the Python bindings. Added `multimodal_tensor_transport` / `multimodal_shm_min_bytes` to `RouterArgs` (argparse + dataclass) and the PyO3 `Router` (`to_router_config`), so the flags work via the Python launcher — and the e2e gateway can set them.

**Review comments**
- 🔴 **SHM filename security (gemini):** `mm_shm.validated_shm_name` now restricts names to the `smg-mm-` / `smg-tokenspeed-` prefixes, preventing arbitrary `/dev/shm` read/unlink.
- 🟠 **flat `sizes_key` KeyError (gemini):** the vLLM servicer raises a clear `ValueError` when a flat sizes key is absent from `model_specific_tensors`.
- 🟠 **byte-length validation (CodeRabbit):** `_tensor_from_proto` checks payload length against dtype×shape before `reshape`.
- 🟠 **placeholder bounds (CodeRabbit):** out-of-bounds / zero-length placeholder ranges are rejected before mask building.
- 🟠 **proto release (Codex P1):** bumped `smg-grpc-proto` 0.4.11 → 0.4.12 and the servicer dependency floor, so the new fields can't resolve against an old proto package.
- 🟡 **orphan sweep (Claude):** also reclaims legacy `smg-tokenspeed-*` crash files (rolling-upgrade safety).
- 🟡 **log message (Claude):** dropped "TokenSpeed" from the now-engine-agnostic transport-config log.
- 🟡 **unknown-transport warning (Claude):** dedupes per distinct value instead of once-ever, so a second worker misconfigured with a different typo still warns. (Kept `WorkerSpec` as a plain serde type rather than introducing an enum; the CLI still validates via `value_parser`.)

**pre-commit:** ran `ruff format`.

Rebased onto `main` (`b2af3740`). Local gate green: `cargo +nightly fmt`, `clippy -D warnings` (default features), `cargo test -p smg` (1100 lib + integration), `smg-python`/`smg-golang` builds, `mm_shm` unit tests, `ruff`.

PR #1821, opened 2026-06-23
test(e2e): don't assert exact image count for duplicate images

slin1237:
### Problem

`e2e_test/chat_completions/test_multimodal.py::TestMultimodalQwen3VL::test_multi_images_mixed[grpc]` has been failing on `main` for the **vLLM** engine (`Expected model to count 3 images, got: "two images"`).

### Root cause

Bisected to #1604 (e2e passed on the parent commit, failed on the merge; the test + fixtures + `mm_hashes` were unchanged across that boundary). #1604 switched the Qwen image resize to a PIL-exact **deterministic** bicubic for HF/vLLM parity. A side effect: the two **byte-identical** pug inputs (sent as base64 + URL) now produce **byte-identical pixel tensors**.

Engines then legitimately differ on byte-identical multimodal inputs:
- **vLLM** deduplicates them (encodes the duplicate once) → model sees 2 → *"two images, uploaded twice, one unique."*
- **sglang** keeps both → model sees 3 → *"three images, two are duplicates."*

So the literal "count 3" assertion is engine-dependent for duplicate inputs. (This is not a gateway bug — the gateway emits 3 distinct image regions; it's vLLM's content-level dedup, which is defensible.)

### Fix

Drop the exact-count assertion in `test_multi_images_mixed`. Keep the duplicate-detection assertion (both engines satisfy it) and the dog/pug content checks, so the test still validates multi-image + mixed base64/URL + duplicate handling without depending on engine-specific dedup of identical images.

## Test Plan

- `ruff check` + `ruff format --check` clean.
- The change makes `e2e-1gpu-chat (vllm)` `test_multi_images_mixed[grpc]` pass (it already passes on sglang); verified by this PR's GPU e2e.

<details>
<summary>Checklist</summary>

- [x] `ruff` clean
- [x] (Optional) Documentation updated — n/a (test-only)

</details>

## Summary by CodeRabbit

* **Tests**
  * Updated multimodal image test to account for engine-specific deduplication behavior.

### DIFF
--- a/e2e_test/chat_completions/test_multimodal.py
+++ b/e2e_test/chat_completions/test_multimodal.py
@@ -142,10 +142,10 @@ class TestMultimodalQwen3VL:
         assert text is not None and len(text) > 0
         text_lower = text.lower()
 
-        # Should acknowledge all 3 images
-        assert "3" in text or "three" in text_lower, (
-            f"Expected model to count 3 images, got: {text}"
-        )
+        # Don't assert an exact image count: the two pug inputs are byte-identical,
+        # and engines legitimately differ on whether identical multimodal inputs are
+        # deduplicated (vLLM encodes the duplicate once; sglang keeps both). The
+        # duplicate-detection assertion below covers the intent of this test.
         # Should identify both dog and pug
         assert any(k in text_lower for k in ["dog", "puppy", "labrador"]), (
             f"Expected dog-related content, got: {text}"


slin1237:
Repurposed this PR: the `mm_hashes` disambiguation was the wrong lever (vLLM dedups on tensor content, not caller hashes — GPU CI confirmed it didn't help). Reverted it; this PR now just relaxes the engine-dependent `count 3` assertion in `test_multi_images_mixed`. Root cause is #1604's deterministic bicubic resize making byte-identical duplicate images collapse under vLLM's content-dedup (sglang keeps both) — not a gateway bug, so the test shouldn't assert an exact count for duplicate inputs.

PR #1845, opened 2026-06-26
feat(harmony): emit parallel tool calls for gpt-oss on the Chat API

key4ng:
### Problem

gpt-oss-120b scores **0 on every parallel function-calling category** in the BFCL A/B (`parallel`, `parallel_multiple`, `live_parallel`, `live_parallel_multiple`) — on **both** arms (vllm baseline and SMG), e.g. run 28157011801.

Root cause: gpt-oss/harmony renders each tool call as a `commentary to=functions.*` message terminated by `<|call|>`, and `<|call|>` is one of harmony's `stop_tokens_for_assistant_actions` (`[<|call|>, <|return|>]`). The Chat build path forwarded that whole set as stop tokens, so generation stopped at the **first** `<|call|>` and only one tool call was ever returned. A 2-call request ("play Taylor Swift for 20m **and** Maroon 5 for 15m") returned just the first call → "Wrong number of functions" → 0.

The model *does* support parallel — it's purely the stop convention. On the H100, generating with `ignore_eos` so it runs past `<|call|>`:
```
analysis: "We need to call it twice … We'll output calls."
commentary to=functions.spotify_play {"artist":"Taylor Swift","duration":20}<|call|>
commentary to=functions.spotify_play {"artist":"Maroon 5","duration":15}<|call|>
final: ...<|return|>
```

### Solution

For the **Chat API** the *client* executes tools, so we can let the model emit all calls in one turn. Drop `<|call|>` from the Chat stop set (keep `<|return|>`) in `HarmonyBuilder::build_from_chat`. Generation then flows `analysis<|end|> -> call_1<|call|> -> call_2<|call|> -> final<|return|>`(stop), and the existing `HarmonyParserAdapter::parse_messages` already collects every `functions.*` message into the `tool_calls` vec.

The **Responses/agentic path is unchanged** (`build_from_responses` keeps `<|call|>` as a stop) because SMG executes the tools itself in a loop and must yield after each call.

This makes the SMG arm score higher than the pure-vLLM baseline on parallel categories (vLLM still stops at the first `<|call|>`).

## Changes

- `model_gateway/src/routers/grpc/harmony/builder.rs` — `build_from_chat` filters `<|call|>` (200012) out of the stop-token set; `build_from_responses` unchanged.
- Regression test `chat_drops_call_stop_token_responses_keeps_it`: Chat stop tokens exclude `<|call|>` / include `<|return|>`; Responses keep `<|call|>`.

## Test Plan

- `cargo test -p smg --lib chat_drops_call_stop_token` — passes.
- Single-call requests verified to still emit exactly one call (the model emits one call then its final).
- Live validation in progress on the H100 (gpt-oss-120b, SMG → grpc worker): confirming a parallel request now returns multiple `tool_calls`, plus a `multi_turn` regression check. Results to follow in a comment.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted chat completion stop-token handling so assistant actions no longer stop on the call token, while the return token still does.
  * Preserved the existing stop-token behavior for responses-based flows.
* **Tests**
  * Added regression coverage to verify the updated stop-token behavior across both chat completion and responses paths.

### DIFF
--- a/model_gateway/src/routers/grpc/harmony/builder.rs
+++ b/model_gateway/src/routers/grpc/harmony/builder.rs
@@ -281,12 +281,21 @@ impl HarmonyBuilder {
 
         let selection_text = self.extract_selection_text(&all_messages);
 
-        // Get stop tokens for Harmony assistant actions (<|return|> and <|call|>)
+        // Chat API: the *client* executes tools, so let the model emit ALL tool
+        // calls in one turn (parallel function calling) instead of stopping at
+        // the first <|call|>. Drop <|call|> (HARMONY_CALL_TOKEN) from the stop
+        // set, keeping <|return|>: generation then flows
+        // analysis<|end|> -> call_1<|call|> -> call_2<|call|> -> final<|return|>
+        // (stop), and `HarmonyParserAdapter::parse_messages` collects every tool
+        // call. (The Responses/agentic path keeps <|call|> as a stop because SMG
+        // executes the tools itself in a loop and must yield after each call.)
+        const HARMONY_CALL_TOKEN: u32 = 200012; // <|call|>
         let stop_token_ids: Vec<u32> = self
             .encoding
             .stop_tokens_for_assistant_actions()
             .into_iter()
             .flat_map(|set| set.into_iter())
+            .filter(|&token_id| token_id != HARMONY_CALL_TOKEN)
             .collect();
 
         Ok(HarmonyBuildOutput {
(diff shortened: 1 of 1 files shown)


key4ng:
## Investigation findings — this approach is a no-op; converting to draft

I validated this on an H100 (gpt-oss-120b, TP2: `vllm serve` baseline on :31199 vs SMG → vllm gRPC raw-token worker on :31200). **Dropping `<|call|>` from SMG's stop set has zero effect**, and the real blocker is two layers deeper. Evidence below is reproducible via `vllm /v1/completions` with a harmony-rendered prompt (`skip_special_tokens=false`).

### 1. `<|call|>` (200012) is a *model-level EOS*, re-added by vLLM regardless of what SMG sends

`generation_config.json` for gpt-oss-120b:
```json
"eos_token_id": [200002, 199999, 200012]
```
vLLM's `SamplingParams.update_from_generation_config` re-adds these to `stop_token_ids` whenever `ignore_eos=False`:
```python
if not self.ignore_eos:
    eos_ids.update(self.stop_token_ids)
    self.stop_token_ids = list(eos_ids)   # 200012 goes back in
```
Isolating test (temp=0.001, identical harmony prompt):

| request `stop_token_ids` | `ignore_eos` | result |
|---|---|---|
| `[200002]` — 200012 **omitted** (this PR's behavior) | False | **stops at `stop_reason=200012`** |
| `[200002, 200012]` (vllm default) | False | stops at `stop_reason=200012` |

Both stop at `<|call|>`. SMG omitting 200012 changes nothing — the engine puts it back. End-to-end, SMG :31200 with this patch returns exactly **1** tool call, identical to `vllm serve`.

### 2. Even if 200012 is suppressed, greedy decoding degenerates

To stop only at `<|return|>` you need `ignore_eos=True` + `stop_token_ids=[200002]`. But at BFCL's `temperature=0.001` (effectively greedy) the model then emits **one** call followed by hundreds of repeated `<|call|>` tokens — it never begins the second call:

| temp | outcome |
|---|---|
| 0.001 / 0.1 / 0.3 | degenerate — 1 call then ~444 repeated `<\|call\|>` |
| 0.6 / 1.0 | clean **2 parallel calls** → `final` message → `<\|return\|>` |

Greedy gpt-oss does **sequential** tool calling (call → await result → next call). Parallel-in-one-turn is only a *sampled* path (temp ≳ 0.6). At temp 0.6 the natural structure is exactly what we'd want:
```
analysis<|end|>
commentary to=functions.spotify_play {call 1}<|call|>
commentary to=functions.spotify_play {call 2}<|call|>
final<|message|>Both songs queued…<|return|>
```

### 3. `vllm serve` has the same limitation

vLLM's harmony chat serving (`chat_completion/serving.py`) extends every gpt-oss request's `stop_token_ids` with `get_stop_tokens_for_assistant_actions()` = `{<|return|>, <|call|>}`, so it also stops at the first `<|call|>` → also ~0 on BFCL parallel. This isn't an SMG regression; it's a property of greedy harmony.

### What would actually work

Parallel for gpt-oss needs **both**:
1. `ignore_eos=True` + `stop_token_ids=[<|return|>]` (run through all `commentary` tool-call messages until the turn's final `<|return|>`; SMG's `HarmonyParserAdapter` already collects every `functions.*` message), **and**
2. a sampling temperature ≳ 0.6 for gpt-oss tool calls (greedy degenerates).

(2) is a determinism/accuracy tradeoff that touches *all* categories (simple/multiple/multi_turn), so the safe shape is an **opt-in config flag**, default off (= today's behavior, matches `vllm serve`). I can implement that — stop change + temperature floor gated behind e.g. `--gpt-oss-parallel-sampling`, validated end-to-end on the H100 — once you confirm you want that tradeoff.

Converting to **draft**: the current diff is a no-op and should not merge.

key4ng:
Opened #1846 to revert this — the diff is a proven no-op (vLLM re-adds `<|call|>`/200012 from `generation_config` `eos_token_id` regardless of SMG's stop set; verified on H100). Left as a **draft** pending the determinism tradeoff decision for a real opt-in fix. See the investigation comment above.

PR #1848, opened 2026-06-26
feat(routing): honor X-SMG-Routing-Key on any policy (#1843)

### LINKED ISSUE #1843: [Feature]: `X-SMG-Routing-Key` consistent support in cache_aware
### Problem Statement

Sometimes we want `cache_aware` routing as a good default for models - however, we also want consumers to be able to optionally choose to have a specific sticky session (example for agentic workloads, multi-turn rollouts) to optimize their workload.

### Proposed Solution

Could SMG support, either by default or a flag, `X-SMG-Routing-Key` header (usually in `manual` policy) for other routing policies like `cache_aware`? The key would take priority over any other routing decisions if it's provided and there's a match.

### Alternatives Considered

Currently we're selecting one or the other, per-model policy (`manual` or `cache_aware`) and default to `cache_aware`. It would be nice to not have to choose ahead-of-time for each model.

### Feature Area

Routing & Load Balancing

### Affected Component(s)

_No response_

### Use Case

RL sampler deployments and agentic workloads

### Priority to You

Nice to have

### Contribution

- [x] I am willing to contribute this feature (with guidance)
- [x] I am willing to help test this feature
- [x] I can provide more detailed requirements if needed

### Additional Context

_No response_

slin1237:
Closes #1843.

## What
Opt-in per-request **sticky routing**: when a request carries `X-SMG-Routing-Key`, it is pinned to a worker using manual sticky-map semantics, overriding the model's configured policy (e.g. `cache_aware`). Without the header, the configured policy decides — unchanged behavior. Off by default.

This lets operators keep `cache_aware` as the default while consumers (RL samplers, agentic multi-turn rollouts) opt into session affinity per request, instead of committing each model to `manual` vs `cache_aware` ahead of time.

## How
- **`RoutingKeyOverride` decorator** wraps the model's policy plus an internal `ManualPolicy`; key present → sticky map, absent → inner policy. Non-selection trait methods delegate to the inner policy so a wrapped load-aware policy still works.
- **`WorkerLeg`** (`Single`/`Prefill`/`Decode`) added to `SelectWorkerInfo`; `ManualPolicy` namespaces its sticky entries per leg so PD prefill/decode stick independently (`Single` is byte-identical to before). The gRPC PD path tags each leg.
- **`PolicyRegistry::with_override`** wraps key-ignoring policies (`cache_aware`, `least_load`, `power_of_two`, `round_robin`, `random`, `bucket`) and skips key-native ones (`manual`, `consistent_hashing`, `prefix_hash`, `passthrough`). The decorator is made transparent to the registry's `CacheAwarePolicy` monitor/load-receiver injection and load-aware detection (the one non-obvious part).
- Enabled via `--routing-key-override` (CLI), `routing_key_override=True` (Python), or `routing_key_override: { enabled: true }` (config). Reuses the existing manual eviction/idle/assignment knobs for the sticky map.

## Testing
- Unit tests: `WorkerLeg` default, `ManualPolicy` leg namespacing, `RoutingKeyOverride` (key→sticky / no-key→inner / per-leg independence), registry wrap/skip/disabled, and a load-aware **transparency** test (a wrapped `power_of_two` is still detected + injected).
- `cargo build` / `cargo +nightly fmt` / `cargo clippy --all-targets` clean across `smg`, `smg-python`, `smg-golang`.
- An e2e sticky-routing test is a planned follow-up (needs a multi-worker lane).

## Scope (v1)
Routing-key only (not `X-SMG-Target-Worker`), a global flag (not per-model), manual semantics (no consistent-hash strategy option) — all additive later.

## Summary by CodeRabbit

* **New Features**
  * Added an optional routing-key override for sticky routing, allowing requests with `X-SMG-Routing-Key` to stay consistent across supported routing policies.
  * Introduced per-request worker-leg routing so prefill, decode, and single-request paths are handled distinctly.

* **Bug Fixes**
  * Improved sticky routing behavior so identical routing keys no longer collide between prefill and decode paths.
  * Kept default routing behavior unchanged when the override is not enabled.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -368,6 +368,7 @@ struct Router {
     host: String,
     port: u16,
     health_check_port: Option<u16>,
+    routing_key_override: bool,
     worker_urls: Vec<String>,
     policy: PolicyType,
     worker_startup_timeout_secs: u64,
@@ -504,6 +505,19 @@ impl Router {
         })
     }
 
+    fn parse_assignment_mode(&self) -> Result<config::ManualAssignmentMode, config::ConfigError> {
+        match self.assignment_mode.as_str() {
+            "random" => Ok(config::ManualAssignmentMode::Random),
+            "min_load" => Ok(config::ManualAssignmentMode::MinLoad),
+            "min_group" => Ok(config::ManualAssignmentMode::MinGroup),
+            other => Err(config::ConfigError::InvalidValue {
+                field: "assignment_mode".to_string(),
+                value: other.to_string(),
+                reason: "expected 'random', 'min_load', or 'min_group'".to_string(),
+            }),
+        }
+    }
+
     pub fn to_router_config(&self) -> config::ConfigResult<config::RouterConfig> {
         use config::{
             DiscoveryConfig, MetricsConfig, PolicyConfig as ConfigPolicyConfig, RoutingMode,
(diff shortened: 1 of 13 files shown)


ekzhang (approved):
🥳

PR #1877, opened 2026-07-04
fix(grpc): router-authoritative string stop sequences for SGLang skip_tokenizer_init workers

gongwei-130:
## Summary

SGLang gRPC workers run with `skip_tokenizer_init=True` (the router owns the tokenizer). The router forwards string `stop` sequences as-is into the gRPC `SamplingParams`, and SGLang's `SamplingParams.verify()` **rejects** them in that mode:

```
stop=['.'] is unavailable when skip_tokenizer_init=True
(requires tokenizer to decode tokens to text for matching).
```

So any request carrying the OpenAI-compatible `stop` parameter against an SGLang gRPC worker gets a **400**. This PR makes the **router authoritative for string stops** end to end: the 400 is gone, output text is trimmed correctly, `finish_reason` / `matched_stop` / Anthropic `stop_reason`+`stop_sequence` match what vLLM reports for the same request, and streaming requests **abort the worker** when the router matches a stop — so multi-token stops no longer over-generate to `max_tokens` (no inflated `usage.completion_tokens`, no wasted GPU time).

SGLang-specific: the vLLM servicer forces `detokenize=bool(stop)`, TRT-LLM tokenizes stop words server-side, MLX has no string-`stop` field.

## Design

The router already decodes worker output through a `StopSequenceDecoder` per request. This PR completes that into full ownership of string stops:

**1. Drain string stops from SGLang requests** (`resolve_sglang_string_stops` in `common/stages/helpers.rs`), called from the chat / completions / messages request-building stages — exactly the paths whose streaming *and* non-streaming handlers both run a trimming decoder. As an optimization, a stop string that encodes to a **single token** is forwarded as a `stop_token_ids` entry so the worker halts immediately (skipped when `ignore_eos=true`, where SGLang skips all `stop_token_ids` matching — see `Req._check_token_based_finish`).

**2. Record what matched** — `StopSequenceDecoder` now tracks `MatchedStop::Sequence(String) | TokenId(u32)` when it stops.

**3. Refine response metadata** (`utils::refine_stop_metadata`, applied at all six response sites: chat/completions/messages × streaming/non-streaming):
   - decoder matched a string stop → `finish_reason="stop"`, `matched_stop` = the original stop string (was: `"length"` + null);
   - worker halted on a *converted* single-token stop and reported a numeric token id → mapped back to the user's stop string;
   - EOS and user-provided `stop_token_ids` pass through unchanged.
   Also gates the streaming Messages `stop_reason` on an actual string match (a numeric matched_stop is `end_turn`, not `stop_sequence` with a null sequence), and removes the completions endpoint's stream-vs-non-stream `finish_reason` divergence.

**4. Streaming early termination** — when the router decoder matches a string stop on an SGLang stream (n=1), the loop records usage from the chunk's cumulative counters, emits the final chunks, and breaks **without** `mark_completed()`; dropping the stream sends the Abort RPC (`AbortOnDropStream`). Gated to SGLang streams and actual `Sequence` matches — vLLM/TRT-LLM paths are untouched.

### Deliberately not drained (documented `NOTE`s at the call sites)

- **Harmony (gpt-oss)**: no router-side decoder exists there, and injecting a user stop token can truncate before `<|return|>`/`<|call|>`, corrupting Harmony parsing. Keeps the pre-existing 400.
- **Native `/generate`**: its streaming path has no decoder, so a cleared stop would stream untrimmed text. Keeps the pre-existing 400.

### Known remaining limitations (in the helper docstring)

- Non-streaming multi-token stops: `finish_reason`/`matched_stop` are correct, but the worker still generates to EOS/`max_tokens` before the response is assembled. Needs early abort in `response_collection` — follow-up.
- SentencePiece leading-space tokenization can keep the single-token optimization from firing mid-sentence; correctness holds via the decoder.

## Tests

- `resolve_sglang_string_stops`: 12 unit tests (single/multi-token, mixed, dedup, empty/unknown, whitespace single-token via a wrapper tokenizer, two distinct single tokens, `ignore_eos`, missing `sampling_params`, no tokenizer, non-SGLang untouched)
- `refine_stop_metadata`: 5 unit tests
- `StopSequenceDecoder::matched_stop`: 2 unit tests
- Full `smg` lib suite green; clippy/fmt clean

## End-to-end verification (B200, real SGLang gRPC worker)

Same host, same engine image and weights, only the router build swapped:

- `stop=["."]` (non-stream) → 200, `finish_reason="stop"`, `matched_stop="."` (the string, not a token id), text trimmed at the stop
- multi-token stop (non-stream) → `finish_reason="stop"`, `matched_stop=<stop string>`, text trimmed exactly before the stop
- multi-token stop (**streaming**) → stream terminates at the match with `usage.completion_tokens=3` vs **488** for the identical request non-streamed (worker log shows `Receive abort request: ...`), confirming the Abort path works

This is the same change merged in our production deployment after review and e2e validation.

### DIFF
--- a/model_gateway/src/routers/grpc/common/stages/helpers.rs
+++ b/model_gateway/src/routers/grpc/common/stages/helpers.rs
@@ -2,6 +2,7 @@
 
 use std::sync::Arc;
 
+use llm_tokenizer::traits::Tokenizer;
 use rand::RngExt;
 use smg_grpc_client::{
     mlx_proto,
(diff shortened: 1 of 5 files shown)


slin1237 (commented):
it doesn't look like the fix should be here? 
And even if we wanna add a code like this for SGLang, the code shouldn't be in helper
Also. Please combine all the if statement to single if. 
And follow pull request template.

gongwei-130:
Closing — superseded by #2057 (`feat(zmq): forward EOS and resolve string stops for the direct backend`), later refactored by #2068.

`resolve_string_stops()` in `model_gateway/src/routers/grpc/common/stages/helpers.rs` now does what this PR proposed, and generalises it:

```rust
ProtoGenerateRequest::Sglang(req) => {
    if let Some(params) = req.sampling_params.as_mut() {
        let stops = std::mem::take(&mut params.stop);
        encode_single_token_stops(stops.clone(), &mut params.stop_token_ids, tokenizer);
        return stops;
    }
}
```

Same approach — drop the string `stop` list that a `skip_tokenizer_init=True` worker rejects, forward single-token stops as `stop_token_ids` for early stopping, and leave multi-token stops to the router-side `StopSequenceDecoder`. Two improvements over what I had here:

1. It covers every token-only wire, not just SGLang gRPC — the direct-ZMQ backends (vLLM EngineCore, TokenSpeed) receive token ids only and cannot match a stop string either. That path did not exist when this PR was opened.
2. It returns the stripped stop strings, making the router's residual obligation explicit: response processing must trim them from the output text. This PR left that implicit in the decoder.

Nothing here is worth carrying forward on top of that. Thanks to whoever picked the approach up.

PR #1882, opened 2026-07-07
refactor(grpc/chat): unify thinking resolution by layer

slin1237:
### Problem

Follow-up to #1876. That fix made `reasoning_effort: "none"/"minimal"` disable thinking, but resolved it **entirely in the gateway** with a blind dual-key write (`enable_thinking` *and* `thinking`), duplicated in `message_utils.rs` (Anthropic `ThinkingConfig`). Two smells surfaced in review of #1876:

1. The **write** path guessed both key names, while the **read** path (`extract_thinking_from_kwargs`) precisely used the template's *one* key — the PR contradicted its own principle.
2. The `none/minimal` rule lived in two places (`chat_utils` and `parsers`) that could drift and desync template-render vs. parser start-state.

It also left the Go binding ignoring `reasoning_effort` entirely.

### Solution

Give thinking-resolution **one owner per concern**, keyed off tokenizer metadata that already exists (`thinking_key_name` / `thinking_toggle`):

- **Interpret** (OpenAI `reasoning_effort` / Anthropic `ThinkingConfig` → `Option<bool>`) at the protocol layer.
- **Apply** (set the model's own toggle key) inside the tokenizer.
- The gateway just composes the two.

## Changes

- **`openai-protocol`** — `thinking_from_reasoning_effort()`: model-agnostic interpretation (`none`/`minimal` → off); the `minimal` decision is documented inline (closest "off" for boolean-toggle models).
- **`llm-tokenizer`** — `ChatTemplateParams.thinking: Option<bool>` + `ThinkingKeyName::as_kwarg()`. `ChatTemplateState::apply` injects the template's **own** toggle key (single correct key, not both) as a default that an explicit `template_kwargs` entry overrides. DeepSeek V3.2/V4 native encoders bypass `apply`, so `derive_thinking_mode` consumes the same preference (explicit kwarg still wins).
- **gateway** — `chat_utils.rs` and `message_utils.rs` drop the dual-key writes and pass the resolved `thinking`; `parsers.rs` delegates the mapping to the protocol fn; `extract_thinking_from_kwargs` demoted to `pub(crate)`.
- **`smg-golang`** — migrated to `resolve_user_thinking` so the Go path honors `reasoning_effort`.

## Test Plan

Behavior is unchanged for target models (GLM/Qwen3, DeepSeek V3.1/V3.2/V4, Kimi). Precedence preserved everywhere: explicit `chat_template_kwargs` toggle > `reasoning_effort`/`ThinkingConfig` mapping > template default.

- `cargo test -p openai-protocol` (145) and `cargo test -p llm-tokenizer --lib` (145) pass — including new tests for the mapping, the `apply` injection + explicit-override, and the DeepSeek `derive_thinking_mode` fallback.
- Gateway `grpc::utils` unit tests pass (21).
- `cargo +nightly fmt` clean; `clippy -D warnings` clean (locally on default features — `--all-features` needs the OpenCV system dep and runs on CI).
- End-to-end covered by existing `e2e_test/chat_completions/test_enable_thinking.py`.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (CI; default-features locally)

</details>

## Summary by CodeRabbit

* **New Features**
  * Enhanced “thinking”/reasoning preference handling in chat rendering, including mapping `reasoning_effort` (e.g., `none`, `minimal`) and projecting it into template rendering.
* **Bug Fixes**
  * Fixed inconsistent interpretation of reasoning settings by ensuring clear precedence between explicit template overrides and request-level values.
  * Improved fallback behavior when `reasoning_effort` is missing or unrecognized, and aligned provider behavior with the unified “thinking” preference.
* **Tests**
  * Added/updated unit tests for `reasoning_effort` mapping, precedence rules, and tokenizer override behavior.

### DIFF
--- a/bindings/golang/src/utils.rs
+++ b/bindings/golang/src/utils.rs
@@ -2,7 +2,7 @@
 
 use llm_tokenizer::traits::Tokenizer;
 use openai_protocol::chat::ChatCompletionRequest;
-use smg::routers::grpc::utils::{extract_thinking_from_kwargs, should_mark_reasoning_started};
+use smg::routers::grpc::utils::{resolve_user_thinking, should_mark_reasoning_started};
 use uuid::Uuid;
 
 /// Helper function to generate tool call ID (matches router implementation)
@@ -32,7 +32,11 @@ pub(crate) fn chat_requires_reasoning(
     tokenizer: &dyn Tokenizer,
 ) -> bool {
     should_mark_reasoning_started(
-        extract_thinking_from_kwargs(request.chat_template_kwargs.as_ref(), tokenizer),
+        resolve_user_thinking(
+            request.chat_template_kwargs.as_ref(),
+            request.reasoning_effort.as_deref(),
+            tokenizer,
+        ),
         tokenizer,
     )
 }
--- a/crates/protocols/src/chat.rs
+++ b/crates/protocols/src/chat.rs
@@ -320,6 +320,25 @@ pub struct ChatCompletionRequest {
     pub other: Map<String, Value>,
 }
 
+/// Map an OpenAI `reasoning_effort` to a thinking on/off preference.
+///
+/// This is the protocol-level interpretation of "does the caller want
+/// reasoning?" — independent of any model/template. `reasoning_effort` is a
+/// *level* (`"low"`/`"medium"`/`"high"`) plus the vendor-extension `"none"`.
+///
+/// Both `"none"` and `"minimal"` map to thinking OFF (`Some(false)`).
+/// `"minimal"` is treated as an off-signal deliberately: templates that expose
+/// only a boolean thinking toggle (GLM/Qwen3) cannot do "a little" reasoning,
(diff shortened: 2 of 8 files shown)


slin1237:
Addressed the review feedback in 95f65d6f:

- **@chatgpt-codex-connector (P2 — DeepSeek thinking):** Good catch, this was a real regression. DeepSeek V3.2/V4 use native encoders (`apply_deepseek_v32/v4`) that bypass `ChatTemplateState::apply` and read `template_kwargs["thinking"]` via `derive_thinking_mode`, so the refactor silently dropped the preference for those models. `derive_thinking_mode` now falls back to `params.thinking` (an explicit `template_kwargs["thinking"]` still wins — same precedence as the Jinja path). Added a regression test covering all three cases.
- **@gemini-code-assist (clone nit):** Applied — `ChatTemplateState::apply` now checks for the key first and skips cloning `template_kwargs` entirely when the toggle is already set explicitly.

PR #1890, opened 2026-07-08
refactor(multimodal): split grpc multimodal.rs into a module directory

slin1237:
### Problem

`model_gateway/src/routers/grpc/multimodal.rs` had grown to ~2.65k lines — the
largest flat file under `routers/grpc` (its neighbors `harmony/` and `regular/`
are already module directories). Everything from content detection to tensor
serialization to SHM transport resolution lived in one file, and the module's
real public surface was buried among ~60 functions with ad-hoc visibility.

### Solution

Split `multimodal.rs` into a `multimodal/` module directory along the pipeline's
own phase boundaries. **This is a pure move — no behavior change.**

| Module | Responsibility |
|--------|----------------|
| `detect` | modality detection + content extraction (chat + messages) |
| `config` | model config-file registry + per-router component bundle |
| `process` | fetch → preprocess → expand placeholder tokens → build intermediate |
| `assemble` | intermediate → backend-specific `MultimodalData` + per-item split |
| `serialize` | tensor byte/dtype serialization (f32 → bf16/f16) |
| `transport` | SHM-vs-inline resolution + `/dev/shm` namespace verification |

The crate-facing API is **unchanged**: the same symbols are re-exported from
`mod.rs`, so every `multimodal::X` call site keeps working. Internal helpers are
now enforced-private (`pub(super)`) instead of implicitly module-scoped, so the
re-export block in `mod.rs` is now the explicit module contract.

## Changes

- `git mv multimodal.rs multimodal/mod.rs`; `mod.rs` now holds the module docs,
  the shared intermediate types (`MultimodalOutput`, `MultimodalIntermediate`,
  `PrecomputedMultimodalIntermediate`), and the `pub(crate)` re-export surface.
- Extracted the six submodules above. Tests move next to the code they cover.
- Net delta +137 lines (per-file imports + module docs); no logic changed.

**Reviewing:** view with `git diff --color-moved=zebra` — the vast majority is
verbatim moves. The only non-move edits are per-file `use` blocks and
visibility bumps (private → `pub(super)`/`pub(crate)`).

## Test Plan

- `cargo check -p smg --lib --tests` — clean
- `cargo clippy -p smg --lib --tests -- -D warnings` — clean
- `cargo +nightly fmt` — clean (only the new files touched)
- `cargo test -p smg --lib multimodal` — **24 passed, 0 failed** (the same
  tests as before, now under `assemble/detect/config/process/serialize/transport`)

No functional or config surface changed, so no bindings/e2e impact.

## Checklist

- [x] `cargo +nightly fmt` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] Tests pass
- [x] No behavior change (pure refactor)
- [x] No config/protocol/bindings surface touched

## Summary by CodeRabbit

* **New Features**
  * Updated multimodal gRPC handling with image/video detection for chat and Messages API inputs.
  * Added a multimodal request pipeline that preprocesses media, expands multimodal placeholders, and assembles backend-specific payloads.
  * Improved multimodal encoder data transfer, including SHM vs inline selection and efficient tensor serialization with float16/bfloat16 support.
* **Bug Fixes**
  * Enhanced multimodal model/preprocessor configuration loading with resilient fallbacks and cached reuse.
  * Added safer SHM lifecycle handling and cleanup during multimodal assembly.


slin1237:
Thanks for the reviews. Dispositions:

- **`log_mm_timing_enabled` env re-read** — fixed in 175f72b0 (cached via `OnceLock`, read once instead of per request).
- **`expand_tokens` silent excess placeholders** — fixed in 175f72b0 (added a warning + regression test; token output is unchanged).
- **`get_or_load` single-flight** — valid, but it's a real async-coordination change (per-key single-flight + concurrency tests) and out of scope for a pure-move refactor. The registry is preloaded at tokenizer registration so the cold-miss race is rare; tracking it as a dedicated follow-up.
- **transport SHM "too broad"** — this looks like a misread of the EPD data flow: the SHM-backed `encoder_input` *and* `model_specific_tensors` are dispatched to the **encode** workers, while the prefill/decode legs call `clear_mm_pixel_values()` and receive the embedding over Mooncake. So gating `shm_enabled` on `encode_assignments` sharing the gateway's `/dev/shm` is correct — there is no second destination that would need a split flag. No change.
- **Gemini doc-link paths** — the intra-doc links resolve correctly and the repo has no absolute-path convention (existing code mixes `crate::`, `super::`, and bare `[Type]`); keeping the concise relative form for a module listing its own submodules.

For context: this PR is a pure module split (commit 1) plus the two small review fixes above (commit 2), kept separate so the move stays reviewable as a no-op via `git diff --color-moved`.

PR #1892, opened 2026-07-08
feat(multimodal): configurable engine-agnostic tensor transport

slin1237:
### Problem

Multimodal tensor transport (inline vs `/dev/shm`) was tunable only via
`SMG_TOKENSPEED_MM_*` environment variables, TokenSpeed-specific and env-only.
There was no CLI/YAML config, no per-worker control (co-located vs remote
workers can't differ), and the knob was framed as TokenSpeed-only even though
the transport is engine-agnostic.

### Solution

Promote it to a first-class, engine-agnostic config surface with a clear
resolution precedence:

> per-worker `WorkerSpec` override → router config / CLI → `SMG_MM_*` env
> (legacy `SMG_TOKENSPEED_MM_*` aliases) → built-in default (`inline`, 64 KiB)

Router-level defaults are resolved once at startup (`server::startup`); the
per-request mode drives `shm_enabled`, and the resolved size threshold rides on
`TokenSpeedMultimodalData` so both write paths (encoder input in `serialize.rs`
and model-specific tensors in `proto_wrapper.rs`) honor it consistently.

## Changes

- **`openai-protocol`**: new `TransportMode` enum (`inline|shm|auto`), serde +
  `JsonSchema` + `FromStr`. `WorkerSpec` gains `multimodal_tensor_transport` +
  `multimodal_shm_min_bytes` overrides.
- **`RouterConfig`** (`config/types.rs`, `builder.rs`): the two fields + builder
  methods + `Default`.
- **CLI** (`main.rs`): `--multimodal-tensor-transport` (value_parser
  `inline|shm|auto`) + `--multimodal-shm-min-bytes`; wired into `to_router_config`.
- **`transport.rs`**: resolvers renamed `tokenspeed_* → mm_*`; precedence logic;
  per-worker override readers; startup seed `init_mm_transport_defaults`.
- **Python SDK**: `RouterArgs` dataclass + argparse + `PyRouterConfig`
  (`bindings/python`). (Go SDK is a request client — no router-config surface.)
- **Docs**: generalized the "Multimodal Tensor Transport" reference section.

### Behavior change (explicit `shm` only)

`shm` now **forces** SHM whenever SMG can write `/dev/shm` (operator asserts
co-location). Previously `shm` was identical to `auto` (both required verified
worker sharing), so `shm` was redundant. `auto` is unchanged. Default is
`inline`, so existing deployments are unaffected.

## Test Plan

- `cargo clippy -p smg --all-targets -- -D warnings` — clean (lib + tests + main.rs)
- `cargo check -p smg-python` — clean (Python binding compiles)
- `cargo +nightly fmt` — clean (scoped to the change set)
- `cargo test -p openai-protocol --lib` — 77 pass
- `cargo test -p smg --lib` — 1146 pass (one unrelated `middleware::metrics`
  interner test is a known parallel-test flake; passes in isolation)
- `cargo test -p smg --bins` — new two-path guard
  `multimodal_transport_flows_into_both_configs` passes
- `python3 -m py_compile` on `router_args.py` — OK

## Checklist

- [x] `cargo +nightly fmt` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] Both config paths covered (`to_router_config`; `to_server_config` wraps it) + two-path test
- [x] CLI `value_parser` validation
- [x] `Default` + `#[serde(default, skip_serializing_if)]`
- [x] Python bindings (RouterArgs + PyRouterConfig) updated
- [x] Docs updated

## Summary by CodeRabbit

* **New Features**
  * Added configurable multimodal tensor transport (mode: `inline`/`shm`/`auto`) and `multimodal_shm_min_bytes` across router config, CLI, Python bindings, and per-worker overrides.
  * Added validation with user-facing errors for unsupported transport mode values.
* **Bug Fixes**
  * Improved inline vs SHM decision-making to consistently apply the configured SHM minimum-byte threshold during multimodal tensor handling.
  * Preserved safe fallback behavior when shared memory is unavailable or not writable.
* **Documentation**
  * Updated configuration docs with new precedence rules, new CLI flags, and migrated `SMG_MM_*` environment variables (legacy `SMG_TOKENSPEED_MM_*` aliases retained).
* **Tests**
  * Added coverage to verify the new flags propagate into both router and server configuration outputs.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -389,8 +389,6 @@ struct Router {
     assignment_mode: String,
     max_payload_size: usize,
     dp_aware: bool,
-    multimodal_tensor_transport: Option<String>,
-    multimodal_shm_min_bytes: Option<usize>,
     dp_minimum_tokens_scheduler: bool,
     api_key: Option<String>,
     log_dir: Option<String>,
@@ -486,6 +484,8 @@ struct Router {
     epd_disaggregation: bool,
     encode_urls: Option<Vec<(String, Option<u16>)>>,
     encode_policy: Option<PolicyType>,
+    multimodal_tensor_transport: Option<String>,
+    multimodal_shm_min_bytes: Option<usize>,
 }
 
 impl Router {
@@ -528,6 +528,23 @@ impl Router {
         use config::{
             DiscoveryConfig, MetricsConfig, PolicyConfig as ConfigPolicyConfig, RoutingMode,
         };
+
+        // Validate the transport mode up front. The CLI (value_parser) and the
+        // argparse path (choices) already reject bad values; this covers direct
+        // programmatic `RouterArgs` use, matching the CLI/Rust parsing contract.
+        let multimodal_tensor_transport = self
+            .multimodal_tensor_transport
+            .as_deref()
+            .map(|value| {
+                config::TransportMode::parse(value).ok_or_else(|| {
+                    config::ConfigError::InvalidValue {
+                        field: "multimodal_tensor_transport".to_string(),
+                        value: value.to_string(),
+                        reason: "expected 'inline', 'shm', or 'auto'".to_string(),
+                    }
+                })
+            })
+            .transpose()?;
 
         let convert_policy = |policy: &PolicyType| -> config::ConfigResult<ConfigPolicyConfig> {
             Ok(match policy {
@@ -787,11 +804,7 @@ impl Router {
             .maybe_storage_hook_wasm_path(self.storage_hook_wasm_path.as_deref())
             .enable_wasm(self.enable_wasm)
             .dp_aware(self.dp_aware)
-            .multimodal_tensor_transport(
-                self.multimodal_tensor_transport
-                    .as_deref()
-                    .and_then(config::TransportMode::parse),
-            )
+            .multimodal_tensor_transport(multimodal_tensor_transport)
             .multimodal_shm_min_bytes(self.multimodal_shm_min_bytes)
             .routing_key_override(config::RoutingKeyOverrideConfig {
                 enabled: self.routing_key_override,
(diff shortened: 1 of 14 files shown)


slin1237:
Thanks for the reviews. Addressed in 145613e6:

- **Invalid `multimodal_tensor_transport` silently defaulting** (claude, gemini, coderabbit) — fixed: direct programmatic `RouterArgs` use now returns `ConfigError::InvalidValue` on a bad mode string, matching the CLI (`value_parser`) and Rust parsing contract instead of falling back to `inline`.
- **`primary_worker` uses prefill, not the encode worker, in EPD** (gemini, coderabbit) — fixed: it now returns the first encode worker when `encode_assignments` is present, so per-worker overrides on the encode spec (transport mode / shm-min-bytes / encoder dtype) are honored where the tensors actually land. Plain PD still falls back to prefill.
- **`_Router` constructor params inserted mid-list** (coderabbit) — fixed: moved `multimodal_tensor_transport` / `multimodal_shm_min_bytes` to the end of the struct, `#[pyo3(signature)]`, `new()` params, and body, per the file's documented positional-compatibility convention. (The Python `RouterArgs` dataclass is kwargs-only, so its field order is left grouped.)
- **Missing `shm_min_bytes` boundary coverage** (coderabbit) — added `shm_min_bytes_threshold_gates_inline_vs_shm`: inline just below the threshold, SHM at/above.

Also added (independent of the review): a one-time deprecation warning when a legacy `SMG_TOKENSPEED_MM_*` env var is read — the aliases stay for backward compatibility, the warning nudges migration to `SMG_MM_*` so they can be dropped in a future major.

slin1237:
Addressed in d2ef4bb1: `mm_shm_min_bytes_from_env` now logs a one-time warning on an unparseable `SMG_MM_SHM_MIN_BYTES` (or legacy alias) value before falling back to the default, matching `mm_tensor_transport_mode_from_env`'s behavior. Good catch — thanks.

PR #1893, opened 2026-07-08
feat(multimodal): vLLM SHM tensor transport

slin1237:
### Problem

The `/dev/shm` tensor transport (added for TokenSpeed) and the engine-neutral
`ShmHandle` (hoisted to `common.proto`) + configurable transport layer are all
in place, but vLLM still only accepts multimodal tensors **inline** in the gRPC
message. Large image tensors to a co-located vLLM worker pay a full gRPC copy.

### Solution

Bring the same-host SHM transport to vLLM, reusing the common `ShmHandle` and
the `--multimodal-tensor-transport` config. Default is `inline`, so existing
deployments are unaffected.

## Changes

**Proto** (`vllm_engine.proto`, already imports `common.proto`):
- `TensorData`: `bytes data = 1` → `oneof payload { bytes inline = 1;
  smg.grpc.common.ShmHandle shm = 4; smg.grpc.common.RemoteTensorHandle
  remote = 5; }`. **Field 1 stays `inline` for wire compatibility.**
- `GetServerInfoResponse`: `+ shm_namespace_id = 10` (so the router can verify a
  shared `/dev/shm` under `auto`).

**Gateway (Rust)**:
- Extracted an engine-neutral `resolve_mm_tensor_payload` (the inline-vs-SHM
  decision + metrics); both `tokenspeed_tensor_payload` and the new
  `vllm_tensor_payload` map its result onto their own oneof.
- `VllmMultimodalData` carries resolved `shm_enabled`/`shm_min_bytes`;
  `assemble_vllm` resolves them (transport config → worker override); `into_proto`
  emits the oneof payload.
- `collect_vllm_*_shm_handles` + the vLLM client cleans up SHM on send failure
  (mirrors TokenSpeed). Renamed the engine-neutral `cleanup_tokenspeed_shm_handles`
  → `cleanup_mm_shm_handles`.

**Servicer (Python)**:
- New shared `mm_shm` module (reads a `TensorData` payload inline-or-SHM,
  validates SHM names, unlinks after read, computes the `/dev/shm` namespace id).
- The vLLM servicer reads tensors through `mm_shm` (the `oneof` renames `.data`
  → `.inline`, so this is required) and advertises `shm_namespace_id`.

## Test Plan

- `cargo clippy -p smg --all-targets -- -D warnings` — clean
- `cargo test -p smg --lib` — 1147 pass (one unrelated `middleware::metrics`
  interner test is a known parallel-test flake; passes in isolation)
- `cargo +nightly fmt` — clean
- **Python proto regen (grpc_tools)** — verified `vllm_engine_pb2.TensorData`
  exposes the `inline|shm|remote` oneof (`shm` → `common_pb2.ShmHandle`) and
  `GetServerInfoResponse.shm_namespace_id` exists
- `python -m py_compile` on `mm_shm.py` + the vLLM servicer — OK

e2e for the SHM path (co-located worker, assert SHM is taken) is deferred to the
multimodal coverage PR.

## Follow-ups

- Migrate the TokenSpeed servicer onto the shared `mm_shm` module (mechanical
  dedup; kept out of this PR to avoid churning the critical servicer).
- vLLM **video** (`is_video` → `pixel_values_videos`) — PR #3b.

## Checklist

- [x] `cargo +nightly fmt` + `clippy -- -D warnings` clean
- [x] Tests pass (Rust) + Python proto regen / py_compile verified
- [x] Wire-compatible proto (`inline` keeps field 1)
- [x] SHM cleanup on failure (no `/dev/shm` leak)

## Summary by CodeRabbit

* **New Features**
  * Added shared-memory transport for vLLM multimodal tensor payloads, with automatic fallback to inline when SHM is unavailable or below size thresholds.
  * Extended server information with a shared-memory namespace identifier to improve compatibility checks between workers and clients.
* **Bug Fixes**
  * Improved multimodal shared-memory cleanup across both success and error paths to reduce leaked segments.
  * Hardened shared-memory payload reading with stricter validation and safer file access.
  * Fixed vLLM sequential decode reuse to avoid carrying over vision inputs.

### DIFF
--- a/crates/grpc_client/proto/vllm_engine.proto
+++ b/crates/grpc_client/proto/vllm_engine.proto
@@ -95,11 +95,22 @@ message TokenizedInput {
   repeated uint32 input_ids = 2;  // Actual token IDs to process
 }
 
-// A typed tensor: raw little-endian bytes + shape + dtype.
+// A typed tensor descriptor. The raw bytes live in exactly one payload
+// transport; shape and dtype describe how the receiver interprets them.
 message TensorData {
-  bytes data = 1;              // Raw little-endian bytes (f32/i64/u32)
   repeated uint32 shape = 2;   // Dimension sizes
   string dtype = 3;            // "float32", "int64", "uint32"
+
+  oneof payload {
+    // Current path: raw little-endian bytes carried in the gRPC message
+    // (field 1, formerly `bytes data`, kept for wire compatibility).
+    bytes inline = 1;
+    // Same-host CPU shared memory path; preferred for large payloads when
+    // SMG and the vLLM worker share /dev/shm.
+    smg.grpc.common.ShmHandle shm = 4;
+    // Cross-node or non-shared-memory transport (e.g. NIXL). Not implemented.
+    smg.grpc.common.RemoteTensorHandle remote = 5;
+  }
 }
 
 message PlaceholderRange {
@@ -339,6 +350,11 @@ message GetServerInfoResponse {
   string kv_engine_id = 8;  // kv_transfer_config.engine_id, "" if not configured
 
   int32 data_parallel_size = 9;  // parallel_config.data_parallel_size (1 when DP is off)
+
+  // This worker's /dev/shm tmpfs identity (<boot_id>:<st_dev>), advertised so
+  // the router can verify a shared /dev/shm before using the SHM tensor
+  // transport under `auto`. Empty when it can't be determined.
+  string shm_namespace_id = 10;
 }
 
 // =====================
--- a/grpc_servicer/smg_grpc_servicer/mm_shm.py
+++ b/grpc_servicer/smg_grpc_servicer/mm_shm.py
@@ -8,6 +8,7 @@ oneof shape (``inline`` | ``shm`` | ``remote``) via ``common.proto``.
 """
 
 import os
+import stat
 
 # Unlink each /dev/shm segment right after the worker reads it (default on) so
 # same-host SHM tensors don't accumulate. Disable with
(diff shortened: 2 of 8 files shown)


slin1237:
Thanks all — addressed in 0a7d4ade:

- **P1: sequential vLLM PD reuses the SHM handle across prefill+decode** (codex) — fixed. The decode leg now clears its multimodal inputs (mirrors the parallel PD path); it doesn't run the vision encoder, and with SHM the prefill leg already read+unlinked the segments. Confirmed the parallel path already did this (`clear_mm_pixel_values`), so decode provably doesn't need the tensors.
- **Build-failure SHM leak** (claude, coderabbit, codex) — fixed. Added `finish_vllm_request` (mirrors `finish_tokenspeed_request`): collects the SHM handles before the build and unlinks them if sampling/tool-constraint validation fails, so a bad request no longer leaks `/dev/shm` files. Both the chat and messages vLLM build arms use it. The send-failure path was already covered.
- **SHM symlink hardening** (coderabbit) — fixed. `mm_shm` opens with `O_NOFOLLOW` and requires a regular file (fail closed), so a crafted name resolving to a pre-existing `/dev/shm` symlink can't redirect the read.
- **Cache `shm_namespace_id`** (gemini) — fixed (module-level cache; static per process).
- **`log_..._timing` env re-read** (gemini) — fixed (`OnceLock` cache).
- **Pre-commit ruff import order** — fixed (`mm_shm` is first-party).

Declined / deferred:
- **`bytearray` → zero-copy `frombuffer`** (gemini) — kept `bytearray` to preserve the existing writable-tensor behavior; passing `bytes` yields a read-only tensor, a behavior change I'd rather validate separately (perf follow-up).
- **`cleanup_mm_shm_handles` internals / `TOKENSPEED_UNLINK_MM_SHM_AFTER_READ` still say "TokenSpeed"** (claude nits) — deferred to the TokenSpeed-servicer migration follow-up, where the shared SHM naming (`smg-tokenspeed-` prefix, env var, log strings) gets made engine-neutral together.

Verified: `cargo clippy --all-targets -D warnings` clean, 1147 lib tests pass, `ruff check`/`ruff format` clean, `py_compile` OK.

PR #1895, opened 2026-07-09
feat(multimodal): vLLM video via shared Modality enum

slin1237:
## What this changes

This lets chat requests include **video** when using the vLLM inference engine. Before this change the vLLM path accepted still images only and rejected any request that contained a video, even though the rest of the system was already able to handle video.

## Why it was needed

The image and video paths were almost identical, but two problems stopped video from working:

1. **Videos have to be described frame by frame.** The engine expects one marked section per moment in time. The code only produced this frame-by-frame layout for one newer model variant and fell back to a single combined block for the main Qwen video model. A single block made the engine fail the moment a video had more than one frame.

2. **The marked region for the video started one step too late.** It began just after the "start of picture" marker instead of on it. The engine looks for that marker to line up each frame, so it never found the first one. This broke even single-frame videos.

## What was changed

- Video requests are now accepted on the vLLM path (previously images only).
- The frame-by-frame video layout is now produced for the whole Qwen video model family, not just one variant.
- The marked region for a video now includes the leading "start of picture" marker, so the engine can line up every frame.
- A shared "media type" value (image / video / audio) is used across the engines that take prepared data, instead of a one-off yes/no "is this a video" flag, so the engines agree on how media is labeled.

Scope: this focuses on the vLLM engine. The other engine (SGLang) is intentionally left for a later change, audio is still not handled on this path, and it is one video per request.

## Testing

Verified live on a machine with four graphics cards running the Qwen vision model (Qwen3-VL):

- A short test video that fades from **red to blue** was sent through the system. The model answered "**red, blue**" — confirming the video was decoded, delivered to the engine, and understood from end to end.
- Confirmed the video data travels correctly over both delivery methods (packed into the request, and via the faster shared-memory hand-off).
- Existing automated tests and the new video tests pass.

## How the two data-delivery methods compare

The image or video data has to be handed from the front door of the system to the engine. There are two ways to do it: **packed into the request** (the data is copied into the message), or **shared memory** (the data is written once to an area both sides can see, and only a small pointer is sent). While testing, I measured both.

**One request at a time — shared memory is faster, and the gap grows with data size:**

| Data per request | Packed into request | Shared memory | Shared memory saves |
|---|---|---|---|
| 25 MB (1 image)  | 152 ms | 128 ms | 24 ms (16%) |
| 50 MB (2 images) | 245 ms | 206 ms | 39 ms (16%) |
| 100 MB (4 images)| 414 ms | 358 ms | 56 ms (14%) |

Shared memory is consistently about **15% faster per request**, and the time saved grows with the amount of data, because it skips copying that data into and out of the message.

**Many requests at once:**

- With **one** graphics card, both methods reach the same rate (about **35 requests per second**). The card's own compute is the limit, so how the data is delivered does not matter.
- With **four** graphics cards, both reach about **105 requests per second**. At that point the step that prepares each image (decoding it and turning it into numbers) is the limit, and that work is identical for both methods, so their total rates come out about the same. I confirmed this ceiling is on the system side, not the test tool: running two test tools at once did not raise the combined rate.

**Takeaway:** shared memory gives a clear per-request speed-up that grows with data size and frees up processor time. A gain in *total throughput* only appears when moving the data — rather than preparing the image or the graphics card itself — is the bottleneck. That is most likely with very large media (long videos) or when the engine runs on a separate machine.

## Summary by CodeRabbit

* **New Features**
  * Added shared multimodal `modality` support to route image vs video through the gRPC pipeline (including vLLM assembly).
  * Added structural-prefix support to placeholder ranges for prompt token expansion.
  * Updated Qwen3-VL vision specs to generate per-frame video placeholder layouts for base models.
* **Bug Fixes**
  * Video now uses modality-specific tensors/placeholders and expands video placeholders (audio remains rejected).
  * Multimodal caching hashes now use video hashes for video items; placeholder sizing aligns with modality.
* **Tests**
  * Extended modality mapping/round-trip checks, added structural-prefix coverage, and added Qwen3-VL video regression tests.
* **Chores**
  * Bumped Python gRPC proto package dependency/version.

### DIFF
--- a/crates/grpc_client/proto/common.proto
+++ b/crates/grpc_client/proto/common.proto
@@ -115,3 +115,12 @@ message RemoteTensorHandle {
   bytes descriptor = 2;
   uint64 nbytes = 3;
 }
+
+// Multimodal input modality, shared across engine protos so the servicers can
+// route each request to the matching encoder (image vs video vs audio).
+enum Modality {
+  MODALITY_UNSPECIFIED = 0;
+  IMAGE = 1;
+  AUDIO = 2;
+  VIDEO = 3;
+}
--- a/crates/grpc_client/proto/tokenspeed_scheduler.proto
+++ b/crates/grpc_client/proto/tokenspeed_scheduler.proto
@@ -191,15 +191,8 @@ message PlaceholderRange {
   uint32 length = 2;
 }
 
-enum Modality {
-  MODALITY_UNSPECIFIED = 0;
-  IMAGE = 1;
-  AUDIO = 2;
-  VIDEO = 3;
-}
-
 message MultimodalItem {
-  Modality modality = 1;
+  smg.grpc.common.Modality modality = 1;
   bytes content_hash = 2;
   // Primary input to the multimodal encoder for this item. For vision models
   // this is the preprocessed image/video tensor; for audio models this can be
@@ -325,7 +318,7 @@ message GetModelInfoResponse {
   // non-vision workers before tokenization).
   bool supports_vision = 14;
   bool supports_multimodal = 15;
-  repeated Modality supported_modalities = 16;
+  repeated smg.grpc.common.Modality supported_modalities = 16;
   string model_dtype = 17;
   string multimodal_encoder_dtype = 18;
 }
--- a/crates/grpc_client/proto/vllm_engine.proto
+++ b/crates/grpc_client/proto/vllm_engine.proto
@@ -151,6 +151,11 @@ message MultimodalInputs {
   // Tensor keys that should remain on CPU (not transferred to GPU).
   // Maps to vLLM's MultiModalFieldConfig keep_on_cpu flag.
   repeated string keep_on_cpu_keys = 9;
+
+  // Input modality (image/video). The servicer routes video to vLLM's video
+  // modality (`pixel_values_videos` / `video_grid_thw`) and expands the video
+  // placeholder token instead of the image one. Unset (0) is treated as image.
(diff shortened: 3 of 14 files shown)


slin1237:
Addressed the CI failure + review feedback (pushed `3401545d`):

- **CI (`unit-tests` clippy)**: `trivially_copy_pass_by_ref` fired on the generated `Modality::as_str_name(&self)` once the enum moved to `common.proto` — the `common_proto` module allowed `clippy::all` but not that *pedantic* lint. Added it to the module `#![allow(...)]`, matching `tokenspeed_proto`.
- **Codex P1** (encoder_servicer): the TokenSpeed *encode* servicer still read `tokenspeed_scheduler_pb2.IMAGE/VIDEO/MODALITY_UNSPECIFIED` — switched to `common_pb2` and dropped the now-unused import.
- **CodeRabbit** (getattr fallback): replaced `getattr(common_pb2, "IMAGE", 1)` defaults with direct `common_pb2.IMAGE/VIDEO/AUDIO` (fail-fast, matching the vLLM servicer).
- **Codex P2** (proto floor): bumped `smg-grpc-proto` to `0.4.13` (the version that ships `modality`) and raised the servicer requirement to `>=0.4.13`, so a stale proto can't satisfy it.
- **Claude nit** (ImageEmbeds): tightened `ensure_image_or_video` to accept only `Image | Video` — `ImageEmbeds`/`Audio` stay rejected (unchanged from `ensure_image_only`), so no silent relaxation.

Local: `cargo check`/`clippy` + `ruff` clean.

PR #1914, opened 2026-07-13
refactor(grpc): centralize backend×modality capability check

slin1237:
### Problem

Which engine supports which modality (audio = TokenSpeed-only; video = vLLM/TokenSpeed; image = all-but-MLX) was enforced by scattered `bail!`s deep in `assemble.rs`, with inconsistent messages, *after* media was already fetched and preprocessed.

### Solution

Add a single capability matrix and check it early, at worker selection, before any media is fetched — one consistent `multimodal_not_supported` error. Keep a defense-in-depth assertion in assembly.

## Changes

- `multimodal/capability.rs` (new): `runtime_supports_modality(runtime, modality)` + `ensure_backend_supports_modalities(...)`, with a full backend×modality unit test.
- `grpc/client.rs`: `GrpcClient::runtime_type()`.
- `common/stages/worker_selection.rs`: early rejection using the request's `MultimodalIntermediate` + selected runtime.
- `multimodal/assemble.rs`: merge `into_single_image_batch`/`into_single_vision_batch` → `into_single_batch`; centralize the per-modality check.

## Test Plan

`cargo test -p smg --lib routers::grpc` passes. Behavior-preserving: the same (backend, modality) combinations are rejected as before — just earlier and with one consistent message (verified against each prior `bail!`).

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/model_gateway/src/routers/grpc/client.rs
+++ b/model_gateway/src/routers/grpc/client.rs
@@ -184,6 +184,19 @@ impl GrpcClient {
         matches!(self, Self::TokenSpeed(_))
     }
 
+    /// Runtime type backing this client. Lets shared logic (e.g. the multimodal
+    /// capability matrix) key on the backend without matching every variant.
+    pub fn runtime_type(&self) -> crate::worker::RuntimeType {
+        use crate::worker::RuntimeType;
+        match self {
+            Self::Sglang(_) => RuntimeType::Sglang,
+            Self::Vllm(_) => RuntimeType::Vllm,
+            Self::Trtllm(_) => RuntimeType::Trtllm,
+            Self::Mlx(_) => RuntimeType::Mlx,
+            Self::TokenSpeed(_) => RuntimeType::TokenSpeed,
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


slin1237:
Addressed all review comments (pushed):

- **coderabbit (Major):** `ensure_client_supports_intermediate` now delegates to `capability::ensure_backend_supports_modalities`; dropped the duplicated loop and the now-unused `runtime_supports_modality` import.
- **gemini:** `ImageEmbeds` separated from `Image` in the matrix — it's now **TokenSpeed-only** (pre-computed embeddings are an EPD feature); unit-test matrix updated (Sglang/vLLM/TRT-LLM/MLX = false, TokenSpeed = true). Note it can't actually reach the check (a `MediaBatch` only yields Image/Video/Audio) — kept for correctness/defense-in-depth.
- **gemini:** doc links switched from `super::` to absolute `crate::…` paths.

`cargo test -p smg --lib routers::grpc::multimodal` (capability + assemble) green; `clippy -D warnings` clean.

PR #1964, opened 2026-07-24
fix(router): support model aliases end to end

junliu-mde:
## Summary

- resolve model aliases once at each request entry: the regular HTTP router, the PD router, and the gRPC pipeline
- use the canonical model ID for every per-model lookup downstream — hash rings, retry configs, policies, tokenizers, and metrics labels — since only `get_by_model` understands aliases
- forward the canonical model ID to backends, which were registered under it and have never heard of the alias
- accept aliases at the `/v1/responses` availability gate, which runs before canonicalization
- hold the alias entry across the whole decision in every alias write, so a concurrent registration cannot erase or shadow an alias
- resolve alias/model-ID collisions from both sides, so registration order does not change the outcome
- keep model discovery and statistics canonical-only

## Behavior change

A request addressed to an alias is answered with the canonical model ID, matching how the OpenAI API reports the model it actually ran.

## Tests

- registry: alias resolution, replacement, removal, handover to a remaining model, and collision with a registered model ID
- `/v1/responses` gate: accepts aliases, still rejects unknown models and the `unknown` wildcard
- gRPC PD pipeline: an alias is canonical before worker selection and tokenizer lookup
- router selection for aliased PD models
- HTTP integration, regular and PD: asserts the body the worker actually received carries the canonical model ID

## Summary by CodeRabbit

* **New Features**
  * Added model-alias support across HTTP and gRPC routing, including standard, prefill/decode, rerank, and audio transcription flows.
  * Alias model names are resolved at request entry and the canonical model is forwarded consistently to workers.

* **Bug Fixes**
  * Unknown model names continue to be rejected with consistent `404 NOT_FOUND` responses.
  * Worker alias availability and routing selection stay correct when workers are replaced or removed.

* **Tests**
  * Added/extended unit and integration tests to verify alias canonicalization, correct request forwarding, and expected gateway/worker behavior across routes.

### DIFF
--- a/model_gateway/src/routers/grpc/common/responses/utils.rs
+++ b/model_gateway/src/routers/grpc/common/responses/utils.rs
@@ -98,14 +98,17 @@ pub(crate) async fn ensure_mcp_connection(
     Ok((false, Vec::new()))
 }
 
-/// Validate that workers are available for the requested model
+/// Validate that workers are available for the requested model.
+///
+/// Runs on the client-supplied name, before the pipeline canonicalizes it, so
+/// it has to accept aliases as well as canonical model IDs. `contains_model`
+/// covers both; listing `get_models()` and testing membership would reject
+/// every alias here.
 pub(crate) fn validate_worker_availability(
     worker_registry: &Arc<WorkerRegistry>,
     model: &str,
 ) -> Option<Response> {
-    let available_models = worker_registry.get_models();
-
-    if !available_models.contains(&model.to_string()) {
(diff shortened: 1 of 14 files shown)


junliu-mde:
> ⚠️ The sha of the head commit of this PR conflicts with #1961. Mergify cannot evaluate rules on this PR. Once #1961 is merged or closed, Mergify will resume processing this PR. ⚠️

Pushed on wrong base... rebased

slin1237 (approved):

PR #1973, opened 2026-07-27
ci: add engine version watch and nightly failure triage automation

slin1237:
### Problem

Engine upgrades (SGLang/vLLM/TensorRT-LLM/TokenSpeed) only happen when someone remembers to check upstream, and the nightly benchmark/eval workflows (bfcl, tau2, benchmark, docker builds) fail with nobody looking.

### Solution

Two scheduled workflows built on the same rails as the existing Claude code-review action (`k8s-runner-cpu` pod-env API key, `claude-opus-4-6`, gh CLI), both structured so **Claude only runs when there's something to act on** — detection is deterministic and token-free.

**`engine-version-watch.yml`** (every 3 days + manual):
- `scripts/check_engine_versions.sh` reads the pins from the same files upgrade PRs edit (installer scripts, release-docker defaults, TokenSpeed REF) and compares against PyPI, pypi.nvidia.com, and tokenspeed `main`. Verified live: it currently reports exactly the four known-behind engines.
- Engines with an open `engine-watch` issue for that version are dropped (dedup), so each release is reported once.
- For the remainder, Claude researches the upgrade — pin locations, whether version-tied workarounds become obsolete (via requires_dist metadata), wheel/image existence — and files one issue per engine with an upgrade checklist. Issue-only by design: no file edits, no PRs.

**`nightly-triage.yml`** (daily 15:30 UTC, after the ~7h bfcl/tau2 windows + manual):
- Deterministically gathers non-success completed runs of the six nightly workflows from the last 24h; exits token-free when everything passed.
- On failures, Claude pulls `--log-failed`, classifies (regression vs infra/upstream flake vs config), reruns **only** clear first-attempt infra flakes exactly once, and appends the day's verdict to one rolling `[nightly-triage]` issue per workflow — recurring failure signatures surface instead of vanishing.

Both are cron/dispatch-only (no fork-triggerable inputs reach the prompts), labels are created idempotently, and permissions are scoped (`issues: write`, plus `actions: write` only for the triage rerun).

## Test Plan

- `scripts/check_engine_versions.sh` run live: emits correct current/latest for all four engines against today's upstream state.
- Both workflows YAML-validated; gather/dedup steps use `gh run list`/`gh issue list` JSON paths exercised manually.
- After merge: `workflow_dispatch` each once to shake out the runner environment (I can babysit the first runs).

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added **Engine Version Watch** to periodically check for newer supported inference engine versions and automatically file one enhancement issue per updated engine.
  * Added **Nightly Triage** to review daily workflow failures, collect evidence, classify likely causes, rerun clear infra/flake cases, and post results to per-day rolling issues.
* **Chores**
  * Added an engine version checking script and supporting workflow automation with scheduled/manual triggers, controlled concurrency, and safer API key handling.

### DIFF
--- a/.github/workflows/engine-version-watch.yml
+++ b/.github/workflows/engine-version-watch.yml
@@ -1,0 +1,166 @@
+name: Engine Version Watch
+
+# Every 3 days: detect newer engine releases and have Claude file one
+# researched upgrade issue per engine (deduped against open issues, so a
+# version is only reported once).
+
+on:
+  schedule:
+    - cron: '23 8 */3 * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: engine-version-watch
+  cancel-in-progress: false
+
+jobs:
+  detect:
+    runs-on: k8s-runner-cpu
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      issues: write
+    outputs:
+      pending: ${{ steps.dedup.outputs.pending }}
+      has_pending: ${{ steps.dedup.outputs.has_pending }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Install gh CLI
+        run: |
+          if ! command -v gh &>/dev/null; then
+            mkdir -p "$HOME/.local/bin"
+            GH_VERSION="2.74.0"
+            GH_SHA256="e55c9d49dc49c0b0fef0a9acd3510482fd9e27ff52ae80f8a6e838cd25b4cd89"
+            curl -fsSL --connect-timeout 10 --max-time 120 -o /tmp/gh.tgz \
+              "https://github.com/cli/cli/releases/download/v${GH_VERSION}/gh_${GH_VERSION}_linux_amd64.tar.gz"
+            echo "${GH_SHA256}  /tmp/gh.tgz" | sha256sum --check --quiet
+            tar xzf /tmp/gh.tgz --strip-components=2 -C "$HOME/.local/bin" "gh_${GH_VERSION}_linux_amd64/bin/gh"
+            echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+          fi
+
+      - name: Check versions
+        id: check
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          bash scripts/check_engine_versions.sh | tee versions.jsonl
+          jq -cs '[.[] | select(.update)]' versions.jsonl > updates.json
+          echo "updates=$(cat updates.json)" >> "$GITHUB_OUTPUT"
+
+      - name: Drop engines that already have an open issue
+        id: dedup
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh label create engine-watch --repo "$GITHUB_REPOSITORY" \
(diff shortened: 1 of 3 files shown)


slin1237:
Addressed the review findings; all fixes are in the current head.

**Fixed**
- PEP 440 pre-release ordering (codex): comparisons and the tensorrt-llm latest-selection map `rc` through `~` before `sort -V`, so a final release supersedes its RCs (`1.3.0rc18` → `1.3.0` is detected; an RC never "upgrades" a GA pin).
- Empty-extraction guards: every pin/upstream lookup fails the detect job loudly via `require()` instead of emitting `current: ""`.
- curl timeouts (`--connect-timeout 10 --max-time 60`) on all lookups and the gh download.
- Duplicate-run TOCTOU: both workflows declare `concurrency` groups (no cancel-in-progress).
- Least-privilege permissions: workflow level is `contents: read`; `issues: write` scoped to the jobs that file/comment, `actions: write` only on triage, collect gets `actions: read`.
- `persist-credentials: false` on every checkout.
- gh CLI tarball is sha256-pinned and verified before extraction.
- `in:title` normalization: dedup and the rolling-issue lookup now exact-match against `--json title` output instead of GitHub search.
- `ANTHROPIC_API_KEY` presence validated with a clear `::error` before masking.
- Collect no longer converts API errors into "no failures" — the `|| rows='[]'` fallback and stderr suppression are gone.
- `enhancement` label ensured idempotently alongside `engine-watch`.
- Late-finishing runs (codex): valid for nightly-benchmark (`timeout-minutes: 1440` straddles the 15:30 collection; bfcl/tau2 at 360m cannot). Lookback is now 48h and the triage prompt skips run_ids already recorded in the rolling issue.
- `WebFetch`/`WebSearch` added to the triage allowlist; `WebSearch` added to engine-watch.

**Skipped**
- "Keep the Claude prompt declarative": intentional — these are unattended scheduled runs on trusted input (cron/dispatch only); exact commands are the rails that keep behavior deterministic.

**Validation**
- detect/collect ran green on k8s-runner-cpu; the detector reports all four engines behind, matching #1970/#1971/#1972 and the TokenSpeed drift.
- Full end-to-end runs including the Claude jobs are being exercised via a temporary `pull_request` trigger (claude-code-action rejects `push` events); it will be removed before merge.

slin1237:
Pre-merge test-run results (via a temporary trigger, since new workflows can't be dispatched before they exist on the default branch):

- **detect / collect (deterministic halves): green on k8s-runner-cpu.** The detector emitted correct results and proved the design live: the moment #1971 merged, vllm dropped out of the pending list (pins are read from the files the upgrade PRs edit), leaving sglang / tensorrt-llm / tokenspeed.
- **Caught a real bug**: the Claude jobs failed OIDC token exchange — `id-token: write` was missing (also from the original draft). Fixed on both jobs; this would otherwise have surfaced only on the first scheduled run.
- **Platform facts learned**: `claude-code-action@v1` rejects `push` events outright, and on `pull_request` it refuses to run unless the workflow file is byte-identical to the default-branch version (anti-tamper for the app-token exchange). So the Claude legs cannot execute pre-merge, by design — everything up to and including the action's validation gate is green.
- Temporary triggers are removed; the head is the final merge-ready state.

Post-merge: `workflow_dispatch` both workflows once to exercise the Claude legs end-to-end.

PR #1991, opened 2026-07-28
fix(multimodal): composite transparent images for Kimi-K3

key4ng:
### Problem

Kimi-K3's reference vision processor composites alpha-carrying images over a configured background *before* the NaViT resize and patchify ([`kimi_k3_vision_processing.py#L110-L114`](https://huggingface.co/moonshotai/Kimi-K3/blob/main/kimi_k3_vision_processing.py#L110-L114)):

```python
item = ensure_media_type(
    item,
    transparent_bg_config=self._transparent_bg_config,
    transparent_bg_fill_stage=self._transparent_bg_fill_stage,
)
resize_config = self.get_resize_config(item)
```

SMG registered `kimi-k3` against `KimiK25Processor`, which has no such step — it *drops* the alpha channel via `to_rgb8()`. Dropping alpha is not the same operation as compositing over a background. A fully transparent PNG pixel usually stores RGB `(0,0,0)` beneath its alpha, so dropping alpha normalizes it to `-1.0` — solid black — where the model was trained to see the chessboard the checkpoint describes.

Anything served as RGBA reached the encoder with its background inverted: logos, icons, diagrams exported with transparency, screenshots with rounded corners. Opaque JPEG/PNG traffic was unaffected, which is why MMMU and OCRBench did not surface it.

Only the gRPC worker path is affected. On the HTTP/OpenAI proxy path the engine preprocesses, so SMG's Rust never runs.

### Solution

Two commits, so the refactor is reviewable separately from the behavior change.

**1. `refactor(multimodal)` — a provable no-op.** K2.5 and K3 share the MoonViT stack and differ only in parameters, but there was nowhere to express a K3-specific difference. The reference factors its NaViT helpers into a shared `media_utils` module with a thin per-model processor class on top; this mirrors that split using the shape `qwen_vl_base` already provides for the Qwen VL family. A test asserts both processors still produce byte-identical output.

**2. `fix(multimodal)` — the compositing step.** All values come from the checkpoint, never hardcoded, matching the reference's fallback to dropping alpha when `transparent_bg_config` is absent. Hardcoded checkpoint values are the bug class that caused this incident in the first place.

The fill stage is load-bearing, not cosmetic: a chessboard is generated at the resolution of the image it lands on, so `before_resize` and `after_resize` yield different square sizes relative to the content. K3's checkpoint ships `after_resize`.

## Changes

- **`processors/moonvit.rs`** (new): `MoonVitProcessorBase` + `MoonVitConfig` hold the shared resize → pad → normalize → patchify pipeline, moved verbatim out of `kimi_k25.rs`.
- **`processors/kimi_k25.rs`**: reduced to a thin wrapper that `Deref`s to the base. Declares `transparent_bg: None` — its reference class has no transparency handling at all, so a config carrying the keys must not change its output.
- **`processors/kimi_k3.rs`** (new): second thin wrapper, `model_name: "kimi-k3"`, resolving the checkpoint's transparency settings per request.
- **`transforms.rs`**: `TransparentBgConfig` / `TransparentBgPattern` / `TransparentBgFillStage` deserialize straight from a `preprocessor_config.json`; `fill_transparent_bg` blends `a*img + (1-a)*bg` with the reference's truncating uint8 cast.
- **`transforms.rs`**: `resize_straight_alpha`. Our SIMD resizer premultiplies by alpha by default, dragging RGB toward zero wherever alpha is partial. Pillow convolves channels independently, so compositing *after* a resize requires the non-premultiplied variant.
- **`preprocessor_config.rs`**: lifts `transparent_bg_config` / `transparent_bg_fill_stage` out of `media_proc_cfg`; `transparent_bg()` returns them together, since the stage is only meaningful alongside a config.
- **`processor.rs`**: `kimi-k3` / `kimi_k3` now resolve to `KimiK3Processor`.

## Test Plan

`cargo test -p llm-multimodal` — 274 lib tests + all integration suites (incl. 81 golden tests) pass, 0 failures. 12 new tests:

| Test | Asserts |
|---|---|
| `opaque_images_match_k25_byte_for_byte` | commit 1 is a no-op |
| `transparent_pixels_composite_over_the_configured_background` | **the fix.** Fully transparent input over a white background normalizes to `+1.0`; before this PR it was `-1.0` |
| `transparency_config_does_not_reach_k25` | K2.5 still drops alpha given the same config |
| `opaque_images_ignore_the_transparency_config` | a declared background does not perturb opaque images |
| `fill_stage_paints_the_board_at_its_own_resolution` | `before_resize` ≠ `after_resize` under a 4× downscale |
| `chessboard_background_matches_reference` | every pixel of a 24×24 board matches `_create_chessboard_background`, both `square_on_top_left` phases |
| `fill_transparent_bg_blends_partial_alpha` | exact float32 arithmetic + truncating cast |
| `straight_alpha_resize_keeps_rgba_and_ignores_premultiply` | RGB under fully transparent pixels survives the resize |
| `test_parse_kimi_k3_transparency_settings` | K3's real checkpoint JSON deserializes end to end |
| `test_transparency_fill_stage_defaults_without_explicit_key` | absent stage falls back to `before_resize`, as the reference does |

Plus `fill_transparent_bg_is_identity_without_alpha`, `fill_transparent_bg_survives_zero_square_size`, `transparent_bg_config_fills_missing_fields`.

Before → after for a fully transparent 56×56 input with K3's config, mean = std = 0.5:

| | normalized value | what the encoder sees |
|---|---|---|
| before | `-1.0` | solid black |
| after | `+1.0` / `+0.569` | chessboard (255 / 200) |

### Deliberately out of scope

Kept separate so this PR stays reviewable and bisectable:

- **`in_patch_limit`**: K3's checkpoint raises it to 65536; SMG uses K2.5's 16384. Large images get ~4× fewer visual tokens than they should (4000×3000 → 4200 instead of 15444). Changes token accounting on both models, so it needs its own PR.
- Resize kernel divergence above 3.2 MP, K2.5's cross-path alpha inconsistency (shared with Qwen/Phi/Inkling), missing Kimi golden fixtures.

<details>
<summary>Checklist</summary>

</details>

## Summary by CodeRabbit

* **New Features**
  * Added support for Kimi K3 vision processing.
  * Added configurable transparent-background handling for images with alpha channels, including chessboard backgrounds and selectable processing stages.
  * Added improved alpha-aware image resizing.
  * Added shared MoonViT-based image preprocessing for Kimi vision models.
* **Bug Fixes**
  * Corrected model registration so Kimi K3 uses its dedicated processor.
  * Preserved expected processing behavior for opaque images and Kimi K2.5.

### DIFF
--- a/crates/multimodal/src/vision/preprocessor_config.rs
+++ b/crates/multimodal/src/vision/preprocessor_config.rs
@@ -278,9 +278,14 @@ impl PreProcessorConfig {
                 .and_then(|v| v.as_u64())
                 .map(|v| v as usize);
         }
-        // Also extract Kimi-specific limits into the extra map
-        // so processors can read them via get_extra()
-        for key in ["in_patch_limit", "patch_limit_on_one_side"] {
+        // Also extract Kimi-specific limits and the transparency settings into
+        // the extra map so processors can read them via get_extra()
+        for key in [
+            "in_patch_limit",
+            "patch_limit_on_one_side",
+            "transparent_bg_config",
+            "transparent_bg_fill_stage",
+        ] {
             if !config.extra.contains_key(key) {
                 if let Some(v) = media_cfg.get(key) {
                     config.extra.insert(key.to_string(), v.clone());
@@ -432,6 +437,20 @@ impl PreProcessorConfig {
         self.extra
             .get(key)
             .and_then(|v| serde_json::from_value(v.clone()).ok())
+    }
+
+    /// The checkpoint's alpha-flattening behavior, if it declares any.
+    ///
+    /// `None` means the checkpoint asks for no compositing, matching the
+    /// reference's behavior when `transparent_bg_config` is absent. The fill
+    /// stage is only meaningful alongside a config, so it is read here rather
+    /// than exposed on its own.
+    pub fn transparent_bg(&self) -> Option<transforms::TransparentBg> {
+        let config = self.get_extra::<transforms::TransparentBgConfig>("transparent_bg_config")?;
+        let stage = self
+            .get_extra::<transforms::TransparentBgFillStage>("transparent_bg_fill_stage")
+            .unwrap_or_default();
+        Some(transforms::TransparentBg { config, stage })
     }
 
     // Common default values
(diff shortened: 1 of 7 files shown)


key4ng:
Superseded by #1984, which implements the same transparency fix plus the `in_patch_limit` correction, and follows SMG's convention that each vision processor owns its model-specific defaults (see `grpc/multimodal/config.rs`: `preprocessor_config.json` is optional). Closing in favor of that PR.

PR #1995, opened 2026-07-29
fix(kimi-k3): move the media wrapper and thinking-effort default to the prompt-encoding layer

key4ng:
## Motivation

Two divergences from the K3 reference implementation, both in prompt encoding rather than in the vision processor.

**1. The media wrapper carried no dimensions.** K3's image prompt is one wrapper per image:

```
<|media_begin|>image {width}x{height}<|media_content|><|media_pad|>…<|media_end|>
```

where the dimensions are the *pre-resize* decoded size. SMG emitted only the bare `<|media_pad|>` run, because K3 was routed to the K2.5 registry spec — and K2.5's jinja chat template emits its own, dimensionless wrapper. Every K3 image prompt was 9 tokens short of the reference and told the model nothing about the image's original size.

**2. Every request was missing the thinking-effort directive.** The K3 checkpoint splits rendering across two layers. `encoding_k3.build_chat_segments` injects no directive; the entry point above it, `tokenization_kimi.apply_chat_template`, runs `kwargs.setdefault("thinking_effort", "max")` first. vLLM calls the latter, so every served K3 request carries the directive. SMG's renderer is a faithful port of the *lower* layer and was wired directly into `TiktokenTokenizer::apply_chat_template`, so it omitted the directive — a 67-token divergence on every request, image or not.

## Approach

**Wrapper.** It cannot be built while rendering: the chat template runs before any media is fetched, so the dimensions do not exist yet. A new `KimiK3VisionSpec` builds it in `prompt_replacements`, from the sizes the preprocessor reports — the same place vLLM builds it (`kimi_k3.py::_get_prompt_updates`). The renderer emits a bare `<|media_pad|>` anchor per image and prompt expansion replaces it with the full block. `with_feature_span` keeps the encoder-feature positions on the pad run alone; the surrounding wrapper is text.

The K2.5 matcher is narrowed to K2.5 and K3 is registered ahead of it. The two families share the MoonViT transport layout but not a prompt shape — implicit sharing is exactly what produced the pixel-pipeline divergence fixed in #1984.

**Thinking effort.** `apply_kimi_k3_xtml` stays a faithful `build_chat_segments` port; a new `apply_kimi_k3_xtml_with_effort_default` models the served layer and is what `TiktokenTokenizer::apply_chat_template` now calls. The default is threaded through the fallback branch rather than pre-seeding `template_kwargs`, which preserves the precedence **explicit `thinking_effort` > OpenAI `reasoning_effort` > default** — pre-seeding would have permanently starved the `reasoning_effort` bridge. All seven golden fixtures stay byte-valid.

## Verification against the checkpoint

Run inside the K3 container on a B300 node, against `moonshotai/Kimi-K3`:

- **Wrapper composition.** For 1024x768, 4000x3000, 224x448, 3x4 and 512x512, `[<|media_begin|>] + encode("image {w}x{h}") + [<|media_content|>, <|media_pad|>, <|media_end|>]` is byte-identical to the reference's one-shot encoding of `make_image_prompt(w, h)`. This confirms the one real assumption in the spec: the media tokens are hard segment boundaries for the tiktoken encoder, so the dimension text can be encoded on its own. Wrapper cost beyond the bare pad: **9 tokens**.
- **Served entry point.** `tokenization_kimi.apply_chat_template` output is exactly `build_chat_segments` output prefixed by the `thinking_effort=max` directive, **67 tokens**.

Together these close the token accounting on the reference request: 1092 (SMG before) + 9 + 67 = **1168**, matching what vLLM sends.

## Behaviour change

Text-only K3 requests get 67 tokens longer, and image requests get 9 more per image. Both are the point — that is what the reference serves.

## Tests

- `registry::kimi_k3`: matcher by model_id and model_type; the exact wrapper layout and feature span; per-image dimensions across a two-image batch; a checkpoint missing the structural tokens fails loudly rather than silently emitting a bare pad run.
- `registry::kimi_k25`: K3 no longer resolves to the K2.5 spec, by model_id and by model_type.
- `encoders::kimi_k3_xtml`: the served path defaults to `max`; the default yields to both an explicit `thinking_effort` and an OpenAI `reasoning_effort`; the default is suppressed when thinking is off.
- `tests/kimi_k3_renderer.rs`: the end-to-end no-chat-template case now asserts the directive is present and is the *only* addition.

Full workspace suite: 99 suites, 4151 passed, 0 failed.

> [!NOTE]
> `cargo clippy --all-features` could not be run locally — it pulls the optional `opencv` dependency and the dev machine has neither `pkg-config` nor `opencv` installed. `cargo clippy --workspace --all-targets -- -D warnings` is clean, and no changed code sits behind a `cfg(feature)` gate. CI covers the all-features build.

## Summary by CodeRabbit

* **New Features**
  * Added multimodal support for Kimi-K3 vision models, including media wrapper placeholders, per-image dimension text, and correct image feature-span mapping.
  * Kimi-K3 chat rendering can default `thinking-effort` to `max` when no effort is specified.
* **Bug Fixes**
  * Kimi-K3 model IDs are no longer matched to Kimi-K2.5 specs.
  * Improved test tokenizer text and batch encoding so each input is encoded correctly.
* **Error Handling**
  * Added a dedicated error when a model spec fails to encode provided text.

### DIFF
--- a/crates/multimodal/src/registry/kimi_k25.rs
+++ b/crates/multimodal/src/registry/kimi_k25.rs
@@ -28,12 +28,9 @@ impl ModelProcessorSpec for KimiK25VisionSpec {
     }
 
     fn matches(&self, metadata: &ModelMetadata) -> bool {
-        // K2.5 only. Kimi-K3 shares the `<|media_pad|>` fill token
-        // (media_placeholder_token_id 163605) and the patchification layout, but
-        // its prompt carries a per-image `<|media_begin|>image {w}x{h}…` wrapper
-        // that K2.5's chat template emits itself — see `registry::kimi_k3`. The
-        // two do not share a pixel pipeline either; `vision::processors::kimi_k3`
-        // has the patch budget and alpha-compositing differences.
+        // K2.5 only — K3 shares the `<|media_pad|>` fill token and the
+        // patchification layout, but neither the prompt shape nor the pixel
+        // pipeline. See `registry::kimi_k3` and `vision::processors::kimi_k3`.
         let id = metadata.model_id.to_ascii_lowercase();
         id.contains("kimi") && id.contains("k2")
             || metadata
--- a/crates/multimodal/src/registry/kimi_k25.rs
+++ b/crates/multimodal/src/registry/kimi_k25.rs
@@ -28,16 +28,17 @@ impl ModelProcessorSpec for KimiK25VisionSpec {
     }
 
     fn matches(&self, metadata: &ModelMetadata) -> bool {
-        // Kimi-K3 uses the same `<|media_pad|>` placeholder
-        // (media_placeholder_token_id 163605), patchification layout and
-        // prompt-replacement shape, so it shares this spec. Note that the two
-        // do *not* share a pixel pipeline — see `vision::processors::kimi_k3`
-        // for the patch budget and alpha-compositing differences.
+        // K2.5 only. Kimi-K3 shares the `<|media_pad|>` fill token
+        // (media_placeholder_token_id 163605) and the patchification layout, but
+        // its prompt carries a per-image `<|media_begin|>image {w}x{h}…` wrapper
+        // that K2.5's chat template emits itself — see `registry::kimi_k3`. The
+        // two do not share a pixel pipeline either; `vision::processors::kimi_k3`
+        // has the patch budget and alpha-compositing differences.
         let id = metadata.model_id.to_ascii_lowercase();
-        id.contains("kimi") && (id.contains("k2") || id.contains("k3"))
+        id.contains("kimi") && id.contains("k2")
             || metadata
                 .config_model_type()
-                .is_some_and(|mt| mt == "kimi_k25" || mt == "kimi_k3")
+                .is_some_and(|mt| mt == "kimi_k25")
     }
 
     fn placeholder_token(&self, _metadata: &ModelMetadata) -> RegistryResult<String> {
@@ -126,31 +127,25 @@ mod tests {
     }
 
     #[test]
-    fn kimi_k3_matches_model_id_and_model_type() {
+    fn kimi_k3_does_not_use_the_k25_spec() {
+        // K3's prompt carries per-image dimensions that this spec cannot emit,
+        // so it must route to `kimi_k3` by model_id and by model_type alike.
         let tokenizer = TestTokenizer::new(&[("<|media_pad|>", 163605)]);
-        // Match by model_id containing kimi + k3.
         let config = json!({
             "model_type": "kimi_k3",
             "media_placeholder_token_id": 163605
         });
-        let metadata = ModelMetadata {
-            model_id: "moonshotai/Kimi-K3",
-            tokenizer: &tokenizer,
-            config: &config,
-        };
         let registry = ModelRegistry::new();
-        let spec = registry
-            .lookup(&metadata)
-            .expect("kimi_k3 -> kimi_k25 spec");
-        assert_eq!(spec.name(), "kimi_k25");
 
-        // Also match by model_type alone (id without a k3 hint).
-        let metadata_by_type = ModelMetadata {
-            model_id: "internal/checkpoint-final",
-            tokenizer: &tokenizer,
-            config: &config,
-        };
-        assert!(registry.lookup(&metadata_by_type).is_some());
+        for model_id in ["moonshotai/Kimi-K3", "internal/checkpoint-final"] {
+            let metadata = ModelMetadata {
+                model_id,
+                tokenizer: &tokenizer,
+                config: &config,
+            };
+            let spec = registry.lookup(&metadata).expect("kimi_k3 spec");
+            assert_eq!(spec.name(), "kimi_k3", "model_id {model_id}");
+        }
     }
 
     #[test]
--- a/crates/multimodal/src/registry/kimi_k3.rs
+++ b/crates/multimodal/src/registry/kimi_k3.rs
@@ -18,20 +18,14 @@ const MEDIA_END: &str = "<|media_end|>";
 
 /// Kimi-K3.
 ///
-/// K3 shares K2.5's NaViT transport layout and `<|media_pad|>` fill token, but
-/// not its prompt shape: the K3 checkpoint ships no jinja chat template, and its
-/// Python renderer splices a per-image `<|media_begin|>image {w}x{h}
-/// <|media_content|>…<|media_end|>` block into the raw text at each
-/// `<|kimi_image_placeholder|>` marker (`kimi_k3_processor.py::update_raw_text`).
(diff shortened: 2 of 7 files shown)


key4ng:
## Node validation: prompt shape and token parity

Ran this branch on the B300 node against the real `Kimi-K3` checkpoint, using vLLM's own server as the reference rather than a hand-written expectation.

**Setup.** One vLLM gRPC engine (`:8081`, `/raid/models/moonshotai/Kimi-K3`, TP8). Two gateways built from source and run natively against that same engine, so the only variable is the SMG binary:

| Gateway | Build |
|---|---|
| `new` (`:8210`) | this branch, `87901680` |
| `old` (`:8220`) | pre-fix baseline, #1984 tip |

**Reference.** Before handing the GPUs to the gRPC engine, I captured the same requests against vLLM's HTTP server (`:8200`) on the same checkpoint — that path does its own chat templating and its own preprocessing, so it is the behaviour we are trying to match. Asking for `prompt_logprobs` makes the engine echo back the decoded prompt, which gives a direct read of the wrapper instead of inferring it:

```
<|media_begin|>image 1024x768<|media_content|><|media_pad|>…<|media_pad|><|media_end|>
```

for a 1024×768 image — 1036 pads, and 3 marker tokens + 6 tokens of `image 1024x768` = the 9-token wrapper.

### Token parity

`usage.prompt_tokens`, same requests to all three:

| Case | vLLM (ref) | new | old |
|---|---|---|---|
| 1 image, 1024×768 | 1138 | **1138** | 1062 |
| 1 image, 4000×3000 | 15547 | **15547** | 15470 |
| 2 images, 1024×768 then 4000×3000 | 16594 | **16594** | 16508 |
| 2 images, both 1024×768 | 2183 | **2183** | 2098 |
| text only | 89 | **89** | 22 |
| text, `thinking_effort=low` | 90 | **90** | 90 |
| text, `thinking_effort=high` | 90 | **90** | 90 |
| 1 image, `thinking_effort=low` | 1139 | **1139** | 1130 |

**new matches the reference on 8/8. old matches on 2/8** — exactly the two cases that pass an explicit effort and contain no image, i.e. the two things the old build already got right.

Every baseline gap decomposes exactly, with no remainder:

- text-only, `89 − 22 = 67` → the missing thinking-effort directive.
- 1 image with explicit effort, `1139 − 1130 = 9` → the missing wrapper, directive already present.
- 1 image, default effort, `1138 − 1062 = 76 = 67 + 9`.
- 1 image 4000×3000, `15547 − 15470 = 77 = 67 + 10` — the wrapper is 10 here because `image 4000x3000` is 7 tokens, not 6.
- 2 images, `16594 − 16508 = 86 = 67 + 9 + 10`, and `2183 − 2098 = 85 = 67 + 9 + 9`.

The per-image term tracks the actual dimensions rather than being a constant, which is the part that matters.

Gateway debug logging agrees from the inside — for the 4000×3000 case the new build reports `item_sizes=[(4000, 3000)]`, `original_len=94 expanded_len=15547 placeholder_count=1` with 15444 pads, so the wrapper's non-pad cost is `15547 − 93 − 15444 = 10`.

### Width×height ordering, and pre-resize dimensions

Token counts cannot catch a `W`/`H` swap: swapping the operands of `x` leaves the token multiset unchanged, so `image 4000x3000` and `image 3000x4000` both cost 7. I verified that against the checkpoint tokenizer rather than assuming it, then tested ordering a different way — asking the model to repeat the dimensions line from its own context. The numbers only exist there as text, so a correct answer is a direct read of the wrapper.

Strip images make this unambiguous: the same two numbers in both orders, and few enough pads that retrieval is reliable.

| Image | new | old |
|---|---|---|
| 4000×64 | `image 4000x64` ✅ | `image 1024x1024` (guess) |
| 64×4000 | `image 64x4000` ✅ | `image 256x2048` (guess) |
| 1536×384 | `image 1536x384` ✅ | `image 1000x1000` (guess) |

The mirror pair both read back correctly, so this is not the model guessing an orientation. `old` has no dimensions in context and guesses, as expected. Prompt tokens differ by a constant 9 (497 vs 488, 838 vs 829).

This also settles pre- vs post-resize: a 4000×64 strip is heavily resized for patching (~380 pads), yet the wrapper still states the original 4000×64 — `item_sizes` is captured before the resize.

### One caveat

SMG does not forward `prompt_logprobs` to the engine, so the byte-exact decoded prompt is only available on the vLLM path. On the SMG path the evidence is token parity, the gateway's own expansion counts, and the read-back above. I did try to falsify the wrapper contents rather than only confirm them: for the 4000×3000 image the model answered `image 3465x2599`, which would be alarming, but that string also costs 7 tokens so parity could not rule it out. It is a long-context retrieval failure across 15444 identical pad tokens — the same probe reads correctly whenever the pad run is short, and the mirror pair pins the ordering independently.

key4ng:
### Vendor-verifier accuracy check

Ran the two image benchmarks from [MoonshotAI/Kimi-Vendor-Verifier](https://github.com/MoonshotAI/Kimi-Vendor-Verifier) against this branch, to confirm the new media wrapper doesn't cost anything end-to-end.

| Benchmark | Build | Thinking effort | Accuracy | Samples | Wall time |
|---|---|---|---|---|---|
| OCRBench | this PR | `max` (explicit) | **0.896 ± 0.010** | 1000 | 9m43s |
| OCRBench | this PR | unset (PR default) | 0.895 ± 0.010 | 1000 | 9m38s |
| OCRBench | pre-PR (merged #1984) | unset | 0.890 ± 0.010 | 1000 | 8m39s |
| MMMU Pro | this PR | `max` (explicit) | **0.813 ± 0.009** | 1730 | 2h23m44s |

The three OCRBench numbers sit inside one standard error of each other, so this reads as no regression rather than an improvement. The middle row is the useful one for this PR specifically: it sends no `thinking_effort` at all, so the score comes from the `max` default that this PR installs in the K3 XTML render path.

Setup:

- native `smg` built from this branch (`--reasoning-parser kimi_k3 --tool-call-parser kimi_k3`), in front of a TP8 gRPC vLLM engine serving `moonshotai/Kimi-K3` on a B300 node
- unmodified KVV harness via `eval.py <bench> --thinking --think-mode opensource --temperature 1.0 --top-p 0.95 --max-connections 50 --epochs 1 --stream`, `--max-tokens` 16384 for OCRBench and 98304 for MMMU Pro
- token usage — OCRBench 1,349,304 (I 1,100,238 / O 249,066); MMMU Pro 10,209,534 (I 1,655,538 / O 8,553,996)

Two things to note about the numbers rather than the code:

- The harness pins MMMU Pro to the `standard (10 options)` subset (it reports the dataset as `MMMU_Pro_10c`), so that row is the standard 10-option split, not the vision-only split.
- One of the 1000 OCRBench samples hit the 16384-token cap mid-generation and was scored as-is; the harness runs with `fail_on_error: False`. MMMU Pro finished with no errors.

PR #2007, opened 2026-07-30
fix(kimi-k3): treat a sibling closer as the end of an assumed think channel

key4ng:
### Problem

Found during the **BEAM 1M long-context evaluation**: 16 of 700 answers (2.3%) came back from the gateway with `content` absent and the entire answer — raw XTML wrapper markers included — sitting in `reasoning_content`. Every one had `finish_reason: "stop"`, so nothing looked wrong at the transport level; the answer was simply filed in the wrong field, and a client reading `content` saw nothing.

The failure is **random and long-context specific**. Two hypotheses were tested and refuted before the real trigger was isolated:

| Hypothesis | Result |
|---|---|
| Short, low-deliberation prompts trip it | 0 / 12 reproduced |
| Greedy decoding (`temperature 0`) in a long chat | 0 / 20 reproduced |

It needs **both** a long multi-turn context *and* sampling — it never reproduces deterministically at a fixed prompt, which is why it survived to the eval stage.

**Root cause.** K3's tokenizer is `ThinkingToggle::DefaultOn` (`crates/tokenizer/src/tiktoken.rs:573`), so unless the caller passes `thinking: false`, `should_mark_reasoning_started` (`model_gateway/src/routers/grpc/utils/parsers.rs:20-26`) sets `in_reasoning = true` before a token arrives. That is correct in principle — the generation prompt really does end inside the think channel (`crates/tokenizer/src/encoders/kimi_k3_xtml.rs:285-288`):

```
… <|open|>message role="assistant"<|sep|><|open|>think<|sep|>
```

Since `<|open|>think<|sep|>` is part of the *prompt*, it never appears in the completion. The parser has to take the open channel on faith.

Sampling at `temperature 1.0` in a long chat, K3 sometimes disregards that prefilled channel and answers directly, closing the turn as `response`:

```
Yes — here's a concise summary …  well-being.<|close|>response<|sep|><|close|>message<|sep|>
```

The completion carries **only sibling closers**. `think_channel_end` recognised three terminators, none of them a closer:

| Terminator | Recognised before |
|---|---|
| `<\|close\|>think<\|sep\|>` | yes |
| `<\|open\|>response<\|sep\|>` | yes |
| `<\|open\|>tools<\|sep\|>` | yes |
| `<\|close\|>response<\|sep\|>` | **no** |
| `<\|close\|>message<\|sep\|>` | **no** |

With nothing matched it returned `None`, and the caller's fallback — meant for output truncated mid-thought — filed the remainder as reasoning. The markers survived because `strip_content_wrapper` runs on the content path only, which is also what makes their presence in `reasoning_content` a reliable fingerprint of the bug.

### Solution

Sibling closers now terminate the channel, gated on a new `think_observed` argument: true when a literal `<|open|>think<|sep|>` is in the text, false when the channel was merely *assumed* open because the caller pre-marked reasoning. Only in the assumed case does a closer mean anything — it proves the model ignored the prefill and answered directly. With the channel actually observed, the same closer is malformed output and must not retroactively empty the reasoning span.

The span is asymmetric: text before a sibling *opener* is reasoning, text before a sibling *closer* is the answer. The closer branch therefore returns `(from, from)` and lets `strip_content_wrapper` clean the markers. Candidates are compared on **marker position** rather than on the returned span, or `(from, from)` would always win.

**Known limitation (streaming only).** Streaming recovery is partial by construction: reasoning deltas already sent cannot be recalled, so the answer is repeated in `reasoning_text`. What the fix buys on that path is a complete `normal_text` and a cleared `in_reasoning` (which gates the tool parser). **Non-streaming is exact.**

## Changes

`crates/reasoning_parser/src/parsers/kimi_k3.rs` (1 file, +147/−29):

- `think_channel_end` takes a `think_observed: bool` and gains a third terminator kind — a sibling `response`/`message` **closer**, active only when the think channel was assumed rather than observed, spanning `(from, from)`.
- Terminator selection re-keyed from the returned span to the **marker's own position**, so the `(from, from)` candidate cannot win by starting at the lowest offset. Ties still favour think-close.
- Both call sites pass `think_observed`: `m_open.is_some()` non-streaming, the buffer search result streaming.
- `reasoning_text_ready_to_emit` holdback widened with `RESPONSE_CLOSE` and `MESSAGE_CLOSE`. Sharing the `<|close|>` prefix with `THINK_CLOSE` is not enough — `compute_overlap` matches whole suffixes, so a chunk ending `<|close|>res` would leak the `res` into `reasoning_text`.
- Four unit tests: the non-streaming regression, the `think_observed` guard, terminator precedence, and streaming recovery with the closer split across chunks (which also exercises the widened holdback).

No public API, config, or protocol change.

## Test Plan

Validated end-to-end on 8×B300 against the **exact pre-fix binary that served BEAM**, replaying real BEAM long-context requests. `seed` is forwarded to vLLM, so a `(question, seed)` pair is replayable and the same request can be issued byte-identically against both builds.

| Test | Pre-fix | Fixed |
|---|---|---|
| Controlled A/B, byte-identical request, 2 repeats each | misfiled every time | correct every time |
| 20 questions × 6 seeds = 120 requests @ 206k tokens | **6 / 120 (5.0 %)** | **0 / 120** |
| `cargo test -p reasoning-parser` | — | **98 passed** |

Model output was byte-identical across the A/B (same `completion_tokens`); only the field the answer landed in changed. No regression on the normal path: 106/120 still return both fields, 0 marker leaks, large reasoning traces (38 KB, 57 KB) intact.

**Full method, request and response detail in the [validation comment](https://github.com/lightseekorg/smg/pull/2007#issuecomment-5136977837) below.**

<details>
<summary>Checklist</summary>

</details>

<details>
<summary>Note on <code>pre-commit run --all-files</code></summary>

It cannot complete in my local environment: the clippy hook builds the whole workspace and `llm-multimodal`'s build script needs `pkg-config` + `opencv4`, which are not installed here. Pre-existing and unrelated to this change. Compensated with `cargo +nightly fmt --all --check`, `cargo test -p reasoning-parser`, and `cargo clippy -p reasoning-parser --all-targets --all-features -- -D warnings`, all clean, locally and in the node's Linux builder image.

</details>

### DIFF
--- a/crates/reasoning_parser/src/parsers/kimi_k3.rs
+++ b/crates/reasoning_parser/src/parsers/kimi_k3.rs
@@ -102,32 +102,70 @@ impl KimiK3Parser {
 
     /// Locate the end of the think channel in `text`, searching from `from`.
     ///
-    /// Returns `(reasoning_end, content_start)` for whichever comes first: the
-    /// think-close marker (consumed, so content resumes after it) or a sibling
-    /// `response`/`tools` opener (left in place for downstream parsing).
-    /// `None` when neither is present.
+    /// Returns `(reasoning_end, content_start)` for whichever terminator comes
+    /// first. `None` when none is present.
     ///
-    /// A sibling opener has to count as an implicit close because a forced tool
-    /// call can jump straight from the prefilled `<|open|>think<|sep|>` into
-    /// `<|open|>tools<|sep|>`; without it the tools channel would be swallowed
-    /// as reasoning. `message` is excluded — it wraps the whole assistant turn
-    /// and opens *before* the think channel.
-    fn think_channel_end(&self, text: &str, from: usize) -> Option<(usize, usize)> {
+    /// Three kinds of terminator, each with a different span:
+    ///
+    /// - The think-close marker: consumed, so content resumes after it.
+    /// - A sibling `response`/`tools` *opener*: left in place for downstream
+    ///   parsing, so reasoning ends and content starts at the marker. A sibling
+    ///   opener has to count as an implicit close because a forced tool call can
+    ///   jump straight from the prefilled `<|open|>think<|sep|>` into
+    ///   `<|open|>tools<|sep|>`; without it the tools channel would be swallowed
+    ///   as reasoning. `message` is excluded here — it wraps the whole assistant
+    ///   turn and opens *before* the think channel.
+    /// - A sibling `response`/`message` *closer*, but only when `think_observed`
+    ///   is false. See below; the span is `(from, from)`, i.e. no reasoning at
+    ///   all and content covering everything from `from`.
+    ///
+    /// `think_observed` says whether a literal `<|open|>think<|sep|>` was found
+    /// in `text`, as opposed to the channel merely being *assumed* open because
+    /// the caller pre-marked reasoning for a prefilled prompt. Reaching a closer
+    /// for a sibling channel without ever having entered the think channel is a
+    /// contradiction that can only be resolved one way: the model ignored the
+    /// prefilled `<|open|>think<|sep|>` and answered directly, closing the
+    /// channel as `response`/`message`. Everything from `from` is therefore the
+    /// answer, and `strip_content_wrapper` removes the stray closers. When the
+    /// think channel *was* observed, the same closer is just malformed output
+    /// and must not be allowed to retroactively empty the reasoning span.
+    ///
+    /// Note the asymmetry in span: text before a sibling *opener* is reasoning,
+    /// whereas text before a sibling *closer* is the response body.
+    fn think_channel_end(
+        &self,
+        text: &str,
+        from: usize,
+        think_observed: bool,
+    ) -> Option<(usize, usize)> {
         let close = self
             .think_close_re
             .find_at(text, from)
-            .map(|m| (m.start(), m.end()));
-        let sibling = [&self.response_open_re, &self.tools_open_re]
+            .map(|m| (m.start(), (m.start(), m.end())));
+        let sibling_open = [&self.response_open_re, &self.tools_open_re]
             .into_iter()
             .filter_map(|re| re.find_at(text, from).map(|m| m.start()))
             .min()
-            .map(|start| (start, start));
-        // Ties favour `close` (listed first); the markers cannot in fact start
-        // at the same offset, since one begins `<|close|>` and the other `<|open|>`.
-        [close, sibling]
+            .map(|start| (start, (start, start)));
+        let sibling_close = (!think_observed)
+            .then(|| {
+                [&self.response_close_re, &self.message_close_re]
+                    .into_iter()
+                    .filter_map(|re| re.find_at(text, from).map(|m| m.start()))
+                    .min()
+            })
+            .flatten()
+            .map(|start| (start, (from, from)));
+        // Compare on the marker's own position, not on the returned span, so the
+        // `(from, from)` case does not win by virtue of starting at the lowest
+        // possible offset. Ties favour whichever is listed first, preserving the
+        // original preference for `close`; the markers cannot in fact start at
+        // the same offset, since no two of them share a literal prefix.
+        [close, sibling_open, sibling_close]
             .into_iter()
             .flatten()
-            .min_by_key(|&(end, _)| end)
+            .min_by_key(|&(marker_start, _)| marker_start)
+            .map(|(_, span)| span)
     }
 
     /// Return the prefix of accumulated reasoning text that is safe to stream.
@@ -143,10 +181,21 @@ impl KimiK3Parser {
 
         // Sibling openers are held back alongside the think markers: a partial
         // `<|open|>too…` at the tail would otherwise leak into `reasoning_text`
-        // before it completes and ends the channel.
+        // before it completes and ends the channel. Sibling *closers* are held
+        // back for the same reason now that they too can end the channel, and
+        // sharing the `<|close|>` prefix with `THINK_CLOSE` is not enough —
+        // `compute_overlap` matches whole suffixes, so `<|close|>res` would
+        // otherwise leak the `res`.
         let overlap = Self::compute_overlap(
             after_open,
-            &[THINK_OPEN, THINK_CLOSE, RESPONSE_OPEN, TOOLS_OPEN],
+            &[
+                THINK_OPEN,
+                THINK_CLOSE,
+                RESPONSE_OPEN,
+                RESPONSE_CLOSE,
+                MESSAGE_CLOSE,
+                TOOLS_OPEN,
+            ],
         );
         if overlap > 0 {
             &after_open[..after_open.len() - overlap]
@@ -245,7 +294,8 @@ impl ReasoningParser for KimiK3Parser {
             return Ok(ParserResult::normal(self.strip_content_wrapper(text)));
         }
 
-        let Some((reasoning_end, content_start)) = self.think_channel_end(text, reasoning_start)
+        let Some((reasoning_end, content_start)) =
+            self.think_channel_end(text, reasoning_start, m_open.is_some())
         else {
             // Opened but never closed — e.g. output truncated mid-thought.
             // Classify the remainder as reasoning rather than lose it.
@@ -307,13 +357,21 @@ impl ReasoningParser for KimiK3Parser {
         // opener as well as the think-close marker is what lets a forced tools
         // channel reach the content phase instead of streaming as reasoning
         // indefinitely.
-        let r_start = self
-            .think_open_re
-            .find(&self.buffer)
-            .map(|m| m.end())
-            .unwrap_or(0);
+        //
+        // A sibling closer can also end it, but only when the think channel was
+        // assumed rather than observed — see `think_channel_end`. That recovery
+        // is inherently partial while streaming: reasoning deltas for what turns
+        // out to be the answer have already left the building and cannot be
+        // recalled, so the answer is repeated in `reasoning_text`. What the fix
+        // buys here is a correct and complete `normal_text` (a client reading
+        // only `content` sees the whole answer) and a cleared `in_reasoning`,
+        // which is what gates the tool parser. Non-streaming is exact.
+        let think_open = self.think_open_re.find(&self.buffer);
+        let think_observed = think_open.is_some();
+        let r_start = think_open.map(|m| m.end()).unwrap_or(0);
 
-        if let Some((reasoning_end, content_start)) = self.think_channel_end(&self.buffer, r_start)
+        if let Some((reasoning_end, content_start)) =
+            self.think_channel_end(&self.buffer, r_start, think_observed)
         {
             // Transition: reasoning phase ends.
             self.reason_phase_ended = true;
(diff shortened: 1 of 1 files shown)


key4ng:
# Validation result

Replayed real BEAM long-context requests (206k tokens) on 8×B300 against the exact pre-fix binary that served BEAM, then against this PR's build. Same request bytes, same seed, gateway binary the only variable.

| Test | Pre-fix | Fixed |
|---|---|---|
| Controlled A/B — byte-identical request, 2 repeats each | misfiled every time | correct every time |
| Full grid — 20 questions × 6 seeds = 120 requests | **6 / 120 (5.0 %)** misfiled | **0 / 120** |
| Marker leaks into `content` / `reasoning_content` | — | 0 |
| Normal path (both fields populated) | — | 106 / 120, unchanged |
| `cargo test -p reasoning-parser` | — | 98 passed |

The model emitted **identical bytes** before and after (same `completion_tokens`); only the field the answer landed in changed. Concretely, on the same request:

- **before** — `content` absent, `reasoning_content` = the answer + `<|close|>response<|sep|><|close|>message<|sep|>`
- **after** — `content` = the answer, `reasoning_content` = `null`

and `pre_fix_reasoning[:-47] == fixed_content` exactly, 47 bytes being the length of the two stray closers.

<details>
<summary><b>Test method</b></summary>

The bug is a random model behaviour, so a naive before/after run proves nothing: two invocations of the same prompt do not produce the same tokens. The method pins every variable except the gateway binary.

**Determinism.** `seed` is forwarded to vLLM (`crates/grpc_client/src/vllm_engine.rs:271`), so a `(question, seed)` pair is replayable. That turns a lucky hit into a request that can be replayed byte-identically against a second build. Requests are issued **one at a time** for the controlled A/B — vLLM honours the per-request seed but continuous batching changes the numerics of a batched forward pass, so two runs are only comparable if neither is sharing a batch.

**Procedure.**
1. Sweep `(question × seed)` at the BEAM sampling settings against the pre-fix binary until the misfiling appears.
2. Replay each hit twice, one request in flight, on the pre-fix binary — confirm byte-identical failure.
3. Swap the gateway binary only. Verify the swap via `/proc/<pid>/exe` rather than assuming it, since the new build overwrites the old path.
4. Replay the same requests (same sha256) on the fixed binary.
5. Run the full 120-request grid on each binary for a population-level rate.

**Detection criterion.** Empty `content` alone is ambiguous, so the surviving wrapper markers are the proof — `strip_content_wrapper` runs on the content path only, so their presence in `reasoning_content` means the answer came through the reasoning branch:

```python
def misfiled(content, reasoning):
    return (not (content or "").strip()) and (
        "<|close|>response" in (reasoning or "")
        or "<|close|>message" in (reasoning or ""))
```

</details>

<details>
<summary><b>Test settings</b> — hardware, engine, binaries, dataset, sampling</summary>

### Hardware and engine

| | |
|---|---|
| Node | 8 × B300 |
| Engine image | `smg:kimi-k3-vllm-kimi-k3` |
| Model | `moonshotai/Kimi-K3`, served as `kimi-k3`, gRPC on `127.0.0.1:8081` |

```
--grpc --trust-remote-code --load-format fastsafetensors --moe-backend auto
--gpu-memory-utilization 0.88 --tensor-parallel-size 8
--max-model-len 262144 --kv-cache-dtype fp8
--attention-config {"mla_prefill_backend":"TRTLLM_RAGGED","use_prefill_query_quantization":true}
--max-num-batched-tokens 16384 --max-num-seqs 32
--mm-processor-cache-gb 0 --enable-prefix-caching
--served-model-name kimi-k3 --port 8081
```

Note `--max-model-len 262144`. BEAM itself served 1M context; the bug reproduces at 206k, so a 1M context is not a precondition beyond the chat being long and multi-turn.

### Gateway — the two binaries

Identical arguments, only the executable differs:

```
smg --host 0.0.0.0 --port 8210 --prometheus-port 29210 \
    --worker-urls grpc://127.0.0.1:8081 \
    --model-path /raid/models/moonshotai/Kimi-K3 \
    --reasoning-parser kimi_k3 --tool-call-parser kimi_k3 \
    --request-timeout-secs 7200
```

| Build | sha256 |
|---|---|
| **pre-fix** — the exact binary that served BEAM | `76d2cfe495174ce358ad1ea0228ab4a55cd2842136cdbb2cc18b9c85294549d5` |
| **fixed** — built from this PR | `8a8bc56d39b8378258ff70c46822385ffa6ccbb95b97f6f96ee15ee040565959` |

Parser source in the fixed build is byte-identical to this branch: `sha256 80cc64bc166fe7f298ef22cbd7ab85e49e2b54f9041e1d0d3d168fe107f82595` on `crates/reasoning_parser/src/parsers/kimi_k3.rs`.

### Dataset and request construction

| | |
|---|---|
| Source | BEAM `chat_id 15` |
| Questions | 20 questions across 10 BEAM question types |
| History | truncated to 900,000 chars → **418 of 1936** messages, 896,668 chars |
| Truncation | whole messages dropped from the front, first survivor forced to a `user` turn (mirrors the harness's `truncate_messages`) |
| Final turn | BEAM `PROBING_TEMPLATE` appended as a `user` message |
| Prompt size | **206,722 tokens** |

Probing template, verbatim:

```
NOTE: Only provide the answer without any explanations. 
Question: {question}
```

### Sampling parameters

Identical to the BEAM run, plus an explicit `seed`:

```json
{
  "model": "kimi-k3",
  "temperature": 1.0,
  "top_p": 0.95,
  "max_tokens": 32768,
  "seed": 1,
  "chat_template_kwargs": {
    "thinking": true,
    "preserve_thinking": true,
    "thinking_effort": "max"
  }
}
```

</details>

<details>
<summary><b>Failing example — full request</b></summary>

`summarization#1`, `seed=1`, chat 15.

Body sha256 `b035bd3f5656d65a84a6d21c34f9e623212aa777db13a76d73ecb8e7e873fb9b`, **926,223 bytes**, 419 messages (418 history + 1 probe). Sampling block exactly as above.

History fingerprint:

```json
{
  "messages_kept": 418,
  "messages_total": 1936,
  "chars": 896668,
  "sha256": "95ddd1123a63639a7bf373d3c15209e308b0e84a48d9fc6e2c7bcbe489fbd04e",
  "first_role": "user",
  "last_role": "assistant"
}
```

Final (419th) message:

```json
{
  "role": "user",
  "content": "NOTE: Only provide the answer without any explanations. \nQuestion: Can you summarize how my experience with motivation, routine consistency, and resilience has developed over time through my fitness journey?"
}
```

What the model emitted — the prompt ends `…<|open|>message role="assistant"<|sep|><|open|>think<|sep|>`, and the raw 157-token completion is:

```
Yes — here's a concise summary based on your fitness journey so far:

Your motivation has grown from occasional inspiration into a more steady, self-driven commitment, strengthened by tracking progress, setting goals, and drawing inspiration from success stories and role models. Your routine consistency has improved through structured schedules, seasonal adjustments, and a clear effort to protect key habits like morning workouts, rest days, and flexibility work, even during holidays, travel, and work changes. Your resilience has become more evident through how you've handled setbacks, injury concerns, stress, and competing responsibilities—by adjusting rather than quitting, prioritizing recovery, and continuing to show up with intention. Overall, you've developed a more sustainable, adaptable, and confident approach to fitness and well-being.<|close|>response<|sep|><|close|>message<|sep|>
```

No `<|close|>think`. No `<|open|>response`. Sibling closers only.

</details>

<details>
<summary><b>Failing example — response BEFORE the fix</b></summary>

```json
{
  "id": "chatcmpl-26TcpFkUEXipsSNiSOD1uhTI-019fb4e9-556a-7672-85a9-bcfd02b50fef",
  "object": "chat.completion",
  "created": 1785446618,
  "model": "kimi-k3",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "reasoning_content": "Yes — here's a concise summary based on your fitness journey so far:\n\nYour motivation has grown from occasional inspiration into a more steady, self-driven commitment, strengthened by tracking progress, setting goals, and drawing inspiration from success stories and role models. Your routine consistency has improved through structured schedules, seasonal adjustments, and a clear effort to protect key habits like morning workouts, rest days, and flexibility work, even during holidays, travel, and work changes. Your resilience has become more evident through how you've handled setbacks, injury concerns, stress, and competing responsibilities—by adjusting rather than quitting, prioritizing recovery, and continuing to show up with intention. Overall, you've developed a more sustainable, adaptable, and confident approach to fitness and well-being.<|close|>response<|sep|><|close|>message<|sep|>"
      },
      "finish_reason": "stop",
      "matched_stop": 163586
    }
  ],
  "usage": {
    "prompt_tokens": 206722,
    "completion_tokens": 157,
    "total_tokens": 206879,
    "prompt_tokens_details": { "cached_tokens": 205824 }
  },
  "system_fingerprint": "default"
}
```

There is **no `content` key at all** — a client reading `choices[0].message.content` gets `None` — and the raw `<|close|>response<|sep|>` / `<|close|>message<|sep|>` markers are leaked verbatim to the caller.

</details>

<details>
<summary><b>Failing example — response AFTER the fix</b> (same request, same seed)</summary>

```json
{
  "id": "chatcmpl-xJWLUChGUTUsvf50L6RrKZXm-019fb4ec-decb-7fc0-86a2-7be217261150",
  "object": "chat.completion",
  "created": 1785446850,
  "model": "kimi-k3",
  "choices": [
    {
      "index": 0,
      "message": {
        "content": "Yes — here's a concise summary based on your fitness journey so far:\n\nYour motivation has grown from occasional inspiration into a more steady, self-driven commitment, strengthened by tracking progress, setting goals, and drawing inspiration from success stories and role models. Your routine consistency has improved through structured schedules, seasonal adjustments, and a clear effort to protect key habits like morning workouts, rest days, and flexibility work, even during holidays, travel, and work changes. Your resilience has become more evident through how you've handled setbacks, injury concerns, stress, and competing responsibilities—by adjusting rather than quitting, prioritizing recovery, and continuing to show up with intention. Overall, you've developed a more sustainable, adaptable, and confident approach to fitness and well-being.",
        "reasoning_content": null
      },
      "finish_reason": "stop",
      "matched_stop": 163586
    }
  ],
  "usage": {
    "prompt_tokens": 206722,
    "completion_tokens": 157,
    "total_tokens": 206879,
    "prompt_tokens_details": { "cached_tokens": 205824 }
  },
  "system_fingerprint": "default"
}
```

</details>

<details>
<summary><b>Why this is a clean A/B</b></summary>

Verified programmatically on both cases:

| Check | Result |
|---|---|
| Request sha256 identical before vs after | ✅ |
| `completion_tokens` identical — model output unchanged | ✅ 157 and 344 |
| Pre-fix `reasoning_content` ends with `<\|close\|>response<\|sep\|><\|close\|>message<\|sep\|>` | ✅ |
| `pre_fix_reasoning[:-47] == fixed_content` — exact string equality | ✅ |
| Pre-fix `content` | `None` |
| Fixed `reasoning_content` | `null` |
| Any `<\|…\|>` marker in fixed `content` | ✅ none |

47 bytes is exactly the length of the two stray closers. The model produced identical bytes; only the field they landed in changed.

**Determinism check.** Both cases were run with 2 repeats at concurrency 1. All four pre-fix runs misfiled with byte-identical `reasoning_content`; all four fixed runs returned byte-identical `content`.

</details>

<details>
<summary><b>Full results</b> — per-case A/B, 120-request grid, regression check</summary>

### Controlled per-case A/B (one request in flight, 2 repeats each)

| Case | seed | pre-fix `content` | pre-fix `reasoning` | fixed `content` | fixed `reasoning` |
|---|---|---|---|---|---|
| `summarization#0` | 2 | 0 B | 1676 B + markers | 1629 B | 0 B |
| `summarization#1` | 1 | 0 B | 901 B + markers | 854 B | 0 B |

### Full grid — 20 questions × 6 seeds = 120 requests per binary

| Binary | Misfiled | Errors | Marker leaks | Both fields empty |
|---|---|---|---|---|
| pre-fix `76d2cfe4…` | **6 / 120 (5.0 %)** | 0 | — | — |
| fixed `8a8bc56d…` | **0 / 120** | 0 | 0 | 0 |

The six pre-fix failures, all `finish_reason: "stop"`:

| Question | seed | content | reasoning | completion tokens |
|---|---|---|---|---|
| `summarization#1` | 1 | 0 B | 674 B | 117 |
| `summarization#0` | 2 | 0 B | 1731 B | 340 |
| `information_extraction#1` | 3 | 0 B | 788 B | 137 |
| `preference_following#0` | 4 | 0 B | 446 B | 132 |
| `preference_following#0` | 5 | 0 B | 815 B | 181 |
| `preference_following#1` | 6 | 0 B | 527 B | 157 |

On the fixed binary **all six of those `(question, seed)` keys** returned a populated `content` with `reasoning_content` empty — the model again skipped the think channel and the parser now handles it.

### No regression on the normal path

Shape of the 120 fixed-binary responses:

| Shape | Count | Interpretation |
|---|---|---|
| `content` + `reasoning_content` both populated | 106 | normal think-then-answer, unchanged |
| `content` only | 14 | model skipped the think channel — now correct |
| `reasoning_content` only | 0 | the bug; eliminated |
| both empty | 0 | — |
| any `<\|open\|>` / `<\|close\|>` / `<\|sep\|>` leaked into either field | 0 | wrapper stripping intact |

Reasoning traces on the normal path are still routed correctly and untruncated — `temporal_reasoning#1` seed 5 returned 38,161 B of reasoning plus 48 B of content, and `event_ordering#1` seed 6 returned 57,232 B of reasoning plus 738 B of content.

> The skip-think *rate* differs between the two grid runs (6/120 pre-fix vs 14/120 fixed) because the per-request `seed` does not pin batch composition under concurrency, so the two runs sampled differently. That is why the per-case A/B — one request in flight, byte-identical inputs — is the controlled comparison and the grid is the population-level check.

</details>

PR #2025, opened 2026-08-01
fix(reasoning): support DeepSeek V4 Flash 0731 effort encoding

key4ng:
### Problem

DeepSeek V4 Flash 0731 defines a three-level `low` / `high` / `max` reasoning-effort encoding, but SMG's native Rust renderer did not consistently receive the OpenAI-compatible Chat Completions and Responses API effort fields. DeepSeek V4 models were also not automatically selecting the appropriate reasoning parser, and regular Responses conversion could merge reasoning into output text or lose modern usage details on some completion paths.

### Solution

Map the official OpenAI request fields into the native DeepSeek V4 renderer while keeping an explicit native `chat_template_kwargs.reasoning_effort` override. Use the 0731 effort encoding for the DeepSeek V4 renderer, keep thinking disabled when no effort is requested, and automatically register the DeepSeek V4 reasoning parser. Preserve separated reasoning and modern Responses API usage details in non-streaming, streaming, and tool-loop completion paths.

## Changes

- Accept Chat Completions `reasoning_effort` and Responses API `reasoning.effort`.
- Map public effort values into DeepSeek V4 `low`, `high`, and `max` buckets.
- Respect an explicit native `chat_template_kwargs.reasoning_effort` value and reject invalid native buckets.
- Render the official 0731 high/max effort prefixes in the Rust DeepSeek V4 encoder.
- Keep thinking disabled by default when neither public nor native effort is supplied.
- Automatically select the DeepSeek V4 reasoning parser from model identity.
- Keep Chat and Responses rendering behavior in parity, including stateful Responses tool loops.
- Preserve cached-token and reasoning-token usage details in the modern Responses API shape.

## Test Plan

- `cargo test -p smg responses --lib` — 51 passed.
- `cargo test --workspace -- --skip dsml_one_shot_and_incremental_decode_match --skip test_router_with_tracing` — passed, including unit, integration, and doc tests.
- `cargo +nightly fmt --all -- --check` — passed.
- `pre-commit run --all-files` — every hook passed except Clippy, which could not start because the local macOS environment has no `pkg-config`/OpenCV installation.
- The excluded `dsml_one_shot_and_incremental_decode_match` is a pre-existing untracked debug probe requiring `DSV4_TOKENIZER`.
- The excluded upstream `test_router_with_tracing` was run twice independently and consistently received zero OTLP spans; this branch does not modify tracing or observability code.

Live B300 validation used the original `deepseek-ai/DeepSeek-V4-Flash` weights with TP2 on GPUs 4–5. Because the deployed SMG 1.9.0 image predates this patch, the benchmark injected the exact official 0731 prefixes that the new Rust unit tests verify:

| Prompt | Effort | Completion tokens | Reasoning chars | Finish |
| --- | ---: | ---: | ---: | --- |
| 12-coin balance puzzle | low | 4,472 | 10,270 | stop |
| 12-coin balance puzzle | high | 8,192 | 27,138 | length |
| 12-coin balance puzzle | max | 5,038 | 14,190 | stop |
| Decimal comparison | low | 186 | 312 | stop |
| Decimal comparison | high | 242 | 400 | stop |
| Decimal comparison | max | 170 | 300 | stop |

The original Flash weights clearly respond to the 0731 effort prompts. Reasoning length is task-dependent rather than strictly monotonic, so these prefixes should be understood as instructions, not hard token budgets.

<details>
<summary>Checklist</summary>

- [ ] `cargo clippy --all-targets --all-features -- -D warnings` passes (blocked locally by missing `pkg-config`/OpenCV)

</details>

### DIFF
--- a/crates/protocols/src/chat.rs
+++ b/crates/protocols/src/chat.rs
@@ -336,17 +336,14 @@ pub struct ChatCompletionRequest {
 ///
 /// This is the protocol-level interpretation of "does the caller want
 /// reasoning?" — independent of any model/template. `reasoning_effort` is a
-/// *level* (`"low"`/`"medium"`/`"high"`) plus the vendor-extension `"none"`.
-///
-/// Both `"none"` and `"minimal"` map to thinking OFF (`Some(false)`).
-/// `"minimal"` is treated as an off-signal deliberately: templates that expose
-/// only a boolean thinking toggle (GLM/Qwen3) cannot do "a little" reasoning,
-/// so the lowest OpenAI level is the closest available "do not reason".
-/// Level values return `None` — no opinion, defer to the template default or an
-/// explicit thinking kwarg.
+/// *level* plus `"none"`, which explicitly disables reasoning. Every known
+/// non-`none` OpenAI level requests some reasoning and therefore enables the
+/// model's thinking mode. Unknown scalar values remain pass-through data and
+/// do not implicitly toggle thinking.
 pub fn thinking_from_reasoning_effort(reasoning_effort: Option<&str>) -> Option<bool> {
     match reasoning_effort {
-        Some("none") | Some("minimal") => Some(false),
+        Some("none") => Some(false),
+        Some("minimal" | "low" | "medium" | "high" | "xhigh" | "max") => Some(true),
         _ => None,
     }
 }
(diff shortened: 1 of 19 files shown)


key4ng:
Live B300 parity validation completed against the original `DeepSeek-V4-Flash` model (TP2 gRPC backend), using PR commit `86b9a70f`.

Compared Chat Completions requests using:

```json
{"reasoning_effort":"low|max"}
```

and:

```json
{"chat_template_kwargs":{"reasoning_effort":"low|max"}}
```

Results with identical prompts and deterministic decoding:

| effort | request shape | prompt tokens | completion tokens | reasoning chars | answer chars |
|---|---|---:|---:|---:|---:|
| low | top-level | 21 | 125 | 255 | 75 |
| low | template kwargs | 21 | 125 | 255 | 75 |
| max | top-level | 113 | 102 | 157 | 107 |
| max | template kwargs | 113 | 102 | 157 | 107 |

For each effort, the two request shapes produced identical SHA-256 hashes for both `reasoning_content` and `content`.

The live test caught and fixed two separate gaps:

1. Native template effort selected the effort bucket but did not initially enable thinking.
2. After thinking was enabled, native template effort did not initially mark the response parser as starting inside reasoning, so reasoning leaked into `content`.

Both paths now use the same precedence: explicit `chat_template_kwargs.thinking` > native template effort > top-level OpenAI effort. The native-effort response behavior is gated specifically to the DeepSeek V4 renderer.

Focused verification: 19/19 DeepSeek renderer tests, 10/10 DeepSeek V4 encoder tests, 51/51 Responses tests, parser-helper tests, rustfmt, diff check, and tokenizer-library Clippy all pass.

PR #2035, opened 2026-08-03
fix(router): use independent P/D policies in gRPC PD selection

OftenDream:
## Summary
- gRPC PD `select_pd_pair` previously shared one policy instance for prefill and decode, so stateful policies like `round_robin` advanced the same counter twice per request and each role only hit every other worker.
- Use independent `get_prefill_policy` / `get_decode_policy` (same approach as the EPD path).
- Add a regression test showing shared vs independent round-robin coverage across two pools.

## Test plan
- [x] Unit test `test_independent_policies_cover_all_workers_across_two_pools`
- [x] Live PD 4P4D press: 40 requests with `--prefill-policy/--decode-policy round_robin` → each of 4 prefill and 4 decode workers got equal traffic

Co-authored-by: longcat-ia-team <312376714+longcat-ia-team@users.noreply.github.com>

Made with [Cursor](https://cursor.com)

### DIFF
--- a/model_gateway/src/policies/round_robin.rs
+++ b/model_gateway/src/policies/round_robin.rs
@@ -205,9 +205,7 @@ mod tests {
             let p = prefill_policy
                 .select_worker(&prefill_workers, &info)
                 .unwrap();
-            let d = decode_policy
-                .select_worker(&decode_workers, &info)
-                .unwrap();
+            let d = decode_policy.select_worker(&decode_workers, &info).unwrap();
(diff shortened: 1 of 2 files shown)


OftenDream:
Added an integration regression test in `worker_selection.rs` that exercises `WorkerSelectionStage::select_pd_pair` with 4 prefill + 4 decode workers and independent round-robin policies via `PolicyRegistry::get_prefill_policy` / `get_decode_policy` (40 selections → 10 each).

Issue #2031, opened 2026-08-02
SMG Roadmap 2026 H2


slin1237:
This tracking issue collects the six workstreams SMG is pursuing through the rest of 2026. Each links to a pillar epic with its own design discussion and sub-issue task list as they are opened.

### 1. Per-model protocol dialects

Open frontier models no longer speak one uniform OpenAI protocol. SMG will speak each model's dialect faithfully — first dialects: **Kimi K3** and **MiniMax M3**.

- A first-class **dialect layer**: per-model request directives, validation semantics, and model-authored sampling defaults — resolved once per request, no per-model hacks scattered through the core
- **Model-aware media handling**: per-model pixel budgets, image/video count limits, and video FPS constraints enforced in the gateway
- **Vendor-verified compatibility**: each model vendor's own public conformance checker runs in SMG's CI
- **SLA-grade observability**: TTFT percentiles by input-size tier, usage split by modality, and a theoretical prefix-cache hit-rate estimator

### 2. Open provider ecosystem

Any vendor selling tokens behind an OpenAI-compatible API becomes routable through SMG — with none of them compiled into the core.

- **Declarative provider manifests**: adding a bearer-auth vendor is configuration, not code
- **Out-of-process sidecar adapters** for vendors that need request signing (OCI, AWS, GCP, Azure) — SDK weight never enters the core binary
- **Router decomposition**: per-endpoint services over transport-agnostic pipeline stages, replacing today's monolithic router trait
- **"Supported = conformance-green"**: the provider list is backed by a CI conformance suite, with per-provider metrics
- The core stays a pure HTTP/gRPC gateway to self-hosted engines

### 3. RL rollout control plane

Engine-agnostic fleet control for RL rollout generation across mixed vLLM / SGLang / TensorRT-LLM / TokenSpeed fleets.

- **Fleet control verbs** (pause / resume / cancel-in-flight / memory / cache) with per-worker verification — no more half-paused fleets
- **Weight-version registry**: version-consistent routing and per-response version stamping on every engine
- **Rollout-grade data plane**, contract-tested per engine: token-in/token-out fidelity, sampled logprobs, abort-with-partials
- **Drop-in adapters** for slime, verl, TRL, SkyRL, and AReaL
- RFC incoming

### 4. API stability & governance

Interfaces stop breaking by accident — enforcement is mechanical, not cultural.

- **Breaking-change gates in CI**: semver checks on published crates, proto compatibility checks, OpenAPI spec diffs
[...]

PR #2061, opened 2026-08-05
feat(discovery): level-triggered informer reconcile with multi-port pods

slin1237:
## Description

Rewrites K8s service discovery from the raw-watcher design to the informer pattern (roadmap #2031 §5): `watcher().default_backoff().reflect(writer)` maintains a reflector `Store`, and a level-triggered reconciler diffs the desired worker set (from the store snapshot) against the worker registry, submitting `AddWorker`/`RemoveWorker` jobs for the gap. Every watch event (coalesced 1s) and a periodic `check_interval` tick trigger one pass; the periodic pass is now an in-memory diff with zero API calls.

**Deleted (the watcher-gap machinery):** the `tracked_pods` shadow set and its (name, uid) identity impls, the periodic full-LIST `reconcile_pods` sweep + panic-restart supervisor, both manual watcher restart/backoff loops, `.applied_objects()` + `deletion_timestamp` delete inference, same-name/new-UID StatefulSet eviction, and duplicate-event suppression. Net **−439 lines** while adding the feature below.

**Multi-port pods (new):** a pod running multiple engine servers lists them in the `smg.ai/worker-ports` annotation (e.g. `"8080,8081,8082,8083"`); each port becomes an independent worker with its own health/circuit-breaker/policy state. Absent annotation = single worker at `--service-discovery-port` (fully backward compatible). `sglang.ai/bootstrap-port` accepts an aligned list for PD pods (one value broadcasts; N values zip; mismatch warns and is ignored).

**Ownership correlation (fixes the scenario in #2031 comment by @neverCase):** every discovery-created worker carries `smg.ai/pod-name` / `smg.ai/pod-uid` labels through the label pipeline. The reconciler only ever adds/removes workers bearing the pod-uid label, so manually added workers are untouchable — and a worker registered by a zombie workflow after its pod died is now removed on the next pass instead of living forever. Failed registrations are retried by construction (registry lacks the worker → re-added next pass).

**Behavioral notes:**
- Terminating pods (deletion_timestamp set) leave the desired set immediately — draining starts at the beginning of the grace period, not the end.
- Same-URL pod restarts (stable IP, new uid) remove the stale worker (revision-guarded) and register the replacement — covering scheme flips and unready successors that an in-place Upsert cannot.
- The reconciler submits at most one in-flight job per URL (pending/processing job-queue statuses gate resubmission), so a configured drain window is never collapsed by a duplicate removal; removals additionally carry `expected_revision` as a stale-guard. The reflector prunes `managedFields` before the Store.
- Router/mesh discovery moves to the same self-healing stream with real `Event::Delete` (hard-deleted routers marked Down promptly) but deliberately no Store — mesh SWIM/CRDT owns convergence.

### Breaking change

> ⚠️ **Discovery metrics:** the `smg_discovery_deregistrations_total` label value `pod_deleted` and the `smg_discovery_registrations_total` label value `duplicate` are removed; all deregistrations now emit `reason="reconciled"`. The `smg_discovery_workers_discovered` gauge now counts desired **workers** (pod × port) instead of pods. Dashboards keying on the removed label values need updating.

## Test Plan

- `cargo +nightly fmt --all --check` clean; `cargo clippy --all-targets -- -D warnings` exit 0
- `cargo test -p smg`: **1930 passed, 0 failed** (21 binaries); `cargo test --workspace --exclude smg`: **2362 passed, 0 failed** (80 binaries)
- `cargo build -p smg-python` / `-p smg-golang` green; Python bindings expose `worker_ports_annotation` (both constructors, RouterArgs, docs)
- 20 new unit tests: port-list parsing, annotation fallback/dedup, bootstrap broadcast/zip/mismatch, desired-state (multi-port, terminating exclusion, unready retention, model-id override), reflector-store snapshot, registry diff (gone-pod removal, uid-change re-registration, DP-rank dedup, ownership scoping), spec label stamping, router event lifecycle

Stacked follow-ups: scripted fake-API-server integration tests (`model_gateway/tests/`), then kind-based real-cluster e2e.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -494,6 +494,7 @@ struct Router {
     multimodal_shm_min_bytes: Option<usize>,
     model_aliases: HashMap<String, String>,
     worker_startup_delay: u64,
+    worker_ports_annotation: String,
 }
 
 impl Router {
@@ -667,6 +668,7 @@ impl Router {
                 prefill_selector: self.prefill_selector.clone(),
                 decode_selector: self.decode_selector.clone(),
                 bootstrap_port_annotation: self.bootstrap_port_annotation.clone(),
+                worker_ports_annotation: self.worker_ports_annotation.clone(),
                 router_selector: self.router_selector.clone(),
                 router_mesh_port_annotation: "sglang.ai/mesh-port".to_string(),
                 model_id_source: self.model_id_from.clone(),
@@ -986,6 +988,7 @@ impl Router {
         multimodal_shm_min_bytes = None,
         model_aliases = HashMap::new(),
         worker_startup_delay = 0,
+        worker_ports_annotation = String::from("smg.ai/worker-ports"),
     ))]
     #[expect(clippy::too_many_arguments)]
     #[expect(
@@ -1114,6 +1117,7 @@ impl Router {
         multimodal_shm_min_bytes: Option<usize>,
         model_aliases: HashMap<String, String>,
         worker_startup_delay: u64,
+        worker_ports_annotation: String,
     ) -> PyResult<Self> {
         let mut all_urls = worker_urls.clone();
 
@@ -1256,6 +1260,7 @@ impl Router {
             multimodal_shm_min_bytes,
             model_aliases,
             worker_startup_delay,
+            worker_ports_annotation,
         })
     }
 
@@ -1294,6 +1299,7 @@ impl Router {
                 prefill_selector: self.prefill_selector.clone(),
                 decode_selector: self.decode_selector.clone(),
                 bootstrap_port_annotation: self.bootstrap_port_annotation.clone(),
+                worker_ports_annotation: self.worker_ports_annotation.clone(),
                 router_selector: self.router_selector.clone(),
                 router_mesh_port_annotation: "sglang.ai/mesh-port".to_string(),
                 model_id_source,
--- a/bindings/python/src/smg/router.py
+++ b/bindings/python/src/smg/router.py
@@ -220,6 +220,10 @@ class Router:
             context keys. Example: {'x-tenant-id': 'tenant_id'}. Default: {}
         bootstrap_port_annotation: Kubernetes annotation name for bootstrap port (PD
             mode). Default: 'sglang.ai/bootstrap-port'
+        worker_ports_annotation: Kubernetes annotation listing a pod's worker data
+            ports (comma-separated) for pods running multiple engine servers.
+            Absent = one worker at service_discovery_port. Default:
+            'smg.ai/worker-ports'
         request_timeout_secs: Request timeout in seconds. Default: 600
         max_concurrent_requests: Maximum number of concurrent requests allowed for
             rate limiting. Default: 256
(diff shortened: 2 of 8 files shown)


slin1237:
## Multi-agent correctness review — results and fixes

Ran a 16-agent review (7 lens-specific finders + adversarial verifiers who must trace each claim end-to-end through real code before upholding it; ~1.9M tokens). 14 raw findings → 5 confirmed, 2 refuted as pre-existing-and-improved, 5 low-severity capped unverified. All confirmed findings are fixed in the amended commit; every fix is mutation-tested where applicable.

### Confirmed and fixed

1. **[major] Duplicate removals collapsed the drain window** (found independently by 3 lenses, confirmed by 3 verifiers). The stateless reconciler re-submitted `RemoveWorker` mid-drain; status transitions don't bump revision so the `expected_revision` guard passed; the duplicate's `DrainWorkersStep` saw the worker already Draining, skipped the settle sleep, and removed immediately — cutting any configured drain (default 5s, up to 600s) to the inter-pass gap on essentially every rollout. **Fix:** the reconciler now submits at most one in-flight job per URL (job-queue `pending`/`processing` statuses gate resubmission; `completed`/`failed` don't block, preserving level-triggered retry). Enforced by a new integration test with a 2s drain against 300ms reconcile passes, anchored on the observed Draining transition — verified to FAIL with the gate removed (in #2062).

2. **[major] Python positional-ABI break**: `worker_ports_annotation` was inserted mid-list in the `_Router` signature, shifting ~70 positional parameters in violation of the file's documented append-only contract. **Fix:** moved to the tail of the struct fields, `#[pyo3(signature)]`, and `fn new` — kwargs callers (including `router.py`) unaffected.

3. **[major] Mesh-origin exclusion had zero coverage** (mutation-proven: deleting the `WorkerOrigin::Local` filter left the whole suite green, and the test named for it didn't exercise it). **Fix:** new test imports a pod-uid-labeled worker through the real `on_remote_worker_state` mesh path and asserts the reconciler never owns it — verified to fail with the filter deleted.

4. **[minor] Stale worker survived behind an unready same-URL replacement** and **[minor] scheme-flip stranded an unremovable sibling**: both stemmed from removal being keyed on URL presence alone. **Fix:** the desired state now records the owning pod uid per URL and removal triggers on uid mismatch too (revision-guarded), subsuming the old presence check. New unit tests cover both.

Also applied from the unverified batch: `managedFields` pruned before the reflector Store (they dominate pod object size), and the kind e2e's `wait_for_count 0` no longer passes if the smg process died (#2063).

### Refuted (for the record)

Two claims that a stale removal can deregister a same-URL replacement mid-drain were traced as **pre-existing shared-machinery behavior on main** — where the old discovery submitted removals with `expected_revision: None` (no guard at all) and never healed the loss; this branch strictly narrows the window and self-heals within one pass.

### Known remaining (accepted)

Discovery counters record job *submissions*, not outcomes — now bounded by the one-in-flight gate but still submission-semantics; PD mode with disjoint role selectors watches unfiltered (pre-existing scope, now cheaper with pruning).

Verification after fixes: fmt/clippy clean, `cargo test -p smg` 1932 passed / 0 failed, integration suite 14/14 (now speaking real watch protocol incl. per-object resourceVersions), Python bindings build green.

PR #2063, opened 2026-08-05
test(discovery): kind-based real-cluster e2e for multi-port discovery

slin1237:
Stacked on #2062 (which stacks on #2061).

## Description

Real-cluster coverage for the layers the fake-API-server tests can't reach: RBAC/kubeconfig auth, real watch bookmarks and relists, kubelet readiness transitions, and actual graceful/force deletion semantics.

- `e2e_test/kind_discovery/manifests.yaml` — two **hostNetwork** pods on a kind node: `multi-engine-0` runs four HTTP engine servers from one container (the exact multiple-servers-per-pod deployment shape) listed in `smg.ai/worker-ports: "28080,28081,28082,28083"`; `single-engine-0` has no annotation and exercises the `--service-discovery-port` fallback. The mock engine is a stdlib-only python script shipped via ConfigMap — no image builds.
- `e2e_test/kind_discovery/run.sh` — creates the cluster, waits for readiness, starts the `smg` binary on the runner host against the kind kubeconfig, then asserts via `GET /workers`: **5 workers** (4 + 1 fallback), all carrying `smg.ai/pod-uid`, exactly 4 labeled `smg.ai/pod-name=multi-engine-0`; graceful `kubectl delete` → 4 workers; `--grace-period=0 --force` → 0. Dumps the smg log tail and pod state on failure; always tears the cluster down.
- `.github/workflows/e2e-kind-discovery.yml` — **workflow_dispatch only** with a `runner` input defaulting to `ubuntu-latest` (GitHub-hosted has Docker, so this runs today); flip the input to the self-hosted pool once dind availability there is confirmed. Not a merge gate.

hostNetwork is what makes the host-side smg able to probe worker ports at the pod IP (= node container IP, routable on Linux runners) — and it also mirrors the real multi-engine GPU-pod deployment pattern.

## Validation

Runtime validation needs a Linux Docker host, so this is dispatch-validated on CI rather than locally (local Docker daemon unavailable, and macOS cannot route to kind node IPs regardless). Locally verified: `bash -n` on the script, workflow + manifest YAML parse, and the embedded multi-port engine server executed standalone — all four ports served `/health` and `/get_model_info` correctly. The full discovery pipeline itself is already covered end-to-end by the deterministic fake-API-server tests in #2062; this PR adds the real-cluster plumbing on top.

Suggested first run: `gh workflow run e2e-kind-discovery.yml` once this merges (or dispatch from this branch).

### DIFF
--- a/.github/workflows/e2e-kind-discovery.yml
+++ b/.github/workflows/e2e-kind-discovery.yml
@@ -1,0 +1,63 @@
+name: E2E Kind Discovery
+
+# Manual-only while runner Docker support (dind on the k8s runner pool) is
+# unconfirmed; GitHub-hosted ubuntu-latest works out of the box. Builds smg
+# in-job: pr-test-rust publishes no server binary artifact (and its
+# artifacts expire after a day), and its self-hosted runner image is not
+# glibc-matched to ubuntu-latest — sccache keeps rebuilds cheap instead.
+# Once dind on k8s-runner-cpu is confirmed, set the runner input and
+# artifact reuse becomes safe.
+on:
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: Runner label (needs Docker for the kind node container)
+        required: false
+        default: ubuntu-latest
+
+env:
+  RUSTC_WRAPPER: sccache
+  SCCACHE_GHA_ENABLED: "true"
+
+jobs:
+  kind-discovery:
+    runs-on: ${{ inputs.runner || 'ubuntu-latest' }}
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Setup Rust
+        uses: ./.github/actions/setup-rust
+        with:
+          sccache-version: 'v0.10.0'
+
+      - name: Install kind and kubectl
+        run: |
+          curl -Lo kind "https://kind.sigs.k8s.io/dl/v0.29.0/kind-linux-amd64"
+          chmod +x kind && sudo mv kind /usr/local/bin/
+          curl -Lo kubectl "https://dl.k8s.io/release/v1.33.0/bin/linux/amd64/kubectl"
+          chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+
+      - name: Build smg and mock-worker
+        run: |
+          source "$HOME/.cargo/env"
+          cargo build -p smg -p mock-worker
+
+      - name: Build mock-worker and gateway images
+        run: |
+          docker build -t smg-mock-worker:e2e \
+            -f e2e_test/kind_discovery/Dockerfile.mock-worker target/debug
+          docker build -t smg-gateway:e2e \
+            -f e2e_test/kind_discovery/Dockerfile.smg target/debug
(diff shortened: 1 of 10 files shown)


slin1237:
Validated end-to-end on a real cluster: temporarily added a branch push trigger (since workflow_dispatch only registers from the default branch), ran on GitHub-hosted ubuntu-latest, then removed the trigger and force-pushed the branch back to its single clean commit.

**Run: https://github.com/smg-project/smg/actions/runs/31047416285 — success.** Timeline from the log: kind cluster up in ~45s → both pods Ready → smg discovered **5 workers in ~4s** (4 annotated ports + fallback) → ownership-label assertions passed → graceful delete converged in ~6s → force delete converged in ~2s → PASS. Total e2e step ~1 minute; whole job ~16 minutes (dominated by the Rust build).

slin1237:
Expanded per review discussion and validated end-to-end on real clusters (temporary branch trigger, since workflow_dispatch only registers from the default branch; trigger removed and the branch squashed back to one commit).

**Final validated run: https://github.com/smg-project/smg/actions/runs/31075739665 — 16/16 passed in 3m42s of test time** (~20 min wall incl. Rust + image builds). Earlier iterations also caught and fixed a mock-fidelity bug worth noting: the ConfigMap engine answered `200 {}` on unknown paths, which silently defeated the gateway's `/model_info` → `/get_model_info` 404-fallback and registered workers under the unknown model — mocks now 404 unknown paths.

## What this suite now covers

**Host-gateway lifecycle (9):** multi-port registration + ownership labels, batch bursts, namespace scoping (decoy pod in a foreign namespace), `kubectl annotate` port-set edits, kubelet readiness flips (probe-driven NotReady removes nothing), gateway restart rebuild, pod recreation rotating the ownership uid, **data plane** (completions fan out across ≥2 discovered engines), graceful vs force deletion.

**mock-worker fleets (3):** one pod running **forty engines** from a single `smg.ai/worker-ports` annotation (URL set verified port-for-port, then unwound), gRPC engines registering with `connection_mode: grpc` via the dual-probe, and a PD prefill/decode mix asserting **per-engine aligned bootstrap ports** (29600→29700, 29601→29701).

**In-cluster operator journey (4):** smg deployed as a Deployment with **in-cluster ServiceAccount auth** and the minimal RBAC in `in_cluster.yaml` (doubles as reference manifests — first coverage of the `Client::try_default` in-cluster branch), discovering a plain no-hostNetwork multi-port engine Deployment over the pod network, driven with `kubectl scale` and `rollout restart`; a typo'd annotation falls back to exactly one worker and is diagnosable from gateway logs.

## Cluster report from the validated run

```
0.0s  initial deploy (2 pods x 4 engines)   [converged to 8]
1.0s  scale 2 -> 4 replicas                 [converged to 16]
6.0s  scale 4 -> 1 replicas                 [converged to 4]
6.0s  post-rollout convergence              [converged to 12]
12.0  rollout restart capacity floor        [never dipped below 12/12]
1.0s  typo deployment falls back to one worker
```

Registration converges in ~1s; removals in ~6s (5s drain settle + one pass); a rolling restart of the engine fleet held **full capacity throughout** (surge pods registered before old ones drained).

PR #2059, opened 2026-08-05
test(zmq): e2e harness support for direct-ZMQ workers (6/7)

slin1237:
### Problem

The e2e test suite has no way to stand up a gateway + worker in direct-ZMQ mode,
so the ZMQ backend cannot be exercised end-to-end.

### Solution

Add ZMQ support to the e2e harness: a connection-mode abstraction, headless
(`--skip-tokenizer-init`) worker command builders, and gateway/worker-pool wiring
plus fixtures/hooks so tests can select the ZMQ path. Python-only harness change,
no product code.

## Changes

- `e2e_test/infra/constants.py` — `ConnectionMode` (HTTP vs ZMQ).
- `e2e_test/infra/worker.py` — headless ZMQ worker command builders.
- `e2e_test/infra/{gateway.py,worker_pool.py,__init__.py}` — wire the connection
  mode through gateway launch and the worker pool.
- `e2e_test/fixtures/{hooks.py,setup_backend.py}` — fixtures/hooks to select and
  set up the ZMQ backend.

## Test Plan

- `python -m py_compile` on all changed modules.
- Consumed by the ZMQ e2e tests + CI lanes added in the next PR
  (`zmq/07-e2e-tests`); unit tests for these builders/hooks also land there.

<details>
<summary>Checklist</summary>

- [x] `cargo +nightly fmt` passes (no Rust changes)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (no Rust changes)
- [ ] (Optional) Documentation updated
- [ ] (Optional) Please join us on Slack [#sig-smg](https://slack.lightseek.org) to discuss, review, and merge PRs

</details>

---
**Stack** (split of the oversized #2041, merge bottom-up): 01-core-dispatch →
02-structured-outputs → 03-multimodal → 04-eos-forwarding → 05-harmony-stop-matcher
→ **06-e2e-infra (this PR)** → 07-e2e-tests. Review/merge after its parent
`zmq/05-harmony-stop-matcher`.

PR #2065, opened 2026-08-06
fix(zmq): correct transport-identity leaks in metrics, admin, and PD paths

slin1237:
## Motivation

An architecture audit of the direct-ZMQ stack (post #2054–#2059) found a set of spots where code assumes every gRPC-pipeline worker speaks gRPC — so ZMQ workers get miscounted, mislabeled in metrics, or hit panicking accessors. This is the small mechanical batch (Phase 0); the capability-model refactor that removes the underlying pattern follows after the stack lands.

## Modifications

- **Metrics**: PD/EPD worker-selection metrics and PD execution error records label each leg with the worker's actual connection mode (`as_metric_label()`) instead of hardcoded `grpc`.
- **Admin**: `flush_cache_all` partitions ZMQ workers out of the fan-out (they have no cache-flush RPC) instead of reporting them all as failed; counts and message updated.
- **Debug**: `GrpcRouter`'s PD/EPD debug output counts all gRPC-pipeline workers (gRPC + ZMQ) rather than only `ConnectionMode::Grpc`.
- **Robustness**: `inject_sglang_bootstrap_metadata` matches the proto variant with a `let-else` + warn instead of the panicking `as_sglang_mut` accessor.
- **Diagnostics**: EPD encode rejection for TokenSpeed-over-ZMQ names the transport limitation ("requires a gRPC TokenSpeed prefill worker") instead of claiming the engine is unsupported.
- **Contract doc**: the `ProtoStream::Zmq` seam now documents that every ZMQ engine emits vllm-shaped responses, so engine-specific response logic must key on `runtime_type()`, never the response variant.

## Checklist

- [x] Format your code: `cargo fmt`
- [x] Unit tests pass: `cargo test -p smg --lib` (1431 passed; the one failure, `middleware::metrics::distinct_ids_on_matched_route_do_not_grow_interner`, fails identically on clean main under the full parallel suite — pre-existing flake, passes in isolation)

### DIFF
--- a/crates/protocols/src/worker.rs
+++ b/crates/protocols/src/worker.rs
@@ -1136,6 +1136,10 @@ pub struct FlushCacheResult {
     pub http_workers: usize,
     #[serde(default)]
     pub grpc_workers: usize,
+    /// Workers skipped because their transport has no cache-flush RPC
+    /// (direct-ZMQ engines). Keeps `total = http + grpc + zmq` exact.
+    #[serde(default)]
+    pub zmq_workers: usize,
     pub message: String,
 }
 
@@ -1325,6 +1329,7 @@ impl IntoResponse for FlushCacheResult {
             "workers_flushed": self.successful.len(),
             "total_http_workers": self.http_workers,
             "total_grpc_workers": self.grpc_workers,
+            "total_zmq_workers_skipped": self.zmq_workers,
             "total_workers": self.total_workers
         });
 
--- a/model_gateway/src/routers/grpc/common/stages/encode.rs
+++ b/model_gateway/src/routers/grpc/common/stages/encode.rs
@@ -320,6 +320,15 @@ fn prepare_items(
             prefill: BackendClient::Grpc(GrpcClient::TokenSpeed(_)),
             ..
         } => prepare_tokenspeed_items(intermediate, workers),
+        // TokenSpeed supports EPD encode, but only over gRPC — name the
+        // transport, not the engine, so the error points at the real limit.
+        ClientSelection::Disaggregated {
+            prefill: prefill @ BackendClient::Zmq(_),
+            ..
+        } if prefill.runtime_type() == RuntimeType::TokenSpeed => Err(anyhow!(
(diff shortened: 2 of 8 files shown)


slin1237:
Addressed all review comments in eb76224e:

- **encode.rs (CodeRabbit, Important) — fixed.** Confirmed valid: `select_encode_prefill_decode_workers` admitted any `uses_grpc_pipeline()` worker into the encode pool, so a direct-ZMQ TokenSpeed encode worker could be assigned items and fail at dispatch (gRPC encoder RPC against an `ipc://` URL). The encode pool is now gated to `ConnectionMode::Grpc` at pool-build time, before assignment.
- **manager.rs all-ZMQ early return (CodeRabbit, Important) — fixed** as proposed: the empty-capable-worker message and log now report the skipped-ZMQ count.
- **FlushCacheResult invariant (Claude) — fixed** by adding a serde-defaulted `zmq_workers` field (struct + `total_zmq_workers_skipped` in the JSON body), restoring exact `total = http + grpc + zmq` arithmetic without breaking wire compatibility.
- **proto_wrapper is_vllm() dispatch (CodeRabbit, Important) — not a bug, skipping.** The `is_vllm()` checks at the cited sites are *semantic* branches: vLLM streams report per-chunk token deltas while SGLang/TRT-LLM report cumulative counts in Complete. The ZMQ TokenSpeed adapter deliberately normalizes to vLLM semantics (delta chunks via the shared `StreamState`, Complete held to the finish tick), so routing its chunks through the vLLM arm is correct — dispatching them by `runtime_type()` into the TokenSpeed arm would double-count/miscount tokens. This contract is documented at the `ProtoStream::Zmq` seam in this PR; typed response identity is planned for the follow-up capability refactor.
- **Test nitpicks — added**: `pd_bootstrap_injection_skips_non_sglang_requests` (no panic, request untouched), `pd_debug_counts_include_zmq_workers` (Debug counts include ZMQ PD workers), and `pd_leg_labels_reflect_each_legs_transport` (mixed-transport PD labels). The full failure-injection PD streaming harness asked for in request_execution.rs doesn't exist yet; the unit test covers the label-selection logic this PR changed, and the harness belongs to the planned streaming-test work item.

PR #2073, opened 2026-08-07
refactor(zmq): dedup engine scaffolding and move connect mechanics to the client layer

slin1237:
## Motivation

Phase 2 of the ZMQ architecture cleanup (follows #2065/#2068): remove the duplication blocks and the one layering inversion the audit identified. No behavior changes.

## Modifications (one commit each)

1. **Shared n>1 fan-out scaffolding** — `fan_out_requests`/`fan_out_tokenspeed_requests` were ~90% identical; `fan_out_n` owns the clone/suffix/collapse dance, each engine keeps only its sampling-tweak closure.
2. **Shared per-tick stream emission** — the twin ~50-line Chunk/Complete stanzas (finish tick carries tokens → emit Chunk, park Complete) collapse into `StreamState::emit_tick`; mappers contribute only finish reason, matched stop, and token ids.
3. **Harmony request building: one dispatch per (backend × kind)** — the ~330-line 7-arm match becomes `HarmonyBody` resolution + `build_harmony_proto`, with one shared error path. vLLM/TokenSpeed build via static translators, so one arm each now covers both gRPC and direct-ZMQ. The 5-arm stop-id injection becomes `ProtoGenerateRequest::extend_stop_token_ids`. Net −164 lines.
4. **ZMQ connect mechanics move to `zmq_client`** — handshake-port derivation (FNV pinned against serve.py, vectors moved with it), socket layout, ipc-dir prep, EOS resolution, and the handshake now live in `zmq_client::connect_for_worker`; `worker.rs` keeps a thin wrapper and stops reaching into the router layer for its transport. Also fixes the misleading EOS warning (connect-time miss falls back to tokenizer EOS at request time — it never meant 'runs to max_tokens').

## Checklist

- [x] `cargo fmt` clean; clippy `--all-targets` clean on changed files
- [x] `cargo test -p smg --lib`: 1435 passed (single failure is the pre-existing `middleware::metrics` interner flake, identical on clean main)

### DIFF
--- a/model_gateway/src/routers/grpc/harmony/stages/request_building.rs
+++ b/model_gateway/src/routers/grpc/harmony/stages/request_building.rs
@@ -123,9 +123,11 @@ impl PipelineStage for HarmonyRequestBuildingStage {
         // were already rejected by the request-id match above; the reject arms
         // here keep this exhaustive without a panic.)
         let body = match &ctx.input.request_type {
-            RequestType::Chat(request) => {
-                HarmonyBody::Chat(modified_request.as_deref().unwrap_or_else(|| request.as_ref()))
-            }
+            RequestType::Chat(request) => HarmonyBody::Chat(
+                modified_request
+                    .as_deref()
+                    .unwrap_or_else(|| request.as_ref()),
+            ),
             RequestType::Responses(request) => HarmonyBody::Responses(request.as_ref()),
             RequestType::Embedding(_) => {
                 return Err(error::bad_request(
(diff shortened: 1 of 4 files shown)


slin1237:
Review comments addressed in f7c1ec8d:

- **proto_wrapper attribute hijack (CodeRabbit, Important) — fixed.** Confirmed real and it was also the cause of this PR's CI lint failure: `extend_stop_token_ids` had been inserted between `as_sglang`'s doc + `#[expect(clippy::panic)]` and the fn, so the expect went unfulfilled and `as_sglang`'s panic went unshielded. Attachment restored; added a unit test for the new method (per-variant params, TRT-LLM field, missing-params no-op).
- **TokenSpeed `error` finish reason (CodeRabbit, Major) — fixed** with the suggested pre-`emit_tick` guard: engine-side failures now surface as a tonic error, mirroring the vLLM stream, instead of a normal Complete.
- **400 vs 500 for the impossible-runtime arm (Claude nit) — fixed** by typing the build error (`Request` → 400, `Wiring` → 500), restoring the pre-refactor semantics.
- **Identity `fail` closure (Claude nit) — removed.**
- **Dispatcher unit tests (CodeRabbit, Minor) — deferred with reason:** `build_harmony_proto`'s arms take connected engine clients (gRPC channels / completed ZMQ handshakes), so unit coverage needs the mock-engine harness per backend — meaningful but not a quick win. The dispatch paths are exercised live by the e2e chat lanes (including the ZMQ lane landing in the tests PR). Tracked for the test-coverage pass alongside the deferred streaming-harness items from #2057/#2058.

PR #2080, opened 2026-08-09
feat(tokenizer): identify DeepSeek-V4 0731 effort encoding per checkpoint

key4ng:
### Problem

The DeepSeek-V4-Flash-0731 refresh changed the reasoning-effort prompt encoding relative to the original checkpoint family, and the two cannot be told apart by `config.json` or `tokenizer_config.json`:

- **Original** (`DeepSeek-V4-Flash`, `-DSpark`, `-Pro`): accepts `high`/`max`; only `max` emits a prefix ("Reasoning Effort: Absolute maximum …"); `high` renders nothing.
- **0731**: levels shifted down one — `high` now emits the original's `max` text, `max` emits a new stronger text ("Reasoning Effort: Beyond maximum …"), and `low` (no prefix) is the default.

`architectures`/`model_type` are identical across the family, `tokenizer_config.json` is byte-identical, and the `dspark_*` config keys don't discriminate (`-DSpark` has them but ships the *original* encoding). vLLM sidesteps this by unconditionally applying 0731 semantics to every `DeepseekV4ForCausalLM` (commit `7743486190`), silently changing behavior for the three original-encoding checkpoints. SMG previously rendered only the original encoding and had **no registered reasoning parser** for DeepSeek V4 — the nightly BFCL leg pinned `deepseek_v31` with a `TODO(confirm)`, which mis-splits because the V4 template prefills `<think>` (the same failure class recorded for Kimi-K2.6 in `nightly-bfcl.yml`).

### Solution

Identify the encoding revision from the checkpoint itself: the only artifact that declares it is the `encoding/encoding_dsv4.py` the checkpoint ships. At renderer-detection time (where we already read the sibling `config.json`), fingerprint that file for the 0731-only `max` prompt text (the marker is derived from the ported prompt constant, so they cannot drift); when the repo was pruned, fall back to a `0731` marker in the model path (covers HF cache layouts like `models--deepseek-ai--DeepSeek-V4-Flash-0731/...`); default to the original encoding, which three of the four released checkpoints use.

The encoder renders the correct per-revision prefixes. A recognized native effort implies thinking mode (an explicit `thinking` toggle still wins); unrecognized values are ignored, preserving the existing template-owns-interpretation contract for the merged public `reasoning_effort` field. The tokenizer exposes its per-revision native effort values through the `Tokenizer` trait (defaulted, non-breaking), and request processing uses that single source of truth to arm reasoning separation consistently with the rendered prompt — for both `chat_template_kwargs.reasoning_effort` and the top-level field. A `deepseek_v4` reasoning parser is registered and auto-selected from the model name.

Scope note: strict validation of effort values and mapping OpenAI-level `reasoning_effort` / Responses `reasoning.effort` names into the native buckets require provenance (native kwarg vs merged public field) and come in a follow-up PR, along with the Responses API usage/reasoning preservation fixes (splitting up #2025).

## Changes

- `crates/tokenizer/src/encoders/deepseek_v4.rs`: `EffortEncoding { Original, V0731 }` with `parse_native`/`valid_native_values` and `detect_from_encoder_source`; `ReasoningEffort::Low`; revision-aware prefix rendering.
- `crates/tokenizer/src/huggingface.rs`: `Renderer::DeepseekV4(EffortEncoding)`; detection (shipped-encoder fingerprint → path fallback → original default); native effort implies thinking mode.
- `crates/tokenizer/src/traits.rs` + `cache/mod.rs`: `native_reasoning_effort_values` trait hook (defaults to empty; only the DeepSeek V4 renderer opts in).
- `model_gateway/src/routers/grpc/utils/parsers.rs`: arm reasoning parsing when the renderer will enter thinking mode from a native effort (kwargs entry wins over the top-level field, mirroring the template-kwargs merge).
- `crates/reasoning_parser/src/factory.rs`: register `deepseek_v4` parser and map `deepseek-v4`/`deepseek_v4` model names to it.

## Test Plan

- `cargo test -p llm-tokenizer` — 188 tests pass, covering per-revision prefix rendering, fingerprint-beats-dir-name, dir-name fallback, unrecognized-effort tolerance, native-effort-implies-thinking with explicit override, and encoder-source detection.
- `cargo test -p reasoning-parser` — 99 tests pass, including V4 parser auto-selection.
- `cargo test -p smg routers::grpc::utils::parsers` — thinking-precedence tiers and kwargs-vs-top-level arming.
- Byte-exact validation against the 0731 checkpoint's own golden encode vectors (`encoding/tests/test_{input,output}_{1..4}`): all 4 cases match.
- `cargo +nightly fmt --all -- --check` and `cargo clippy -p llm-tokenizer -p reasoning-parser -p smg --all-targets -- -D warnings` pass (workspace-wide `--all-features` clippy blocked locally by the pre-existing `llm-multimodal` pkg-config/OpenCV requirement, same as #2025).
- Live B300 validation (TP2 per arm, vLLM 0.26.0): BFCL A/B at parity with pure vLLM (+0.42pp overall, never worse per category; details in comments), full effort matrix verified through the SMG arm for both `chat_template_kwargs.reasoning_effort` and the top-level field.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (changed crates; workspace blocked locally by pkg-config/OpenCV)

</details>

### DIFF
--- a/crates/reasoning_parser/src/factory.rs
+++ b/crates/reasoning_parser/src/factory.rs
@@ -162,6 +162,21 @@ impl ParserFactory {
             Box::new(BaseReasoningParser::new(config).with_model_type("deepseek_v31".to_string()))
         });
 
+        // DeepSeek V4 prefills <think> in the generation prompt when thinking
+        // is enabled, so request processing marks the parser as already in
+        // reasoning (via think_in_prefill) before the first generated token
+        // arrives.
+        registry.register_parser("deepseek_v4", || {
+            let config = ParserConfig {
+                think_start_token: "<think>".to_string(),
+                think_end_token: "</think>".to_string(),
+                stream_reasoning: true,
+                max_buffer_size: DEFAULT_MAX_BUFFER_SIZE,
+                always_in_reasoning: false,
+            };
+            Box::new(BaseReasoningParser::new(config).with_model_type("deepseek_v4".to_string()))
+        });
+
         registry.register_parser("kimi_k25", || {
             let config = ParserConfig {
                 think_start_token: "<think>".to_string(),
@@ -188,6 +203,8 @@ impl ParserFactory {
         registry.register_parser("kimi_k3", || Box::new(KimiK3Parser::new()));
 
         registry.register_pattern("deepseek-r1", "deepseek_r1");
+        registry.register_pattern("deepseek-v4", "deepseek_v4");
+        registry.register_pattern("deepseek_v4", "deepseek_v4");
         registry.register_pattern("deepseek-v3.1", "deepseek_v31");
         registry.register_pattern("deepseek-v3-1", "deepseek_v31");
         registry.register_pattern("qwen3-thinking", "qwen3_thinking");
@@ -271,6 +288,37 @@ mod tests {
         let factory = ParserFactory::new();
         let parser = factory.create("deepseek-r1-distill");
         assert_eq!(parser.model_type(), "deepseek_r1");
+    }
+
+    #[test]
+    fn test_factory_auto_registers_deepseek_v4_models() {
+        let factory = ParserFactory::new();
+        assert!(factory.list_parsers().contains(&"deepseek_v4".to_string()));
+
+        for model in [
+            "deepseek-ai/DeepSeek-V4-Flash",
+            "deepseek-ai/DeepSeek-V4-Flash-0731",
+            "deepseek-ai/DeepSeek-V4-Flash-DSpark",
+            "deepseek-ai/DeepSeek-V4-Pro",
+            "DEEPSEEK_V4_LOCAL",
+        ] {
+            assert_eq!(
+                factory.create(model).model_type(),
+                "deepseek_v4",
+                "model: {model}"
+            );
+        }
+
+        // The prompt prefills <think>; request processing marks the parser as
+        // in-reasoning before the first generated token.
+        let mut parser = factory.create("deepseek-ai/DeepSeek-V4-Flash-0731");
+        assert!(!parser.is_in_reasoning());
+        parser.mark_reasoning_started();
+        let parsed = parser
+            .detect_and_parse_reasoning("analysis</think>answer")
+            .unwrap();
+        assert_eq!(parsed.reasoning_text, "analysis");
+        assert_eq!(parsed.normal_text, "answer");
     }
 
     #[test]
--- a/crates/reasoning_parser/src/factory.rs
+++ b/crates/reasoning_parser/src/factory.rs
@@ -162,19 +162,13 @@ impl ParserFactory {
             Box::new(BaseReasoningParser::new(config).with_model_type("deepseek_v31".to_string()))
         });
 
-        // DeepSeek V4 prefills <think> in the generation prompt when thinking
-        // is enabled, so request processing marks the parser as already in
-        // reasoning (via think_in_prefill) before the first generated token
-        // arrives.
+        // V4 prefills <think>; request processing arms the parser via
+        // mark_reasoning_started.
         registry.register_parser("deepseek_v4", || {
-            let config = ParserConfig {
-                think_start_token: "<think>".to_string(),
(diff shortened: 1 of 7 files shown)


key4ng:
Live B300 validation of this branch against `deepseek-ai/DeepSeek-V4-Flash-0731` (TP2 per arm, vLLM 0.26.0 engines; arm A = pure vLLM OpenAI server with `--tokenizer-mode deepseek_v4`, arm B = vLLM gRPC servicer behind the SMG gateway built from this branch, with **no** explicit `--tool-call-parser`/`--reasoning-parser` — both auto-selected):

# BFCL A/B — smg (candidate) vs vllm (baseline)

| category | n | vllm | smg | Δ (cand−base) |
|---|---|---|---|---|
| simple_python | 400 | 55.75 | 55.75 | +0.00 |
| irrelevance | 240 | 37.08 | 38.33 | +1.25 |
| multiple | 200 | 36.50 | 36.50 | +0.00 |
| **overall (unweighted)** | 840 | **43.11** | **43.53** | **+0.42** |
| **overall (weighted by n)** | 840 | **45.83** | **46.19** | **+0.36** |

_Scores are % accuracy (official BFCL, FC mode). Same model, engine, checkpoint and sampling on both arms — the only difference is the frontend, so Δ is attributable to the tokenization+parsing layer._

The prior nightly failure mode ("SMG arm emits non-terminating garbage, runs to the 360m job cap with nothing scored") did not reproduce: the full 840-case × 2-arm run completed in ~4.5 minutes with clean stops everywhere, and a 50-case `simple_python` pre-check parsed tool calls on 50/50 for both arms. (Absolute accuracy is identical-ish across arms and low in chat mode — that's a model/handler property, not a frontend delta.)

Native effort smoke tests through the SMG arm confirm the 0731 encoding detection and parser arming:

| `chat_template_kwargs.reasoning_effort` | prompt_tokens | reasoning separated | finish |
|---|---:|---|---|
| (none, chat mode) | 17 | n/a (no thinking) | stop |
| `low` | 22 | yes (`reasoning_content`) | stop |
| `high` | 101 ("Absolute maximum" prefix) | yes | stop |
| `max` | 114 ("Beyond maximum" prefix) | yes | stop |
| `turbo` | — | HTTP 400: `must be one of low, high, max for this DeepSeek V4 checkpoint` | — |

Follow-up for the nightly once this lands: re-pin the BFCL/tau2 deepseek-v4 leg to `-0731` and drop the `deepseek_v31` reasoning-parser override.

key4ng:
Simplification pass applied in `2a4a0701` (net −29 lines; comments trimmed, overlapping tests merged). Two behavior corrections found by the review and re-validated live on the B300:

1. **Unrecognized effort values are tolerated again** (superseding the table above): the merged public `reasoning_effort` field reaches the renderer's kwargs, so hard-rejecting unknown values broke top-level `"none"`/`"medium"` requests that main accepts. The renderer now ignores values outside the revision's set, matching the existing template-owns-interpretation contract; strict validation returns in the follow-up PR together with kwarg provenance.
2. **Top-level `reasoning_effort` now arms reasoning separation.** The arming tier previously only saw raw `chat_template_kwargs`, so `{"reasoning_effort":"high"}` rendered a thinking prompt but leaked reasoning into `content`. The tier now mirrors the kwargs merge (kwargs entry wins, else top-level field) against the tokenizer-owned `native_reasoning_effort_values` — no duplicated value list in the gateway.

Re-validated effort matrix through the SMG arm (all `finish=stop`, correct answers):

| request | prompt_tokens | reasoning separated |
|---|---:|---|
| `chat_template_kwargs: low / high / max` | 22 / 101 / 114 | yes (496 / 151 / 147 chars) |
| top-level `high` | 101 | yes (151 chars) |
| top-level `none` / `medium` / `turbo` | 22 | chat mode, no reasoning (matches main) |

BFCL `simple_python` 50-case check on the simplified build: 50/50 tool calls parsed.

slin1237 (approved):

PR #2081, opened 2026-08-10
chore(ci): bump TokenSpeed pin to pick up the headless grammar fix

slin1237:
## What

`TOKENSPEED_REF`: `0f686760` (2026-07-27) → `92728734` — a 150-commit fast-forward on upstream main (GitHub compare: ahead 150, behind 0).

## Why

The target is the merge of [lightseekorg/tokenspeed#1028](https://github.com/lightseekorg/tokenspeed/pull/1028): `GrammarManager` used to discard `--grammar-backend xgrammar` whenever `skip_tokenizer_init` was set — which the headless ZMQ launch forces — so every constrained request over the ZMQ direct-backend wire (`tool_choice=required`/`{function}`, `response_format=json_schema`, `regex`) was aborted at admission with zero tokens and surfaced to clients as an empty 200. This is the root cause of the 44 TokenSpeed failures on the #2060 ZMQ lane; #2079's launcher flags were necessary but could not take effect without this engine fix.

## Drift risk

Bounded and mostly pre-validated:
- The msgpack wire contracts (`io_struct.py`, `sampling_params.py`), the headless entrypoint, and the CLI are unchanged across the span that matters.
- The intermediate SHA `57b49061` already ran this repo's full tokenspeed gRPC e2e lane green (148 passed) on the #2060 branch.
- The remaining churn is scheduler C++/kernel work, which the lanes on this PR validate directly.

## Verification

The gRPC tokenspeed lanes on this PR are the gate. The ZMQ-lane proof (44 failures clearing) runs on #2060, which carries the same pin and rebases onto this after merge.

### DIFF
--- a/e2e_test/fixtures/hooks.py
+++ b/e2e_test/fixtures/hooks.py
@@ -108,6 +108,16 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
             reason = marker.kwargs.get("reason", f"Not supported on {current_runtime}")
             pytest.skip(f"Skipping for {current_runtime}: {reason}")
 
+    # TEMPORARY — remove with the TOKENSPEED_REF bump that picks up the fix
+    # for https://github.com/lightseekorg/tokenspeed/issues/1036: the pinned
+    # tokenspeed has no MXFP4-MoE-with-expert-bias kernel on sm 9.0 (the
+    # triton MoE rewrite dropped bias; marlin/flashinfer/gluon don't cover
+    # H100), so gpt-oss dies at model load on CI's H100 runners.
+    if current_runtime == "tokenspeed":
+        model_marker = resolve_class_marker(item, "model")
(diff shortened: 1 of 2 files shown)


slin1237:
Update: the first run of this bump surfaced an upstream regression — tokenspeed `92728734` has no MXFP4-MoE-with-expert-bias kernel on H100 (the triton MoE rewrite dropped bias; marlin/flashinfer/gluon don't cover sm 9.0), so `gpt-oss-20b` dies at model load on the H100 runners. Filed as lightseekorg/tokenspeed#1036 with a bounded fix sketch (the vendored marlin kernel already implements bias; only the export layer opts out).

Added a commit that narrowly skips gpt-oss on the tokenspeed runtime, tagged with that issue so the next pin bump deletes it. All other models are unaffected — the intermediate pin ran the full suite green minus this.

PR #2082, opened 2026-08-10
ci(bench): re-pin BFCL and tau2 deepseek-v4 legs to Flash-0731

key4ng:
### Problem

The nightly BFCL and tau2 `deepseek-v4` legs were un-pinned from `DeepSeek-V4-Flash-0731` back to the base checkpoint in #2070, because every `-0731` nightly burned the entire job producing non-terminating garbage on the SMG arm with nothing scored. The SMG arm also ran with a `deepseek_v31` reasoning-parser fallback (`TODO(confirm)`) — the mis-split failure class documented for Kimi-K2.6 in the same workflow — since no `deepseek_v4` reasoning parser existed.

### Solution

#2080 landed the 0731 effort encoding with per-checkpoint identification and registered a real `deepseek_v4` SMG reasoning parser, and validated the stack live on a B300 A/B against `DeepSeek-V4-Flash-0731` (TP2 per arm, vLLM 0.26.0 engines): SMG scored at parity with pure vLLM (+0.42pp unweighted overall across `simple_python`/`irrelevance`/`multiple`, never worse on any category), with clean stops throughout — the full 840-case × 2-arm run finished in ~4.5 minutes, where nightlies previously died at the 360m cap. Details: https://github.com/smg-project/smg/pull/2080#issuecomment-5230452525

Re-pin both legs to `-0731` and use the real parser.

## Changes

- `.github/workflows/nightly-bfcl.yml`: `deepseek-v4` leg model/handler → `deepseek-ai/DeepSeek-V4-Flash-0731`(-FC); `smg_reason` `deepseek_v31` → `deepseek_v4`; comments updated. `run_timeout` stays at 5400s until the first clean nightly score.
- `.github/workflows/nightly-tau2.yml`: same re-pin and parser change for the `deepseek-v4` leg.
- `scripts/bfcl/README.md`, `scripts/tau2/README.md`: matrix tables and the reasoning-parser-fallback footnote updated to match.

Note for the Blackwell runner: `-0731` weights (~156 GB) may need pre-staging at `/raid/models/deepseek-ai/DeepSeek-V4-Flash-0731`; otherwise the first run downloads them via the HF id.

## Test Plan

- `python3 -c "import yaml; yaml.safe_load(...)"` on both workflow files — parses clean.
- Config-only change; the serving stack itself was validated end-to-end on B300 in #2080 (BFCL A/B parity table + effort matrix posted there), including a 50/50 `simple_python` tool-call pre-check through the SMG arm with auto-selected parsers.
- First scheduled run (BFCL Mon/Wed/Fri, tau2 Tue/Thu/Sat, 07:17 UTC) will confirm on the CI runner; the exit-2 unscored-arm gate from #2070 fails the leg loudly if anything regresses.

<details>
<summary>Checklist</summary>

- [x] `cargo +nightly fmt` passes (no Rust changes)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (no Rust changes)

</details>

### DIFF
--- a/.github/workflows/nightly-bfcl.yml
+++ b/.github/workflows/nightly-bfcl.yml
@@ -161,16 +161,16 @@ jobs:
               # Concurrent TP=4 per arm (0-3 + 4-7).
               {"name": "deepseek-v4", "runner": "blackwell", "tp": 4,
                "gpu_a": "0,1,2,3", "gpu_b": "4,5,6,7",
-               # Base V4-Flash: the -0731 refresh is not supported yet.
-               "model": "deepseek-ai/DeepSeek-V4-Flash", "bfcl_model": "deepseek-ai/DeepSeek-V4-Flash-FC",
+               # -0731 encoding + deepseek_v4 SMG reasoning parser landed in #2080;
+               # B300 A/B validated at parity (simple_python/irrelevance/multiple).
+               "model": "deepseek-ai/DeepSeek-V4-Flash-0731", "bfcl_model": "deepseek-ai/DeepSeek-V4-Flash-0731-FC",
                "vllm_tool": "deepseek_v4", "vllm_reason": "deepseek_v4",
-               "smg_tool": "deepseek_v4",
-               "smg_reason": "deepseek_v31",  # TODO(confirm): no deepseek_v4 SMG reasoning parser yet
+               "smg_tool": "deepseek_v4", "smg_reason": "deepseek_v4",
                # --kv-cache-dtype fp8 + --block-size 256 REQUIRED: V4's FlashMLA fp8
                # kernel asserts fp8 kv-cache (else EngineCore dies at startup).
                "vllm_extra": "--tokenizer-mode deepseek_v4 --trust-remote-code --kv-cache-dtype fp8 --block-size 256",
                "gpu_mem": "0.90", "startup_timeout": "2400", "model_cache": "/raid/models",
-               # No healthy baseline yet; kept tight until the leg scores cleanly.
+               # Kept tight until the first clean nightly score on this leg.
                "run_timeout": "5400",
                "arm_mode": "concurrent", "max_model_len": "auto"},
               {"name": "minimax-m2.7", "runner": "blackwell", "tp": 4,
--- a/.github/workflows/nightly-tau2.yml
+++ b/.github/workflows/nightly-tau2.yml
@@ -166,10 +166,10 @@ jobs:
               # Concurrent TP=4 per arm (0-3 + 4-7); num_tasks-capped.
               {"name": "deepseek-v4", "runner": "blackwell", "tp": 4,
                "gpu_a": "0,1,2,3", "gpu_b": "4,5,6,7",
-               # Base V4-Flash: the -0731 refresh is not supported yet.
-               "model": "deepseek-ai/DeepSeek-V4-Flash",
+               # -0731 encoding + deepseek_v4 SMG reasoning parser landed in #2080.
+               "model": "deepseek-ai/DeepSeek-V4-Flash-0731",
                "vllm_tool": "deepseek_v4", "vllm_reason": "deepseek_v4",
-               "smg_tool": "deepseek_v4", "smg_reason": "deepseek_v31",
+               "smg_tool": "deepseek_v4", "smg_reason": "deepseek_v4",
                "vllm_extra": "--tokenizer-mode deepseek_v4 --trust-remote-code --kv-cache-dtype fp8 --block-size 256",
                "gpu_mem": "0.90", "startup_timeout": "2400", "model_cache": "/raid/models",
                "arm_mode": "concurrent", "max_model_len": "auto", "num_tasks": 30,
--- a/scripts/bfcl/README.md
+++ b/scripts/bfcl/README.md
@@ -64,7 +64,7 @@ Key env knobs for `launch_arm.sh`: `BFCL_GPU` (CUDA_VISIBLE_DEVICES, e.g. `0,1`)
 |---|---|---|---|---|
 | Qwen3.6-27B (`qwen3.6`) | `4-gpu-h100` | 2 | `qwen3_xml` / `qwen3` | `qwen_xml` / `qwen3` |
 | gpt-oss-120b (`gpt-oss`) | `4-gpu-h100` | 2 | `openai` / — | _(none — SMG auto-routes harmony)_ / — |
-| DeepSeek-V4-Flash (`deepseek-v4`) | `blackwell` | 4 | `deepseek_v4` / `deepseek_v4` (+`--tokenizer-mode deepseek_v4 --trust-remote-code`) | `deepseek_v4` / `deepseek_v31`† |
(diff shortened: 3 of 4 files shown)


key4ng:
Investigation of the dispatch run [31361641789](https://github.com/smg-project/smg/actions/runs/31361641789/job/93373488079) (arm times 59s vs 65min, SMG arm scored 0.36%):

**The SMG frontend is not the cause.** On that run, per-case input token counts were identical across arms for all 399 comparable cases (SMG rendered/tokenized exactly what vLLM did), but the SMG arm's *generations* were corrupted — median 693 output tokens of token-salad (max 57K, one 20-min timeout case), with the same nonsense motifs repeating across unrelated requests (engine-state/KV corruption signature). Only the first request was near-clean.

**Full CI-fidelity reproduction on a B300 is clean.** Using the actual CI-built wheel from this run's artifact, the real `scripts/bfcl/launch_arm.sh`, TP4 per arm, `max-model-len auto` (1M), fp8 KV + block 256, `--tokenizer-mode deepseek_v4` on both engines, `VLLM_USE_BREAKABLE_CUDAGRAPH=1` exported (see below), same vLLM 0.26.0, same 550-case × 16-thread load:

| category | n | vllm | smg | Δ |
|---|---|---|---|---|
| simple_python | 400 | 55.25 | 55.75 | +0.50 |
| simple_java | 100 | 3.00 | 3.00 | +0.00 |
| simple_javascript | 50 | 66.00 | 68.00 | +2.00 |
| **overall (weighted)** | 550 | **46.73** | **47.27** | **+0.55** |

The vLLM-arm scores match the CI run's vLLM arm almost exactly (46.73 weighted, identical), confirming the model/engine/eval pipeline is equivalent — the only divergence is the SMG arm's corrupted generations **on the Blackwell runner specifically**. Individually tested and exonerated along the way: wheel-vs-binary, explicit-vs-auto parser flags, tokenizer-mode on the gRPC engine, breakable-cudagraph env, TP4, 1M context, sustained 16-thread load.

**Runner findings/suggestions:**
1. `VLLM_USE_BREAKABLE_CUDAGRAPH` is set machine-level on the Blackwell runner (it appears nowhere in this repo) and silently disables torch.compile for every leg — it should be removed from the runner env, and the env audited for other strays.
2. The garbage failure mode is per-request-path-asymmetric engine corruption (raw-token gRPC path affected, API-server path not) that manifests on the B200 runner but not on a B300 under identical software — likely kernel-selection (sm_100 vs sm_103) or additional runner state. Suggested bisect on the runner: add `--enforce-eager`, then `--no-enable-prefix-caching` to the leg's `vllm_extra`.
3. The workflow currently uploads arm logs on failure only; this run was "green" (the A/B gate is informational), so the arm-B engine log was lost. Worth switching to always-upload for this leg until it scores cleanly.

Note this also re-frames the Jul-31 nightlies: the "non-terminating garbage" that got the leg un-pinned in #2070 predates #2080 and shows the same signature — the 0731 frontend support was likely never the culprit.

key4ng:
Investigated directly **on the B200 CI runner** (`moirai-b200`). This corrects two claims in my previous comment and substantially changes the conclusion.

### Corrections

1. **`VLLM_USE_BREAKABLE_CUDAGRAPH` is not a stray runner variable.** vLLM sets it itself: `vllm/config/vllm.py:1179-1199` auto-enables it for a fixed architecture list that includes `DeepseekV4ForCausalLM` ("For model classes that don't carry `@support_torch_compile` — the breakable cudagraph is the supported PIECEWISE path"). It is by design and should be left alone. I withdraw the "remove it from the runner env" suggestion.
2. **`CUDA_HOME` was set during the failing run** (`/usr/local/cuda`, visible in the job env block). It comes from the runner environment rather than the workflow file, so my "the workflow pins no CUDA_HOME" note was misleading about the effective state.

### The B200-vs-B300 hypothesis was wrong

I reproduced the leg on the failing runner itself, using that runner's own venv — **byte-identical package versions to the failing run** (`vllm 0.26.0`, `tilelang 0.1.9`, `flashinfer-python 0.6.14`, `torch 2.11.0+cu130`, `apache-tvm-ffi 0.1.10`), the CI-built `smg 1.9.0` wheel, the real `launch_arm.sh`, TP4 per arm on the same 0-3 / 4-7 split, `max-model-len auto`, fp8 KV + block 256, same `CUDA_HOME`, same 550-case × 16-thread load.

Four consecutive A/B runs on that setup:

| run | vllm | smg | Δ | smg `output_tokens` p50 / max | smg text-only |
|---|---|---|---|---|---|
| 1 | 46.91 | 46.36 | −0.55 | 86 / 326 | 0 |
| 2 | 46.91 | 47.27 | +0.36 | 86 / 313 | 0 |
| 3 | 46.36 | 46.36 | +0.00 | 86 / 326 | 0 |
| 4 | 47.27 | **0.00** | **−47.27** | — | 400 |

So the leg is **not** systematically broken on B200 silicon — three clean runs at parity, then a failure. **The failure is a non-deterministic flake** (~1 in 4 here), which also explains why it was never reproducible off-runner.

### What the failure looks like when caught live

Run 4 was captured while still in the bad state:

- All 400 SMG-arm results were `Error during inference: Connection error.` — the harness could not reach the gateway.
- The **engine was healthy throughout**: gRPC `HealthCheck ... healthy=True` every 60s, GPUs idle at 0%, no errors in its log.
- The **gateway process was alive and still `LISTEN`ing** on its port, no panic, no fd leak (12 fds, 131 threads / 128 idle in `futex_wait`), and its background tasks kept ticking — but it served **no HTTP requests for ~18 minutes** (last request 07:35, still unresponsive at 07:53).
- It then **recovered on its own** (HTTP 200 again), and a follow-up 200-request / 16-concurrent stress against it was perfectly healthy: 12.9 s wall, 0 errors, all tool calls parsed, `completion_tokens` p50 = 88 — *faster* than the vLLM arm's 35 s on the identical stress.

So the SMG arm intermittently stops serving HTTP for minutes at a time while its engine stays healthy, then recovers. That is consistent with the original failing run's latency profile (p90 555 s, max 634 s) — that run was the same stall, partial rather than total. The garbage-token generations seen there are not yet explained by this and I have not reproduced them; a longer soak with automatic state capture is running now.

### Separate finding: the leg's startup depends on an undeclared runner variable

With `CUDA_HOME` unset, the DeepSeek-V4 tilelang JIT resolves `nvcc` to the pip `nvidia/cu13` toolchain (13.3) and compiles against incompatible headers, so **every** DSv4 arm dies at startup:

```
nvidia/cu13/include/cccl/cuda/std/__cccl/cuda_toolkit.h:41:8: error: "CUDA compiler and CUDA toolkit headers are incompatible"
Command: .../nvidia/cu13/bin/nvcc --cubin -O3 -lineinfo -arch=sm_100a ...
```

It works today only because the runner happens to export `CUDA_HOME=/usr/local/cuda` (system CUDA 13.0). Worth pinning explicitly in the workflow so the leg doesn't silently die if a runner is rebuilt.

### Recommendations (unchanged in substance, now better justified)

For a flake of this shape, detection matters more than prevention:

1. **Always upload arm logs for this leg**, not just on failure — run 4 here would have been undiagnosable from the job log alone, exactly as the original run was.
2. **Fail-fast canary**: probe each arm with ~20 cases before the full run and abort with exit 2 if tool-call parsing collapses, instead of burning the job.
3. **Stall/runaway detection** in `run_ab.py`: flag when an arm's per-case latency or `output_token_count` median diverges from the baseline arm by a large factor.

I'll follow up with whatever the soak captures.

PR #2093, opened 2026-08-11
fix(observability): bound client-controlled metric label cardinality

pallasathena92:
### Problem

Router and MCP metrics interned the raw, **client-supplied** `model` string (and model-generated MCP `tool_name`) into a never-evicted global interner and minted a Prometheus series per distinct value — **before** any validation against the worker registry. A client sending a flood of unique `model` values (random UUIDs) drives unbounded interner + series growth: RSS growth, larger scrape payloads, slower `/metrics` — unrecoverable without a restart. This violates the file's own invariant ("never intern user-controlled input"). Audit F002 / F038 / F085 (model) and F039 / F040 (MCP tool name), re-verified against `main`.

### Solution

Route the `model` and MCP `tool_name` labels through a bounded interner (`intern_bounded_label`): distinct values are interned up to a hard cap (1024 each), after which further distinct values collapse to a shared `other` sentinel. Because admitted values are never evicted and re-admitted, **no new series is minted past the cap** — an LRU would keep churning series in the recorder, which is why a cap+sentinel (gating at the label source) is used instead. Worker URLs, normalized paths, and error codes keep the unbounded interner — server-controlled and low-cardinality.

## Changes

- `model_gateway/src/observability/metrics.rs`: add `intern_bounded_label` + dedicated `MODEL_LABELS`/`TOOL_LABELS` maps, sentinel, caps; reroute all `model_id`/`tool_name` interning; add a cap+sentinel unit test.

## Test Plan

- `cargo test -p smg --lib observability::metrics` — 32 passed (1 new: cap + sentinel + no map growth past cap)
- `cargo clippy -p smg --all-targets -- -D warnings` — clean (`--all-features` needs opencv locally; covered by CI)
- `cargo +nightly fmt -p smg -- --check` — clean

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (via CI; `--all-features` needs opencv locally)

</details>

### DIFF
--- a/model_gateway/src/observability/metrics.rs
+++ b/model_gateway/src/observability/metrics.rs
@@ -38,6 +38,54 @@ pub(crate) fn interner_size() -> usize {
     STRING_INTERNER.len()
 }
 
+/// Sentinel substituted once a bounded interner reaches its cap, so a flood of
+/// client-controlled label values (model IDs, MCP tool names) cannot mint an
+/// unbounded number of interner entries or Prometheus series.
+const BOUNDED_LABEL_SENTINEL: &str = "other";
+
+/// Max distinct client-supplied model labels retained before collapsing to the
+/// sentinel. A gateway fronts far fewer real models than this; the cap only bites
+/// on adversarial or unvalidated input.
+const MAX_MODEL_LABELS: usize = 1024;
+
+/// Max distinct client/model-controlled MCP tool-name labels.
+const MAX_TOOL_LABELS: usize = 1024;
+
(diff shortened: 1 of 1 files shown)


pallasathena92:
Addressed the out-of-diff note in 47ab481b — dropped `model IDs` from the `STRING_INTERNER` comment and pointed client/model-controlled labels at the bounded interner, so a future contributor isn't invited to reopen the cardinality hole.

PR #2097, opened 2026-08-11
fix(routers): bound SSE response-body channels for backpressure

pallasathena92:
### Problem

Every SSE / streaming HTTP response body was fed by an **unbounded** `mpsc` channel. The backend read loop pushed frames at full engine speed while hyper drained the body only at the client's TCP rate, so a slow (or deliberately slow-reading) client made the gateway **buffer the entire response in RAM** — an unbounded-memory / slowloris OOM vector, quadratic on `/generate`. Audit finding F004 and its ~14 siblings, re-verified against `main`.

### Solution

Introduce a shared bounded SSE channel in `common/sse.rs` (`SSE_CHANNEL_BUFFER = 32`, mirroring the existing `STREAM_RELAY_BUFFER`; `SseSender`/`SseReceiver`; `sse_channel()`), route every response-body channel through it with **await-on-send** so a slow client applies backpressure to the backend read loop instead of growing memory. `build_sse_response` takes the bounded receiver + `ReceiverStream`. Producers that write body frames became `async` and await their sends; the five gRPC streaming entry points and the emitter/helper chains thread `.await` up to their async `execute()` callers. `send_error_sse` became `async`.

## Changes

- `common/sse.rs`: shared bounded channel API + `build_sse_response`.
- gRPC (chat / generate / messages / completion / responses + harmony), OpenAI responses, HTTP + PD relay paths: bounded channels, awaited sends, `ReceiverStream`; producer methods made `async` with `.await` threaded to stage callers. 17 files.

Note: parsers on these paths are per-request (not pooled), so holding a request's own async parser guard across its awaited send only backpressures that request — no cross-request stall and no sync-guard-across-`.await`.

## Test Plan

- `cargo test -p smg --lib` — 1448 passed, 0 failed
- `cargo clippy -p smg --all-targets -- -D warnings` — clean (`--all-features` needs opencv locally; covered by CI)
- `cargo +nightly fmt -p smg -- --check` — clean

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (via CI; `--all-features` needs opencv locally)

</details>

### DIFF
--- a/model_gateway/src/routers/common/sse.rs
+++ b/model_gateway/src/routers/common/sse.rs
@@ -24,7 +24,7 @@ use http::{
 };
 use serde::Serialize;
 use tokio::sync::mpsc;
-use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_stream::wrappers::ReceiverStream;
 
 // ============================================================================
 // SseEncoder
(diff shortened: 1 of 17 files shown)


pallasathena92:
Thanks — two notes:

1. **harmony/streaming.rs disconnect propagation (`send_event_best_effort`)** — this is a real but **pre-existing** issue (the Responses decode loop keeping the engine/MCP loop alive after client disconnect; it matches audit findings F160/F162). This PR is a purely mechanical unbounded→bounded backpressure conversion and deliberately preserves the existing best-effort semantics; switching those sites to propagate disconnects is a separate behavioral concern that deserves its own PR and tests (per CONTRIBUTING's "one concern per PR"). Tracking it as a follow-up.

2. **Duplicate `build_sse_response`** — noted; the gRPC-side and `common/sse` versions differ slightly in headers today, so I'd rather dedup them in a focused cleanup than expand this PR's footprint. Follow-up.

pallasathena92:
Correction on point 2: I re-checked and the two `build_sse_response` are functionally **equivalent** (same status + `text/event-stream` + no-cache/keep-alive headers; the only difference is header constants vs string literals). So it's a clean dedup — I'll re-export the `common/sse` one from the gRPC module in a small focused follow-up rather than expand this backpressure PR.

XinyueZhang369 (approved):

slin1237 (approved):

PR #2122, opened 2026-08-13
fix(grpc): deliver tool calls when a JSON-schema tool constraint meets a thinking model

key4ng:
### Problem

With `tool_choice: "required"` (or a named function) and a tool parser that has no structural tag, the gateway sends the engine a JSON-schema grammar constraint, so the completion is pure JSON — no think tokens can appear. But for thinking models (e.g. Qwen3, whose `enable_thinking` template toggle defaults on), the gRPC pipeline also pre-armed the reasoning parser via `mark_reasoning_started()`. The armed parser never saw a think-end token and classified the entire grammar-forced tool payload as truncated reasoning:

- `tool_calls: null`, `content: null`
- the tool-call JSON dumped into `reasoning_content`
- `finish_reason: "stop"` instead of `"tool_calls"`

Broken on both the OpenAI chat and Anthropic messages surfaces, streaming and non-streaming. Any thinking-toggle template family is affected, not just Qwen.

A compounding bug: the chat-template detector flagged Qwen3 as *think-in-prefill* because its generation prompt contains `<think>\n\n</think>` — a **closed** filler emitted only when thinking is **disabled** — so the pre-arm survived even where it never should have applied.

### Solution

- **Gate the pre-arm** (`utils::should_start_in_reasoning`): under a JSON-schema tool constraint, only genuine think-in-prefill templates keep the pre-arm (their completions really start mid-reasoning on thinking-aware engines, e.g. sglang honoring `require_reasoning`). Toggle templates emit an explicit `<think>` whenever reasoning does happen, so the un-armed parser still catches it.
- **Non-streaming recovery** (`utils::split_reasoning_result`): under the constraint an all-reasoning parse is impossible — hand the original text back to the tool stage. Covers think-in-prefill templates on engines that enforce the grammar from the first token (vLLM).
- **Streaming recovery**: at end of stream, if a still-armed parser consumed the payload (no tool calls emitted, parser still in reasoning), parse the buffered raw text and emit the tool-call deltas so `tool_calls` and the finish reason are still delivered. Applied to both chat (per-index, `n>1`-safe) and messages streaming.
- **Detector fix**: `think_in_prefill` now requires an *unclosed* `<think>` in the `add_generation_prompt` body.

Requests without a JSON-schema tool constraint (`auto`, `none`, no tools) take exactly the same code path as before. Parser crates are untouched — the wrong decisions lived in the gateway pipeline.

## Changes

- `model_gateway/src/routers/grpc/utils/parsers.rs`: new `should_start_in_reasoning` / `split_reasoning_result` helpers + unit tests.
- `model_gateway/src/routers/grpc/regular/processor.rs`: chat + messages non-streaming paths use gated arming and the recovery split; `used_json_schema` hoisted above reasoning parsing.
- `model_gateway/src/routers/grpc/regular/streaming.rs`: chat + messages streaming paths use gated arming; end-of-stream JSON-schema constraint recovery.
- `crates/tokenizer/src/chat_template.rs`: think-in-prefill detection requires an unclosed `<think>`; unit test for the Qwen3-style closed filler vs open-tag templates.
- `e2e_test/chat_completions/test_function_calling.py`: new `TestToolChoiceRequiredThinking` (streaming + non-streaming) on a thinking model — existing required-mode coverage only used a non-thinking Qwen, which is why this was never caught.

## Test Plan

Reproduced and validated on a B300 node: vLLM gRPC backend (`vllm.entrypoints.grpc_server`), `Qwen/Qwen3-0.6B`, gateway `--tool-call-parser qwen --reasoning-parser qwen3`, same request matrix before/after on this branch's base.

Request: `get_weather` tool, "What is the weather in Paris right now?", `max_tokens: 512`.

| Case | Before (main) | After (this PR) |
|---|---|---|
| `required`, non-streaming | `finish: "stop"`, `tool_calls: null`, JSON in `reasoning_content` | `tool_calls: [get_weather {"city":"Paris"}]`, `finish: "tool_calls"`, no reasoning |
| `required`, streaming | payload streamed as `reasoning_content` deltas, no tool deltas | tool-call deltas, `finish: "tool_calls"`, zero reasoning leak |
| named function | params object in `reasoning_content` | correct `tool_calls`, `finish: "tool_calls"` |
| Messages `tool_choice: {type: "any"}` (+ streaming) | same failure mode | `tool_use` block with input, `stop_reason: "tool_use"` |
| `auto` (control) | works | unchanged — reasoning separated, tool call parsed |
| `required` + `enable_thinking: false` (control) | works | unchanged |
| plain chat, thinking on (control) | works | unchanged — content + reasoning separated |

Unit tests: `cargo test -p smg --lib routers::grpc::utils::parsers`, `cargo test -p llm-tokenizer --lib chat_template`. Full workspace `cargo test` passes.

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/crates/tokenizer/src/chat_template.rs
+++ b/crates/tokenizer/src/chat_template.rs
@@ -1157,13 +1157,13 @@ mod tests {
         // Qwen3-style: the generation prompt injects a CLOSED empty think
         // block only when thinking is disabled — completions never start
         // mid-reasoning, so this must not count as think-in-prefill.
-        let qwen3_style = r#"
+        let qwen3_style = r"
 {%- if add_generation_prompt %}
     {{- '<|im_start|>assistant\n' }}
     {%- if enable_thinking is defined and enable_thinking is false %}
         {{- '<think>\n\n</think>\n\n' }}
     {%- endif %}
-{%- endif %}"#;
+{%- endif %}";
         let (_, think_in_prefill) = detect_all_with_ast(qwen3_style);
         assert!(
             !think_in_prefill,
@@ -1172,10 +1172,10 @@ mod tests {
 
         // Thinking-only style: the generation prompt ends with an OPEN
         // <think>, so completions genuinely start mid-reasoning.
-        let thinking_style = r#"
+        let thinking_style = r"
 {%- if add_generation_prompt %}
     {{- '<|im_start|>assistant\n<think>\n' }}
-{%- endif %}"#;
+{%- endif %}";
         let (_, think_in_prefill) = detect_all_with_ast(thinking_style);
         assert!(
             think_in_prefill,
--- a/crates/tokenizer/src/chat_template.rs
+++ b/crates/tokenizer/src/chat_template.rs
@@ -246,13 +246,21 @@ impl<'a> Detector<'a> {
     }
 
     /// Check if a list of statements contains `<think>` in EmitRaw or string constants.
+    /// `<think>` not closed by a later `</think>` in the same literal — an
+    /// open tag leaves the completion mid-reasoning, a closed pair (e.g.
+    /// Qwen3's thinking-off `<think>\n\n</think>` filler) does not.
+    fn str_has_open_think_tag(s: &str) -> bool {
+        s.rfind("<think>")
+            .is_some_and(|idx| !s[idx..].contains("</think>"))
+    }
+
     fn body_has_think_tag(stmts: &[Stmt]) -> bool {
         for stmt in stmts {
             match stmt {
-                Stmt::EmitRaw(raw) if raw.raw.contains("<think>") => return true,
+                Stmt::EmitRaw(raw) if Self::str_has_open_think_tag(raw.raw) => return true,
                 Stmt::EmitExpr(e) => {
                     if let Expr::Const(c) = &e.expr {
-                        if c.value.as_str().is_some_and(|s| s.contains("<think>")) {
+                        if c.value.as_str().is_some_and(Self::str_has_open_think_tag) {
                             return true;
                         }
                     }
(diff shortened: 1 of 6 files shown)


key4ng:
## Additional validation: SGLang gRPC backend

Ran the same before/after matrix against an SGLang worker to confirm the fix introduces no regression on the SGLang path (`lmsysorg/sglang:latest`, `--grpc-mode`, `Qwen/Qwen3-0.6B`, B300 node; gateway from this branch with `--tool-call-parser qwen --reasoning-parser qwen3`):

| Case | Result |
|---|---|
| `required`, non-streaming | `get_weather {"city":"Paris"}`, `finish_reason: "tool_calls"`, no reasoning leak |
| `required`, streaming | incremental tool-call deltas, `finish_reason: "tool_calls"`, zero reasoning leak |
| named function | correct `tool_calls`, `finish_reason: "tool_calls"` |
| `auto` (control) | unchanged — reasoning separated (266 chars), tool call parsed |
| `required` + `enable_thinking: false` (control) | unchanged |
| plain chat, thinking on (control) | unchanged — content + reasoning separated |
| Messages `tool_choice: {type: "any"}` (non-streaming + streaming) | `tool_use` block with input, `stop_reason: "tool_use"` |

One observation worth recording: in `required` mode this SGLang build also enforces the JSON-schema grammar from the first output token (`reasoning_content` is empty — no thinking precedes the payload), i.e. its `--grpc-mode` path currently behaves like vLLM here. So both mainstream gRPC backends exercise the token-0 grammar cell that this PR fixes, and before this fix `required` + thinking model was broken on both. The thinking-aware path (pre-armed parser splitting at the think-end token) is preserved unchanged and unit-tested for engines that honor `require_reasoning` by delaying the grammar until after reasoning.

PR #2153, opened 2026-08-14
feat(multimodal): allow env override of per-request media count limits

FC-Li:
## Summary

Each model spec hardcodes its per-request modality limits (e.g. 10 images for Qwen-VL / Qwen3-VL / Kimi, 8 for Llama4, 4 for LLaVA / Phi-3-V) via `modality_limits()`, with no way to adjust them per deployment.

This adds three environment variables that override the declared limit at validation time:

- `SMG_IMAGE_MAX_COUNT`
- `SMG_VIDEO_MAX_COUNT`
- `SMG_AUDIO_MAX_COUNT`

They follow the existing `SMG_*_MAX_INPUT_BYTES` pattern in `media.rs`: read once and cached in a `OnceLock`; unset, non-numeric, or zero values are ignored and the spec default applies.

## Design notes

- The override **replaces** the spec's declared limit (it can raise or lower it), but it never enables a modality the spec does not declare — an image-only model still rejects video with `UnsupportedModality` regardless of `SMG_VIDEO_MAX_COUNT`.
- Count checking is extracted from the `validate_media_request` trait default method into a free function `check_media_counts` with the override lookup injected as a parameter, so no individual spec needs changes and the logic is unit-testable without touching process env.
- Operators raising the image count should also raise the downstream engine's limit (e.g. vLLM `limit_mm_per_prompt`) to match.

## Testing

- `cargo test -p llm-multimodal --lib registry::` — 48 passed, including 4 new tests covering: override raises the effective limit, override does not enable undeclared modalities, and env parsing ignores unset/invalid/zero values.


slin1237:
LGTM
please address CI failure
- DCO
- PR title
- linting

zhyncs (approved):

PR #2158, opened 2026-08-15
fix(protocols): route batched generate requests on their token ids

slin1237:
### Problem

`GenerateRequest::routing_tokens()` only returns token ids for the `InputIds::Single` form. A `/generate` request with batched `input_ids` (`InputIds::Batch`) returns `None`, so the HTTP router falls back to `extract_text_for_routing()`, which renders the token ids as decimal strings. Token-based routing policies (cache-aware token tree, prefix_hash token hashing) then route on a string rendering instead of the actual tokens and lose their affinity signal.

### Solution

Return the first sequence's tokens for a non-empty batch. A batch is dispatched to a single worker, so the first sequence is the best available affinity signal, and any token-based signal beats a decimal-string rendering. An empty batch, or a batch whose first sequence is empty, still returns `None` and keeps the existing text fallback. Requests with `text` set are unchanged.

## Changes

- `crates/protocols/src/generate.rs`: `routing_tokens()` now handles `InputIds::Batch` by returning the first non-empty sequence.
- Tests: added `routing_tokens_from_batch_first_sequence`; replaced `routing_tokens_none_for_batch_and_empty` with `routing_tokens_none_for_empty_inputs` (empty batch, batch with empty first sequence, no input).

## Test Plan

- `cargo +nightly fmt --all --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean (`Finished dev profile ... in 17m 23s`, exit 0; `--all-features` not used because it pulls opencv, which does not build locally)
- `cargo test -p openai-protocol` — all green, including the new cases:
  ```
  test generate::tests::routing_tokens_from_single_input_ids ... ok
  test generate::tests::routing_tokens_from_batch_first_sequence ... ok
  test generate::tests::routing_tokens_none_for_empty_inputs ... ok
  test generate::tests::routing_tokens_text_takes_priority ... ok
  ```
- `cargo test -p smg` (lib + all integration binaries) — exit 0, zero failures (lib: `1543 passed; 0 failed; 5 ignored`)
- `cargo build -p smg-python && cargo build -p smg-golang` — both build

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/crates/protocols/src/generate.rs
+++ b/crates/protocols/src/generate.rs
@@ -240,6 +240,12 @@ impl GenerationRequest for GenerateRequest {
         // Mirrors extract_text_for_routing priority: explicit text wins.
         match (&self.text, &self.input_ids) {
             (None, Some(InputIds::Single(ids))) => Some(ids),
+            // A batch is dispatched to a single worker; the first sequence is
+            // the best available affinity signal.
+            (None, Some(InputIds::Batch(seqs))) => seqs
+                .first()
+                .map(Vec::as_slice)
+                .filter(|ids| !ids.is_empty()),
             _ => None,
         }
     }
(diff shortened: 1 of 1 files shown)


slin1237:
Heads-up: this PR and #2160 edit the same `match` in `routing_tokens()` and will conflict on whichever merges second. Verified resolution for the combined state (both semantics compose — token ids win over text for the batch form too, consistent with #2160's invariant):

```rust
// Token ids win over text: they key routing on what the backend KV
// cache keys on. Empty ids fall back to text.
match &self.input_ids {
    Some(InputIds::Single(ids)) if !ids.is_empty() => Some(ids),
    // A batch is dispatched to a single worker; the first sequence is
    // the best available affinity signal.
    Some(InputIds::Batch(seqs)) => seqs
        .first()
        .map(Vec::as_slice)
        .filter(|ids| !ids.is_empty()),
    _ => None,
}
```

Note the batch arm drops the `text == None` guard it has on this branch — after #2160, batched `input_ids` should also beat `text`, or single and batch requests would follow different priority rules. Combined state verified: `cargo test -p openai-protocol` green, `smg` builds clean.

PR #2203, opened 2026-08-19
fix: stop words leak unparsed text into streamed content

key4ng:
### Problem

With `stream: true` and a `stop` word set, the client receives characters that are not part of the model's answer. Against Kimi-K3 the reported string is `ep|>` — the tail of K3's `<|sep|>` control token.

Two defects combine to produce it.

**`StopSequenceDecoder` withholds far more text than it needs to.** The jail retains the last `jail_max_bytes` of decoded text on every token, where `jail_max_bytes` is the length of the longest stop sequence. But only a suffix that is a *proper prefix* of a stop sequence can ever grow into a match — the rest of that window is held for nothing. Every streamed token is therefore delayed by up to the length of the longest stop string, and the window slices whatever happens to be crossing it.

**The end-of-stream flush skips the parsers.** Every chunk in the chat and messages streaming loops runs through the reasoning parser and then the tool parser before anything is written out. The `Complete` arm did not: it called `stop_decoder.flush()` and wrote the result straight to the client as assistant content. So whatever the window had sliced went out verbatim.

#### This is not a Kimi problem

The `ep|>` string is Kimi-specific; neither defect is. It was reproduced against Kimi-K3 because that is where it was reported, but both live in paths shared by every model:

- **Any model with a reasoning parser, no special tokens involved.** The flush bypasses the reasoning parser, so the held bytes are emitted as `content` whatever they are. On a truncated stream the client gets the tail of the model's *reasoning* delivered as answer text — a wrong `reasoning_content`/`content` split on DeepSeek-R1, Qwen3, GLM, Step3, MiniMax, or any other.
- **Any parser that needs its control tokens.** Parsers reporting `requires_special_tokens()` force `skip_special_tokens = false`, putting structural tokens in the decoded stream where the window can cut one in half. `kimi_k3` is one such parser, not the only one.
- **Tool calls.** The flush also bypasses the tool parser, so a partial tool-call marker sitting in the jail at end of stream lands in `content` instead of being parsed.
- **Any stop word.** The leak length tracks the longest stop sequence exactly, so a user-supplied `stop` is enough — no unusual configuration required.

The latency half is unconditional: every request that sets `stop`, on every model, has been paying up to `len(longest_stop)` bytes of streaming delay per token.

### Solution

Withhold only what could still match, and make sure nothing reaches the client without being parsed.

## Changes

**`crates/tokenizer/src/stop.rs`** — withhold only the longest suffix that is a proper prefix of some stop sequence. That is usually zero bytes, so text leaves with the token that produced it and the jail is empty at end of stream. Emitted text is unchanged; it just arrives earlier. The Aho-Corasick search window went away with it, since the buffer is now bounded by one token plus a partial match.

**`model_gateway/src/routers/grpc/regular/streaming.rs`** — the match over the response variant now *yields* the text to emit rather than emitting it, and one shared block after it runs the parsers. The `Chunk` arm hands over its decoded text, the `Complete` arm hands over the flush, and neither can bypass parsing. Ordering is unchanged: the flush still precedes the finish-reason and usage chunks. `/v1/completions` keeps its direct flush — that path has no parsers to route through, so its raw text is the response.

> Most of that second diff is re-indentation from lifting the emission out of the match arms. **`git diff -w` shows the actual change** — 88 insertions, 107 deletions.

**`e2e_test/chat_completions/test_reasoning_content.py`** — regression coverage, described below.

### Reproduction

8×B300, `vLLM 0.1.dev19262` gRPC serving Kimi-K3 (TP=8, EP on), behind this gateway with `--reasoning-parser kimi_k3 --tool-call-parser kimi_k3`.

```bash
curl -s --no-buffer 'http://127.0.0.1:8080/v1/chat/completions' \
  -H 'Content-Type: application/json' \
  -d '{"model":"kimi-k3","stream":true,
       "messages":[{"role":"user","content":"你好，你是谁？"}],
       "max_tokens":82,"temperature":0,"stop":["wtfx"]}'
```

Before: a content delta of exactly `"ep|>"`. After: no content delta at all, matching the same request without `stop`.

### Before / after

`max_tokens: 82` cuts the stream just past the think→response boundary. At that point the model is still inside its reasoning block, so `delta.content` must be empty. Varying only `stop`:

| `stop` | `delta.content` before | after |
| --- | --- | --- |
| *(none)* | `""` | `""` |
| `["wtf"]` | `"p\|>"` | `""` |
| `["wtfx"]` | `"ep\|>"` | `""` |
| `["wtfxy"]` | `"sep\|>"` | `""` |
| `["0123456789"]` | `"ink<\|sep\|>"` | `""` |

The leak is exactly as long as the stop word, which is what pins the mechanism to the jail window. Cutting one token earlier or later leaked the other markers — `"e|>"`, `"se|>"` from `<|close|>`; `"n|>"`, `"en|>"` from `<|open|>` — all `""` after the fix.

Sweeping `max_tokens` from 8 to 96: before, every stop word produced a stray content fragment at every truncation point; after, `content` matches the no-stop baseline at every one.

### No regression in stop handling

Verified for matched stops that the stop string is absent from the output, the content is an exact prefix of the unstopped baseline, the cut lands exactly where the stop word starts, and `matched_stop` is correct — including multi-byte stop words (`助手`, `，由`), which exercise the character-boundary handling.

### Tests added

`crates/tokenizer/src/stop.rs`:
- text that cannot match is emitted immediately and leaves nothing to flush
- a partial match is withheld and then released **in full** when it diverges (the `</` that never becomes `</think>` case)
- `flush()` returns a genuinely pending partial match
- a control token sharing an opening with the stop sequence is never sliced

The spanning-token test now asserts the emitted text rather than per-token `Held`/`Text` — the timing is what changed, and the output is what must not.

`e2e_test/chat_completions/test_reasoning_content.py`: generate twice at `temperature: 0`, with and without a stop word the model never emits, and require an identical reasoning/content split. Parametrized over stop-word lengths because the leak was exactly as long as the stop word.

### Suite status

`cargo test --workspace` passes. Two pre-existing conditions on `main`, unchanged by this branch and verified by reproducing them on `main`: a clippy lint at `model_gateway/src/worker/monitor.rs:825` that only fires on rustc 1.97.1 (`unneeded_wildcard_pattern`, newer than CI's toolchain — CI skips clippy in pre-commit), and two order-dependent `middleware::metrics` interner tests that fail in a full-suite run and pass in isolation.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (modulo the pre-existing `monitor.rs` lint above, which also fires on `main`)

</details>

### DIFF
--- a/e2e_test/chat_completions/test_reasoning_content.py
+++ b/e2e_test/chat_completions/test_reasoning_content.py
@@ -121,6 +121,53 @@ class TestReasoningContentAPI:
         assert len(reasoning_content) > 0
         assert len(content) > 0
 
+    @pytest.mark.parametrize("stop", [["wtf"], ["wtfx"], ["0123456789"]])
+    def test_streaming_unmatched_stop_word_does_not_change_output(self, model, api_client, stop):
+        """A `stop` word that never fires must not change what the client sees.
+
+        The stop decoder withholds text that could still complete a stop
+        sequence. When generation is cut short — here by `max_tokens`, before
+        the model ever emits the stop word — whatever is still held is released
+        at end of stream. That release used to skip the reasoning parser, so the
+        held bytes surfaced as assistant `content`: a tail of the reasoning
+        text, or a fragment of the model's own structural tokens. The leak was
+        exactly as long as the stop word, which is what the parametrize covers.
+        """
+
+        def run(stop_words):
+            kwargs = {"stop": stop_words} if stop_words else {}
+            response = api_client.chat.completions.create(
+                model=model,
+                messages=[{"role": "user", "content": "What is 1+3?"}],
+                # Cut generation off early so the stream ends mid-reasoning,
+                # with text still held back by the stop decoder.
+                max_tokens=24,
+                temperature=0,
+                stream=True,
+                extra_body={"separate_reasoning": True},
+                **kwargs,
+            )
+            reasoning, content = "", ""
+            for chunk in response:
+                delta = chunk.choices[0].delta
+                if delta.content:
+                    content += delta.content
+                if delta.reasoning_content:
+                    reasoning += delta.reasoning_content
+            return reasoning, content
+
+        baseline_reasoning, baseline_content = run(None)
+        stopped_reasoning, stopped_content = run(stop)
+
+        assert stopped_content == baseline_content, (
+            f"stop={stop!r} leaked {stopped_content[len(baseline_content) :]!r} "
+            "into content; text held by the stop decoder must still be parsed"
+        )
+        assert stopped_reasoning == baseline_reasoning, (
+            f"stop={stop!r} changed reasoning_content; a stop word that never "
+            "matches must not affect the reasoning/content split"
+        )
+
     def test_nonstreaming_separate_reasoning_false(self, model, api_client):
         """Test non-streaming with separate_reasoning=False, reasoning_content should be empty."""
 
--- a/crates/tokenizer/src/stop.rs
+++ b/crates/tokenizer/src/stop.rs
@@ -72,9 +72,6 @@ pub struct StopSequenceDecoder {
     visible_boundary_idx: usize,
     /// Buffer for partial matches (the "jail")
     jail_buffer: String,
-    /// Maximum bytes to retain in jail_buffer — equal to the longest stop sequence.
-    /// Text beyond this window cannot participate in a future match and is safe to drain.
-    jail_max_bytes: usize,
     /// Whether we've stopped
     stopped: bool,
     /// The string stop sequence that triggered the stop, if any. Set only for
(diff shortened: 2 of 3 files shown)


e2e_test/chat_completions/test_reasoning_content.py
```diff
@@ -121,6 +121,53 @@ def test_streaming_separate_reasoning_true_stream_reasoning_false(self, model, a
         assert len(reasoning_content) > 0
         assert len(content) > 0
 
+    @pytest.mark.parametrize("stop", [["wtf"], ["wtfx"], ["0123456789"]])
+    def test_streaming_unmatched_stop_word_does_not_change_output(self, model, api_client, stop):
```
CatherineSue, line 125: Can we put this in test_openai_server.py and see if we can improve the existing [`test_stop_sequences_stream`](https://github.com/smg-project/smg/blob/main/e2e_test/chat_completions/test_openai_server.py#L346) instead of creating a new test ?

CatherineSue (approved):
Only a comment about e2e tests. Other changes LGTM

PR #2313, opened 2026-08-25
fix: unblock worker removal during cache cleanup

slin1237:
## Summary

- remove workers from the routing registry before expensive policy cleanup so cleanup latency cannot strand draining workers
- batch cache-aware cleanup by model and prune unrelated string and token radix branches
- move radix and KV-event cleanup to bounded blocking workers, with permits held through blocking work
- drop the explicit 10-second policy cleanup override and avoid rebuilding all remaining policies after each removal

## Testing

- cargo +nightly fmt --all
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test


ekzhang (approved):
🙏

PR #2398, opened 2026-09-02
refactor(grpc): make transcription a first-class pipeline endpoint

hello-alexmcc:
### Problem

The gRPC router had ~250 lines of Qwen3-ASR specifics inlined, and audio transcription was architecturally wrong in two ways:

1. **A model-specific file under `routers/`** — detection, prompt construction, output parsing for one model family, in the router layer.
2. **It wrapped the pipeline instead of running through it.** Every other endpoint (chat, generate, messages, completion, …) builds its backend request as a `BuildStage` *inside* the pipeline. Transcription synthesized a `ChatCompletionRequest` in the router, called `execute_chat_for_responses` to re-enter the chat pipeline, then post-processed — request-building happened *outside* the pipeline. (Generate is the model: it isn't even a separate endpoint, it rides `Endpoint::Chat` via composite stages, and `execute_generate` just builds a context and runs the pipeline.)

### Solution — transcription as a first-class pipeline endpoint

- **`Endpoint::Transcription`** (Regular-only, like Embeddings/Classify) with its own stage trio: `TranscriptionPreparationStage` + `TranscriptionRequestBuildingStage` + `TranscriptionResponseProcessingStage`. `route_audio_transcriptions` now just calls `execute_transcription` — running the pipeline, not wrapping it. The audio→chat synthesis moved **into** the preparation stage.
- **The stages are model-agnostic.** All family knowledge lives in `crates/multimodal`'s `qwen3_asr::transcription` as a `TranscriptionFamily` trait + `FAMILIES` registry: identifiers, the language set, the `language {name}<asr_text>` prefill convention, output parsing, capability limits. Preparation resolves a family, reads its data to shape a chat-shaped request, and runs the **shared** chat preparation; response processing decodes plain text (no tool/reasoning parsing) and applies the family's parser. Adding a transcription family is a `FAMILIES` row plus the family's own knowledge — no gateway code, no router branch.
- **Confirmed the framing is not router logic**: the ChatML scaffold, the `<|audio_*|>` placeholder block, and the trailing assistant header are all produced by the model's own Jinja `chat_template` + multimodal expansion, and `continue_final_message` is a generic render feature. The endpoint carries no rendering — only the family's prefill string and output parse.
- **Shared, not duplicated**: chat preparation is factored into `prepare_chat_like` and chat request-building into `build_chat_backed_plan`, both reused by the transcription stages (chat behavior unchanged — 32 chat tests green).
- **The `routers/transcription/` adapter is removed** — no transcription-specific code remains in `router.rs`.

**No behavior change**: same detection, same synthesized request (greedy, `n=1`, `skip_special_tokens`, the language continuation), same error codes, same json/text rendering, whole-file only; transcription stays outside tenant rate limiting and retries as on the old path.

## Changes

- `crates/multimodal/src/registry/qwen3_asr.rs` — `TranscriptionFamily` trait + `Qwen3Asr` impl + `FAMILIES`, atop the existing protocol-free family functions.
- `model_gateway/src/routers/grpc/regular/stages/transcription/{mod,preparation,request_building,response_processing}.rs` — the endpoint stages + generic helpers (family resolve, response-format parse, render, audio_format).
- `model_gateway/src/routers/grpc/{pipeline,context,spec}.rs` — `Endpoint`/`RequestType`/`PreparationOutput`/`ResponseSpec`/`FinalResponse` variants, `for_transcription`, `execute_transcription`, golden parity.
- `.../stages/chat/{preparation,request_building}.rs` — factored `prepare_chat_like` / `build_chat_backed_plan` shared helpers.
- `.../grpc/router.rs` — `route_audio_transcriptions` runs the pipeline; the inlined family (~440 lines earlier + the adapter) is gone.
- `.../routers/transcription/` — deleted.

## Test Plan

- Family logic (identifiers, prefill, languages, sanitation, transcript parse) tested in `crates/multimodal`; endpoint helpers (family resolve, format parse, audio_format, render) tested in the stage module; the pipeline golden test now asserts `Transcription` builds Regular-only and is `None` for PD/EPD.
- `cargo test -p smg --lib` — 1899 passed; `cargo test -p llm-multimodal` — passes; `--test routing_tests` (124), `--test api_tests` (107), `--test spec_test` (96) all pass.
- `cargo clippy -p smg -p llm-multimodal --all-targets -- -D warnings` and `cargo +nightly fmt --all` clean.
- End-to-end gate: the transcription leg on `e2e-1gpu-chat (tokenspeed)`. Behavior preservation is by construction — same synthesized request through the same chat pipeline, now as a stage.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes — run as `-p smg -p llm-multimodal --all-targets`; full-workspace `--all-features` needs the opencv toolchain not present locally, CI owns it
- [x] (Optional) Documentation updated — module docs on the endpoint stages and the family registry

</details>


model_gateway/src/routers/grpc/regular/stages/transcription/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+/// Resolve the transcription family serving `model_id`, or `None` when no
+/// family matches (the endpoint then rejects the request). Detection is the
+/// family's own identifier check against the model id and — for deployments
+/// serving under a neutral alias — the workers' model ids and model-identity
+/// label values.
+pub(crate) fn resolve_family(
+    worker_registry: &WorkerRegistry,
+    model_id: &str,
+) -> Option<&'static dyn TranscriptionFamily> {
+    FAMILIES.iter().copied().find(|family| {
```
CatherineSue, line 54: this dependency is inside qwen3_asr. is it better to put it in transcription stages instead?

hello-alexmcc: Good catch — the seam didn't belong inside `qwen3_asr`. Moved the contract (`TranscriptionFamily`, `FAMILIES`, the validation-error types) to a neutral `llm_multimodal::registry::transcription` module in 108e4518; `qwen3_asr` now only implements it, and the stages import nothing model-named.

I kept the trait on the multimodal side rather than defining it in the transcription stages because the implementations are model knowledge living in `llm-multimodal` (protocol-free), and that crate can't depend on the router — a router-owned trait would pull per-model impls back into `model_gateway`, which is what this PR removes. The stages keep only the generic `resolve_family` lookup over the registry.

PR #2421, opened 2026-09-05
fix(tokenizer): Kimi-K3 prompt segments and skip_special_tokens in the tiktoken backend

syuoni:
### Problem

Kimi-K3 prompts and responses served through smg's gRPC path diverged from the checkpoint's own tokenization (tokenization_kimi.py + encoding_k3.py, which is also what vLLM and SGLang serve) in two ways:

1. TiktokenTokenizer::decode ignored skip_special_tokens. Every special token in the generated ids was rendered into content, so a request whose stop is not the EOS id (ignore_eos, custom stops) received the <|end_of_msg|> terminator as text on every turn. A client that feeds that text back as history sends the model a control token it never meant to.
2. The K3 renderer built one flat string and the tiktoken backend encoded it with all special tokens enabled. Any marker string inside message text (an earlier assistant turn's <|open|>think<|sep|>..., a user quoting a control token) became a control id, and adjacent attribute pieces such as =".hidden" were BPE-merged across the boundary. The reference build_chat_segments + _encode_chat_segments encode a list of EncodeSegment { text, allow_special }: structural markers with special tokens allowed, every other piece (tag names, attribute pieces, message text of every role, reasoning, tool arguments) as ordinary BPE, each piece on its own.

### Solution

- parse_added_tokens_decoder keeps the special flag; decode strips exactly the special: true ids when asked. The special: false markers <|open|>/<|close|>/<|sep|> stay in the text, which is the HuggingFace definition of the flag.
- PromptSegment { text, allow_special }, Encoder::encode_segments, Tokenizer::apply_chat_template_segments, with defaults that reproduce the flat behavior for HF/Jinja/mock backends. The K3 renderer emits the reference's _control/_text segments one for one; the flat apply_kimi_k3_xtml* entry points are their concatenation, so the existing golden fixtures are unchanged. The tiktoken backend encodes text segments with encode_ordinary. Media anchors (<|media_pad|>) inside content remain control segments for prompt expansion.
- Gateway: ProcessedMessages gains segments; the chat and messages preparation stages encode them (encode_segments_blocking, same offload policy). text stays as the joined string for logs and original_text.

## Changes

- crates/tokenizer/src/tiktoken.rs: skip_token_ids, decode filter, encode_segments, apply_chat_template_segments for the K3 renderer.
- crates/tokenizer/src/traits.rs, lib.rs: PromptSegment, join_segments, the two trait hooks with defaults.
- crates/tokenizer/src/encoders/kimi_k3_xtml.rs: segment emitters; render_kimi_k3_xtml_segments[_with_effort_default].
- crates/tokenizer/src/cache/mod.rs: forwards; a single control segment still goes through L0/L1, mixed segments are encoded uncached by the inner tokenizer.
- model_gateway/src/routers/grpc/{mod.rs,utils/chat_utils.rs,utils/message_utils.rs,utils/mod.rs,regular/stages/chat/preparation.rs,regular/stages/messages/preparation.rs}: segment plumbing.
- Tests: tiktoken decode/skip and segment encoding; K3 renderer segment structure; tests/kimi_k3_renderer.rs::segment_encoding_matches_vendor_token_ids with tests/fixtures/kimi_k3/k3_render_ids_fixtures.json (ids recorded from the checkpoint's apply_chat_template(tokenize=True) via transformers with trust_remote_code).

Known follow-up: K3 prompts are now encoded piecewise and bypass the L0/L1 tokenizer cache; a cache-aware segment path is a natural next step.

## Test Plan

Unit and integration tests (aarch64, in the TokenSpeed runner container):

```
$ cargo +nightly fmt --all -- --check
(no output)

$ cargo clippy --workspace --all-targets --features smg/vendored-openssl -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5m 41s

$ cargo test -p llm-tokenizer
     Running unittests src/lib.rs
test result: ok. 187 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.33s
     Running tests/chat_template_format_detection.rs
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/chat_template_integration.rs
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/chat_template_loading.rs
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/deepseek_renderer_detection.rs
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/kimi_k25_renderer_detection.rs
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
     Running tests/kimi_k3_renderer.rs
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
     Running tests/qwen2_vocab_merges.rs
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/qwen3_asr_bpe_parity.rs
test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/tiktoken_integration.rs
test result: ok. 0 passed; 0 failed; 12 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/tokenizer_cache_correctness_test.rs
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.23s
     Running tests/tokenizer_integration.rs
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.04s

$ cargo test -p smg --lib --features vendored-openssl
test result: ok. 1924 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out; finished in 11.18s
```

Reference parity against the real checkpoint (`nvidia/Kimi-K3-NVFP4`, same tokenizer files as `moonshotai/Kimi-K3`). The fixture holds the ids the checkpoint's `apply_chat_template(tokenize=True)` produces for five prompts: plain user turn, system + assistant history with reasoning, marker strings injected into user and assistant text, a tool call round trip with a marker in the tool result, and a `name=".hidden"` attribute. The test asserts the segment encoding equals those ids for all five, and that the flat encoding differs exactly for the three cases with markers or the punctuation attribute:

```
KIMI_K3_MODEL_DIR=<snapshot dir> cargo test -p llm-tokenizer --test kimi_k3_renderer segment_encoding_matches_vendor_token_ids -- --nocapture
test segment_encoding_matches_vendor_token_ids ... ok
```

Before/after for the decode fix, K3 config (`[EOS]` and `<|end_of_msg|>` are `special: true`, `<|open|>` is `special: false`), `decode(ids, skip_special_tokens=true)`:

- before: `a[EOS]<|open|>b` (flag ignored)
- after: `a<|open|>b`

covered by `tiktoken::tests::test_skip_special_tokens_drops_only_special_flagged_ids` and `test_skip_special_tokens_holds_text_in_incremental_decode`.

<details>
<summary>Checklist</summary>

- [x] `cargo clippy --all-targets --all-features -- -D warnings` passes (run with `--features smg/vendored-openssl` instead of `--all-features`: the build host has no system OpenCV for `opencv-video`)

</details>


model_gateway/src/routers/grpc/utils/message_utils.rs
```diff
@@ -96,8 +96,8 @@ pub fn process_messages(
         ..Default::default()
     };
 
-    let formatted_text = tokenizer
-        .apply_chat_template(&transformed_messages, params)
+    let segments = tokenizer
+        .apply_chat_template_segments(&transformed_messages, params)
```
CatherineSue, line 100: segments is not applied to all models. Why are we updating the APIs here?

CatherineSue: I think leak segments into public API might cause diverged user behavior. It is kind of a breaking change. But we can refine this in a follow-up.

Another thing is this PR seems to break `continue_final_message` behavior in K3. This is a bug introduced by the PR. Please confirm that, if the previous message opens a think channel, the PR still respects that behavior. Which i think it doesn't as the prefix is pushed as a text segment in this case.

crates/tokenizer/src/tiktoken.rs
```diff
@@ -481,10 +524,33 @@ impl Encoder for TiktokenTokenizer {
             .map(|input| self.encode(input, add_special_tokens))
             .collect()
     }
+
+    fn encode_segments(&self, segments: &[PromptSegment]) -> Result<Encoding> {
+        let mut ids = Vec::new();
+        for segment in segments {
+            if segment.allow_special {
+                ids.extend(self.tokenizer.encode_with_special_tokens(&segment.text));
```
CatherineSue, line 532: This extends rebuilds the whole 256-entry everytime. It is a performance regression.

PR #2497, opened 2026-09-10
feat(kv-events): route using reported KV cache groups

ai-jz:
> **Update — September 11, 2026**
>
> This now uses Simo's Radix core. For this round, we can focus on the group matching rules and the small extension that keeps ancestry available while live cache evidence depends on it.
>
> Once #2500, #2437 and #2438 land, I'll rebase and integrate the group support with the shared index. This remains a draft for design feedback; that integration still needs validation.

> **Experimental results:** See the [targeted cache-selection test and AgentX workload replay](https://github.com/smg-project/smg/pull/2497#issuecomment-5687971342), including the vLLM fixes used and how the tested build differs from this PR.

## Description

This is a **draft for design feedback** on group-aware cache routing. The implementation and tests make the proposal concrete; feedback on the cache semantics, event contract and scope is welcome before merge readiness.

### Problem

The legacy KV-event index models a contiguous cached prefix. Hybrid models need different cache data at the same prefix position: full-attention keys/values, a sliding-window history, or a saved recurrent state. A **KV cache group** is vLLM's set of model layers sharing one cache block table; several distinct groups can have the same cache kind.

A single `min(group_maxima)` is insufficient. A Mamba state saved at a later position cannot restore every earlier position, and a sliding-window cache may have gaps. Each candidate position must satisfy all reported groups at that same position.

### Solution: select the largest common position

**The matching principle in this PR is consistent with vLLM's native hybrid prefix-cache lookup: select the largest prefix position supported by every participating KV cache group.** In [`HybridKVCacheCoordinator.find_longest_cache_hit`](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/core/kv_cache_coordinator.py#L808-L921), each cache manager accepts the candidate length or returns a shorter hit, and the coordinator reconciles the results until it reaches a common hit. The native implementation uses an iterative fixed-point search; SMG computes full-attention coverage bounds and intersects ordered sliding-window ranges and Mamba checkpoints against the same common-position criterion.

The per-kind reuse conditions below correspond to vLLM's native `find_longest_cache_hit` implementations. Links are pinned to the engine revision used in validation:

| vLLM event kind | Condition at a candidate prefix position | Native vLLM lookup |
| --- | --- | --- |
| `full_attention`, `mla_attention` | All attention keys/values from the start through that prefix are cached. | [`FullAttentionManager`](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/core/single_type_kv_cache_manager.py#L718-L813): scan the cached prefix from the beginning. |
| `sliding_window` | The recent attention history is continuously cached. For ordinary causal attention, window `W` includes the current token, so up to `W - 1` previous tokens are needed. | [`SlidingWindowManager`](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/core/single_type_kv_cache_manager.py#L930-L1037): search backward for a sufficient contiguous history. |
| `mamba` | A saved state exists at exactly that prefix position. This also covers the tested GDN groups using vLLM's `MambaManager`. | [`MambaManager`](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/core/single_type_kv_cache_manager.py#L1457-L1532): search backward for an eligible saved state. |

The [SMG matcher](https://github.com/smg-project/smg/blob/030475625f9b7070939fec227edc45632cc0b8a8/crates/kv_index/src/group_cache.rs#L298-L400) applies these conditions to resolved received reports and returns the largest endpoint supported by **every KV cache group reported so far**. If none qualifies, existing routing fallback applies. The same rule works when a model uses only one kind or a different combination; it does not require every model to contain all three kinds.

**The selection principle is shared; matching live-engine token counts additionally requires the same participating groups, cache state and boundary eligibility.** vLLM queries its actual cache and applies scheduler alignment, partial-hit and speculative-decoding rules. SMG has the received-event view and the three conditions above. Those information/eligibility boundaries, including the observed Gemma sparse-report limitation, remain explicit below.

KV cache group matching with retained SWA spans and Mamba checkpoints

In the illustrative 32,768-token request, the saved state at 29,568 lacks the full-attention prefix. The state at 25,344 has that prefix, but its required SWA history crosses a gap. **16,896** is the largest position accepted by all three groups, inside a retained SWA span of 11,648 tokens. The earlier saved states cannot improve that result. The combined example is hypothetical; the model cards show actual tested configurations, including Gemma's 1,024-token window.

### Design principle: reuse semantics are independent of retention policy

For the supported cache kinds, reuse is defined by the state required to resume computation. SMG selects the largest position supported by every observed KV cache group. Retention and eviction change which positions qualify, while the matching rule stays the same. This principle follows from the computation requirements and is consistent with vLLM's native lookup.

Full attention needs its earlier prefix; SWA needs its local history; a recurrent layer needs the corresponding saved state.

Eviction withdraws cached blocks or states. It changes which positions satisfy each condition, but does not require a new matching rule for each eviction pattern. Retention density and earlier requests can likewise change SWA spans and Mamba points without changing the intersection rule. This small separation between **reported state** and **reuse conditions** is the source of the design's simplicity and expected maintainability. It does not, by itself, prove that every implementation or producer report is correct.

We also checked retention directly. vLLM's [`prefix_cache_retention_interval`](https://github.com/vllm-project/vllm/blob/c3ccc0e957fbcb1f4b55fb10ff935cf0a114e880/vllm/config/cache.py#L157-L165) retains replay/shared-prefix boundaries at `0`, adds periodic checkpoints for positive intervals, and permits dense retention at `None`. Previously stored blocks and subsequent eviction determine what remains available. Thus SWA can retain longer spans or several fragments; Mamba checkpoints follow engine-selected positions. The extracted SWA/Mamba retention methods were identical between tested revision `56d001fa` and inspected revision `c3ccc0e9`, and their unchanged methods reproduced the figure's checkpoint selection. This is concrete compatibility evidence across those revisions, not a guarantee about every past or future version.

### Received events are the interface

The bridge preserves native group identifiers, opaque hashes, parents and token spans. Groups are discovered when their events arrive; there is no preconfigured model roster. Dense, anchored reports establish positions. Sparse reports keep their membership and token span, but missing offsets are not guessed; another report may resolve a shared hash, and unresolved reports remain pending while live evidence depends on them.

Repeated STORE reports are idempotent evidence. REMOVE withdraws the named group's key, CLEAR empties reported memberships, and known sequence gaps, disconnects or malformed input reset the observation session. Unsupported or unresolved evidence yields no group-affinity score and uses the existing fallback; it does not reject inference requests.

The names `full_attention`, `mla_attention`, `sliding_window` and `mamba` are unchanged vLLM event-kind values. MLA uses the same prefix-coverage condition as full attention. The bridge maps `group_idx` to the candidate's `KvGroupEvent.group_id`, `kv_cache_spec_kind` to `kind`, and `kv_cache_spec_sliding_window` to `sliding_window`. These are field mappings for the same concepts. Other vLLM cache kinds are outside this PR's scope.

The installed vLLM event schema selects this converter automatically through the existing KV-events path. There is **no additional opt-in flag** or configured group list. Schemas without the required group metadata retain the legacy converter, including #2495's alignment guard. Group-capable schemas may omit the later-added `locality` and `ownership` fields; absence now retains the older local-GPU semantics, while explicit remote locality and non-None ownership remain filtered. This repairs a compatibility regression in this PR; it requires no vLLM upgrade or producer change.

### Relationship to the shared prefix-cache index work

The group matching rules complement the shared-index work. The merged #2436 supplies the Radix core reused here; #2500 adds recency-based eviction, #2437 supplies the shared index service, and #2438 connects gateways to it. Those changes address how cache knowledge is maintained and shared across gateways. This PR addresses how reports from different KV cache groups jointly support one reuse position.

The implementation has two integration boundaries. The small Radix context API retains prefix identity independently of live membership and now releases unused history through the existing whole-chain collector. Its interaction with #2500's recency-based reclamation still needs joint validation. The group matcher currently lives in the gateway's `KvEventMonitor`; enabling #2438's remote-index mode skips those local subscriptions, and #2437's inspected bridge consumes legacy `events` rather than `group_events`.

The shared query currently carries block-level content hashes and returns matched blocks, while this matcher uses token-level paths and group-specific boundaries. Integration therefore needs to align the event/query contract and preserve group membership, historical parents, pending reconstruction and observation resets through replication and state recovery. Reusing the same Radix crate does not provide that integration automatically.

The plan is to rebase onto the relevant shared-index changes after they land and wire the group support into the shared ingestion/query path. The matching rules and current implementation can be reviewed now, including the two boundaries above. Shared-index integration and combined performance evaluation remain follow-up work. #2439's independent simulation harness may support that evaluation; its existing results do not establish group-aware behavior. The result remains a hint derived from received events, subject to the producer and freshness limits below.

### Alternatives considered: worker-side native prefix-cache lookup

A reasonable alternative is to extend the existing SMG gRPC integration with a query that executes vLLM's native prefix-cache lookup inside each candidate worker's EngineCore and returns its currently reusable token count for the request. This could avoid sparse-offset and full-report-membership defects in KV events and reduce duplicated matching semantics in SMG.

| Trade-off | Event-based matching in this PR | Worker-side native query |
| --- | --- | --- |
| State used | Reconstructed from received worker KV events; depends on publisher accuracy/completeness and delivery freshness. | Actual cache state and native lookup rules at query time. |
| Routing work | Local matching; no additional per-request cache-query RPC. | A pre-routing RPC plus lookup work on the queried candidate workers; candidate fan-out matters. |
| Engine integration | Event schema plus three local reuse conditions. | An EngineCore query interface, correct request/hash preparation, and audited native lookup behavior. |

The existing gRPC library can carry the request, but does not itself own EngineCore's cache. In the pinned engine, AsyncLLM uses a [multiprocess EngineCore client](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/engine/core_client.py#L79-L88); its [utility RPC mechanism](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/engine/core_client.py#L1199-L1214) is a possible transport into that process. A concrete query entry point is still needed.

Two implementation conditions deserve separate discussion:

1. **Read-only native query:** use the real scheduler/cache state and the correct request processing, native hashes and lookup rules. [`KVCacheManager.get_computed_blocks`](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/core/kv_cache_manager.py#L245-L311) can emit `BlockStored` events in full-report mode, so calling it unchanged is not automatically side-effect-free. A query path must audit/isolate such effects; that is a vLLM interface/implementation dependency.
2. **Result lifetime:** the result describes query-time state; eviction can occur before scheduling. As a routing hint, it can accept that boundary and let actual scheduling re-check cache availability. A reservation is not inherently required. Only a promise that the hit remains valid through execution requires stronger coordination, such as a reservation/lease or atomic query-and-admission. That is a concurrency contract choice, not a vLLM bug identified here.

We have **no latency or throughput measurements for this alternative** and have not rejected it experimentally. With a small candidate-worker set, it may be the better trade-off. The draft invites feedback on that choice; implementing the query or a reservation mechanism is outside this patch.

## Changes

One group matcher, a thin Python/protobuf bridge, and monitor/policy integration provide the complete routing behavior. The matcher now reuses `smg-radix-tree` for exact historical prefix identity and traversal. Its small context API appends from a stable parent and returns only completely matched, explicitly registered endpoints. `GroupCache` owns native-key memberships, sparse reconstruction, stream lifecycle and group reuse conditions.

Request tokens are prepared once across worker queries. Each worker retains its own tree; this does not yet provide shared-index integration or cross-worker traversal. Cleanup runs after each event batch: live descendants keep their ancestry, unused pending reports are discarded, and unneeded contexts are released through Radix's existing whole-chain collector. Native identities are removed when their backing chain is collected.

The patch has **2,731 changed text lines across 29 files, including two image files** (2,591 additions and 140 deletions). The PNGs are counted as files, not text lines. This exceeds the repository's small-PR guideline, so feedback on reviewing this end-to-end change together is explicitly requested.

`kv-index` is bumped to 1.5.0 with a dependency on `smg-radix-tree` 0.1.1; the release workflow publishes Radix first.

## Test Plan

Confidence comes from three separate layers:

| Layer | What it establishes |
| --- | --- |
| Deterministic regressions | The common-position rule, sparse checkpoints, group discovery, reconstruction, removal and stream lifecycle behavior. |
| Cross-language/native fixtures | Actual bridge/wire/consumer integration; 159 snapshot cases, of which 63 are native-comparable and 96 check the explicitly received-state contract. |
| Historical real vLLM model runs | The earlier candidate's specific Qwen and Gemma routing/replay results below, including a preserved failing long-SWA case; these do not validate the current Radix implementation on GPU. |

A minimal model-free regression demonstrates why independent maxima cannot be minimized:

```bash
cargo test -p kv-index sparse_joint_resume_is_24_even_though_independent_maxima_are_32_and_40
```

The groups' independent maxima are 32 and 40; their latest common supported position is 24. Additional tests cover later group discovery, sparse positions learned across groups, pending-parent resolution and reclamation, shared native identities, same-batch eviction/append, repeated STORE/removal, CLEAR, sequence gaps and malformed final frames. A Python-to-protobuf-to-Rust fixture changes worker selection after removal of a Mamba checkpoint.

<details>
<summary>Commands and recorded CPU validation</summary>

**September 11 lifecycle update:** the frozen candidate passed the complete serial Rust workspace suite on a CPU pod: **5,539 passed, 0 failed, 43 ignored** (including 4 passing doctests). Nightly formatting, workspace/all-target Clippy using the documented alternative without OpenCV, and applicable file hooks passed. All 1,417 source/symlink paths matched before and after validation. The 23 GroupCache and 3 context tests include the new lifecycle cases and two independent 1,048,576-token prefixes on one worker; these focused results are included in the workspace total. Python/protocol code is unchanged by this update; its earlier validation is recorded below.

A local optimized Rust diagnostic exercised one worker with four live prompts, 65,536 or 1,048,576 tokens each, with independent prefixes or a shared 1,024-token prefix. Across 256 STORE rounds (four initial fills and 252 replacements), retained history plateaued at 4,194,304 and 5,238,784 token units in the million-token cases; the shared-prefix case includes an existing dead tail. All four cases emptied their logical history on CLEAR, and 1,024 removed unresolved reports left zero pending tokens. Native event replay confirmed the documented lost-parent miss and preserved hits when full-attention evidence or a same-batch parent remained available. These are finite CPU lifecycle checks, not a total memory bound or shared-index/GPU validation.

For the previous published snapshot, the complete serialized workspace suite passed **5,535 tests, 0 failed, 43 ignored across 129 result blocks**. Nightly formatting, workspace/all-target Clippy using CONTRIBUTING.md's documented alternative without OpenCV, source-built Python bindings/native import, Ruff and applicable file hooks passed. The available Python suite passed **178**, with two skips for unavailable NumPy-dependent TokenSpeed loads and vLLM; `test_tokenspeed_dp_rank_pin.py` was explicitly excluded because its NumPy/Torch/TokenSpeed runtime dependencies were unavailable.

The original default-parallel workspace run stopped on one unchanged HTTP metrics test: the process-global interner grew by 102 against its slack limit of 100. That same test binary passed the test in isolation, and the complete suite subsequently passed with one test thread. The original failure is retained as a validation caveat; no assertion was weakened or test removed to get this result.

The previous snapshot was validated on main `3d36721e` with all 1,416 source paths matching before and after the checks. A final packaging-only change bumped `kv-index` to 1.5.0 in three Cargo files; all Rust/Python bytes remained identical. The final locked/offline dependency metadata and package suite passed separately (**254 unit tests plus 1 doc test**), with all 1,416 final source paths matching again. Those 255 package results are not added to the full-suite total. That snapshot's GroupCache source SHA-256 is `a016f3456e6bcae62cd171feb97aa6be4ec58c59ea2725bd0fd5cc3545bb7cdb`.

```bash
cargo +nightly fmt --all -- --check
cargo clippy --locked --workspace --all-targets -- -D warnings
cargo test --locked -- --test-threads=1
make python-dev JOBS=96
PYTHONPATH=grpc_servicer:crates/grpc_client/python python -m pytest -q -rs \
  grpc_servicer/tests --ignore=grpc_servicer/tests/test_tokenspeed_dp_rank_pin.py
cargo metadata --locked --offline --format-version=1 --no-deps
cargo test --locked -p kv-index -- --test-threads=1
```

Protocol bindings were regenerated through their owning build. The new core API, group matcher, Python/protobuf boundary and monitor/policy tests ran in the validation above; all 159 fixture snapshots remain. A separate mutation check removed the sequence-width guard in memory: the old doubly-invalid input still passed, while the corrected valid-payload input failed as intended. Source hashes were unchanged by the mutation experiment.

</details>

The ten earlier STORE/REMOVE compatibility cases explicitly distinguish absent attributes from attributes set to None or LOCAL. Before repair, six failed and four passed; all ten now pass. A separate native-schema audit exercised real msgspec encode/decode with unchanged event-class definitions from five pinned vLLM revisions: the legacy-converter boundary, earliest array-form group events without either optional field, locality-only map events, and modern events with ownership. It also checked CLEAR, hash representations, default/tuple extra keys, empty stores and explicit remote/owned filtering. No additional missing accessed field was found in those representative schemas. The exact version/image of the reported failing deployment remains unconfirmed; this is bounded schema compatibility evidence, not a deployed integration rerun.

The model runs below were recorded for the earlier candidate; they are historical correctness/routing evidence, not GPU validation of the current Radix implementation.

Pinned engine: vLLM `56d001faf0f53c72fcedbbdd77e5418f68fe7494`.

| Actual model/groups | HTTP prompt | Native cold → warm reuse | Gateway result |
| --- | ---: | ---: | --- |
| Qwen3.6-35B-A3B: one `full_attention` + three GDN `mamba` groups; all block size 1,056 | 2,049 | 0 → 1,056 | Warm `event_hit` passed. |
| Gemma4-12B: one full group, block 64; five SWA groups, block 16/window 1,024 | 1,025 | 0 → 1,024 | Warm `event_hit` passed. |
| Same Gemma configuration | 2,049 / 8,193 | 0 → 2,048 / 8,192 | Gateway fell back: sparse positions unresolved. |

Both models passed 23 received-state replay queries each. All tested HTTP pairs returned 200 and identical output IDs. Runs used BF16, text-only requests, 2,048-token prefill chunks, a 16K context limit and sequential single-worker checks; the Gemma prefill setting was not reduced to hide the limitation. These are correctness/routing checks, not a performance benchmark. The long Gemma cases remain failed affinity checks despite successful inference.

Rust CPU tests varied cached prompt length from 1K to 1M tokens on one worker, and worker count from 1 to 1,000 at 64K tokens per worker, using one cached prompt per worker with full attention and one Mamba checkpoint. Query-time growth was consistent with the expected prefix traversal and sorting costs. Checking every worker still becomes expensive at large scale. These measurements and the complexity analysis apply to the current independent per-worker matcher; they do not predict performance after shared-index integration or end-to-end serving latency.

### Limits, rollout and follow-up

There are two distinct correctness questions: whether SMG correctly applies received events, and whether those events fully and promptly describe the engine. This proposal addresses the first. The score is resolved received cache evidence, not a reservation or the engine's exact current reusable-token count.

The investigation identified two different producer issues; they should not be conflated with router matching correctness:

| Producer issue | Consequence and evidence | Upstream action / current boundary |
| --- | --- | --- |
| **Missing positions in sparse STORE events** | Null or masked blocks are omitted from `block_hashes`, while `token_ids` still spans the logical range. Ordinal pairing gives incorrect positions; unresolved offsets also explain the long Gemma misses. | #2495 skips ambiguous conversion in the legacy path. [vLLM #44451](https://github.com/vllm-project/vllm/issues/44451) and [#44488](https://github.com/vllm-project/vllm/pull/44488) track this defect. The inspected #44488 proposal emits contiguous runs and carries skipped-prefix context in `skipped_parent_block_hash`, `skipped_token_ids` and `skipped_extra_keys`; it has moved beyond the initial `block_offsets` proposal. Consume the agreed upstream format and rerun captured-trace/GPU integration after it lands. This PR does not guess missing positions or adopt the unmerged extension. |
| **Full reporting can advertise blocks absent from a sparse group** | In the pinned native lifecycle reproduction, blocks 0 and 2 are absent and only 1, 3, 4 and 5 are cached. `emit_cached_block_events` nevertheless reports all six logical hashes. The event has dense-looking lengths, so #2495's alignment check cannot detect this; the consumer can overestimate cache availability. | Reported with executable regressions in [vLLM #52371](https://github.com/vllm-project/vllm/pull/52371), section 2. The inspected revision adds tests only; it does not contain a production fix. The [pinned full-report helper](https://github.com/vllm-project/vllm/blob/56d001faf0f53c72fcedbbdd77e5418f68fe7494/vllm/v1/core/block_pool.py#L373-L445) still needs to report actual per-group cached membership and preserve retained entries' positions. This is separate from #44488's proposed new-store position fix; validate sparse Mamba/SWA membership with the producer repair. |

The second finding is established by executing the extracted native helper against a constructed sparse cache, not by the Qwen/Gemma HTTP runs. Neither SMG change claims to make inaccurate full reports exact. The intended follow-up is to repair and validate the producer contract upstream, then strengthen integration coverage here.

The [upstream minimal regression](https://github.com/vllm-project/vllm/blob/da79a4ffca9e05023fae39db16c5af384f084e14/tests/v1/core/prefix_cache/test_mamba_align_sparse_keys.py#L212-L243) uses ordinary Mamba `align`, block size 4 and an 8-token producer prefix, with no partial hash matching. The token-8 state exists and the token-4 state does not, yet a consumer's `kv_cache_report_mode="full"` lookup publishes the token-4 hash. The test runs through `KVCacheManager` on CPU and marks the desired membership assertion as strict `xfail`.

Completeness and freshness also matter: unreported groups or stale removals can overestimate current engine state. The live rank-zero subscriber does not recover publisher replay, and silent restarts or missing final events are not fully detectable. Native scheduler alignment, partial-hit rules and speculative decoding eligibility are not reconstructed from these events. CPU/remote/offload reports, nonzero ranks and namespaced requests do not establish supported plain-token group affinity.

**History lifecycle:** there is no additional aggregate capacity limit or capacity-triggered reset. Unused history follows Radix's existing whole-chain reclamation; dead tails within live chains may remain, so this is not a total memory bound. If all supporting history has been collected, a later child waits for a usable parent report. This can lose affinity—for example, a Mamba checkpoint whose parent was removed in an earlier event batch—but never turns an unknown parent into a shorter rooted prefix.

**Scope and fallback:** only local-GPU, rank-zero reports establish supported group affinity; engine-specific reuse eligibility is not reconstructed. If any eligible worker uses group mode, namespaced requests use load routing for the whole candidate set. A subscription that loses its group marker stays unavailable until valid group batches or re-registration restore it.

Deploy the updated bridge and gateway together. Older gateways ignore the additive fields and receive no group-affinity evidence from those batches. Before broader rollout, canary cold/warm requests and stream-disconnect fallback on the intended model; roll back bridge/gateway versions if behavior regresses.

### Feedback requested

1. Does the largest-common-position rule and its received-event boundary express the cache evidence SMG should use?
2. Are discovery from events, additive protocol fields and automatic schema selection the right integration choices?
3. Can the implementation be made materially simpler, or should any independently useful part be reviewed separately?
4. Would a read-only, worker-side native prefix-cache lookup exposed through the existing gRPC integration be a better trade-off than reconstructing cache affinity from events?
5. Are there missing correctness cases within the supported cache kinds?


ai-jz:
### Targeted cache-selection test and AgentX replay

FYI @slin1237.

The targeted Qwen test measures the benefit in the cache-selection scenario this PR addresses, even with original vLLM. AgentX workload replay measures the overall impact: Qwen results were mixed; Gemma improved with [vLLM PR44488](https://github.com/vllm-project/vllm/pull/44488) and [PR55092](https://github.com/vllm-project/vllm/pull/55092), while the same SMG candidate without those fixes regressed.

Targeted cache-selection benefit, followed by overall impact in Qwen and Gemma AgentX workload replay

**Tested build:** both PR2497 columns include the same local support for the proposed vLLM event format, beyond the published head. This is evidence for the proposal, not qualification of the published head unchanged. Gemma's observed benefit relies on the repaired event semantics; the fixes were tested together, so their individual contributions are not isolated.

<details>
<summary>Test setup and interpretation</summary>

Original: SMG `07c8e829` + vLLM `56d001fa`, without an extra event filter. Candidate: SMG `03047562` plus the local event-format support. The patched version changes only vLLM, adding [PR44488 at efc443a](https://github.com/vllm-project/vllm/commit/efc443a645f4986bc2f458d74ebf841e158a4813) and [PR55092 at 7a76be1](https://github.com/vllm-project/vllm/commit/7a76be1f37e745668149515d74bc550e0195146a).

Four workers per AgentX run: Qwen/Qwen3.6-35B-A3B, two H100s per worker; google/gemma-4-12B-it, one H100 per worker. Both used BF16 and a 262,144-token context limit. Each version replayed the same corpus/settings from fresh engine caches; 3,600 seconds of dispatch plus about 900 seconds of grace, one run per version. Parallel sessions are agent trajectories, not worker count.

Cache hit rate and mean TTFT use common successful turns: 9,926 / 6,525 / 1,544 for Qwen-64 / Qwen-256 / Gemma-64. Cache hits are vLLM's cached prompt tokens at first prefill, before any later recomputation; throughput covers the full run. Qwen-256 had four timeouts per version. Gemma had 12 / 12 / 11 timeouts, 2 / 1 / 1 cancellations, and warmup parsing failures; its mean latency gain was concentrated in longer waits. These limits preclude a general speedup claim.

The targeted Qwen test used two workers and eight target requests per version, each with 57,128 input and 64 output tokens. It tests choosing usable cached state instead of a longer prefix that cannot resume the request; it does not establish how often that case occurs in real traffic. No original-SMG + patched-vLLM comparison or shared-index integration was tested.

</details>

PR #2539, opened 2026-09-15
feat(multimodal): support multimodal inputs over the TokenSpeed ZMQ backend

slin1237:
### Problem

Multimodal requests to TokenSpeed workers over the direct ZMQ transport were rejected at assembly (`multimodal inputs are not supported over the {runtime} ZMQ backend yet`), paired with a second rejection in `translate_request_tokenspeed` — because the Rust wire struct `TokenizedGenerateReqInput` modeled only the 11-element text prefix and had no multimodal slot, even though the engine's Python struct carries `multimodal_inputs` (wire index 22) and its msgpack codec fully supports tensor payloads.

### Solution

Complete the sender half of the wire and remove both paired bails:

- **`protocol/tokenspeed/multimodal.rs` (new)**: `MultimodalInputs` (7-slot) and `MultimodalDataItem` (10-slot) positional wire structs, tensors as the engine's `(dtype, shape, ext-3)` tuples via the shared `WireTensor` codec, and `mm_pad_value()` reproducing the engine's `set_pad_value()` formula bit-for-bit.
- **`TokenizedGenerateReqInput`**: extended through wire index 22 for multimodal requests (neutral defaults for the unmodeled fields in between); text requests keep the historical 11-element prefix.
- **`translate_request_tokenspeed`**: does the engine `InputProcessor`'s job, which the msgpack transport bypasses — derives each item's pad value from its content hash (same u64 LE fold the gRPC servicer uses), substitutes it into the inclusive placeholder ranges of `input_ids`, ships originals as `input_ids_unpadded`, and maps the proto↔engine modality enums (which disagree: proto AUDIO=2/VIDEO=3 vs engine VIDEO=2/AUDIO=3). Inline-only tensor payloads, with byte-length validation since malformed frames are silently dropped engine-side.
- **`assemble.rs`**: ZMQ TokenSpeed arm reuses the gRPC assembly with SHM off (inline wire); shared `spawn_tokenspeed_assembly()` helper keeps the SHM-cleanup cancellation dance in one place.
- **`zmq_tokenspeed_mm`**: `into_proto(false)` — no RDMA staging on a wire whose engine has no puller.

Known limitation: MRoPE position tensors are not derivable gateway-side yet; the slots ride as nil and MRoPE models degrade to 1-D positions (engine warns).

## Changes

- `crates/engine_zmq_client/src/protocol/tokenspeed/multimodal.rs` (new) + `request.rs` arity extension
- `model_gateway/src/routers/grpc/zmq_client.rs`: multimodal translate + validation + tests
- `model_gateway/src/routers/grpc/multimodal/assemble.rs`: ZMQ arm, shared spawn helper, V4.1-layout test with exact `uint32` dtype pins
- `model_gateway/src/routers/grpc/backend_client.rs`: `into_proto(false)`

## Test Plan

- **Cross-language round-trip**: the Rust-encoded multimodal request vector (pinned in `mm_request_emits_wire_arity_through_multimodal_inputs`) was decoded field-for-field by TokenSpeed's own `MsgpackDecoder` (msgspec 0.21.1, tokenspeed `eaf66b5b`): pad-substituted `input_ids`, `input_ids_unpadded`, bfloat16 feature tensor, uint32 grid values, inclusive offsets — and the engine's `set_pad_value()` independently reproduces the precomputed pad value.
- `cargo +nightly fmt --all` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean (full workspace).
- `cargo test -p engine-zmq-client`: 94 passed. `cargo test -p smg --lib tokenspeed`: 33 passed (incl. new translate tests: pad substitution, error paths for OOB placeholders / missing hash / byte-length mismatch).
- Known pre-existing failures, reproduced on clean main and unrelated to this diff: 2 `middleware::metrics` interner tests (full-suite parallelism artifact) and 3 `test_client_connect_invalid_endpoint` tests in `smg-grpc-client` (devbox transparent proxy accepts bogus endpoints; crate untouched here).
- E2E pending engine-side vision: tokenspeed#1567 (DeepSeek-V4.1 vision inputs) is about to land, which makes this path exercisable end-to-end.

<details>
<summary>Checklist</summary>

</details>


slin1237:
Confirmed the 🔴 finding — `translate_request_tokenspeed` still rejects `mm_inputs` and the Rust `TokenizedGenerateReqInput` has no multimodal slot (the description's wire-struct claim conflated the engine's Python struct, which has `multimodal_inputs`, with this crate's Rust mirror, which does not). As-is this arm just moves the rejection later and makes it more expensive. Converting to draft while reworking: either completing the Rust wire half (arity extension + mm encoding + removing the paired rejection, with Rust↔Python msgpack round-trip tests) or restoring the paired bail and landing only the honest subset. Will also pin the dtype asserts to `uint32` and extract the shared spawn/cleanup helper per the nits.

slin1237:
Review feedback addressed in `fce6c61e`:

- **🔴 wire gap**: fixed by completing the sender half rather than restoring the bail — new `protocol/tokenspeed/multimodal.rs` wire structs, `TokenizedGenerateReqInput` extended through index 22, `translate_request_tokenspeed` now performs the bypassed InputProcessor work (pad-value derivation via the engine's exact formula, inclusive-range pad substitution, `input_ids_unpadded`), and the `mm_inputs.is_some()` rejection + its test are replaced with positive translation tests. Validated by decoding the Rust-encoded frame with TokenSpeed's own `MsgpackDecoder` — including `set_pad_value()` agreeing with our precomputed values; the vector is pinned in-tree.
- **🟡 dtype asserts**: pinned to exactly `"uint32"` for all three grid/types tensors.
- **🟡 duplicated spawn dance**: extracted `spawn_tokenspeed_assembly()`; both arms now share the SHM-cleanup cancellation path.
- **🟡 arm-level shm test**: not added as suggested — `ZmqEngineClient` has no socketless constructor, and the invariant now fails loudly instead of silently: if a refactor drops the `shm_enabled: false` override, assembly emits SHM handles and `wire_tensor_tokenspeed` rejects them (`ZMQ multimodal requires inline tensor payloads`) before anything reaches the engine. Happy to add socket scaffolding in a follow-up if you'd still prefer the direct test.

slin1237:
Second round addressed in `149e899b`:

- **🔴 mrope_positions silent degradation**: agreed that "succeeds with worse output" was the wrong default. Now gated at translate: items carrying `image_grid_thw`/`video_grid_thw` (the known MRoPE families) are rejected with an explicit `MRoPE position tensors are not derivable over the TokenSpeed ZMQ wire yet; use the gRPC transport` error — same shape as the top-logprobs bails. Remaining multimodal traffic emits a once-per-process gateway `warn!` so the gap is visible in SMG's own logs, and the module docs describe the gating. Deriving positions gateway-side (or an engine-side capability flag) stays follow-up work tied to how tokenspeed#1567 settles V4.1's position scheme.
- **🟡 Unspecified → Image**: fixed — the translate now mirrors the gRPC servicer's `_modality_from_proto` and rejects unspecified/unknown modalities. Both cases covered in `tokenspeed_multimodal_rejects_bad_items`.

Gates re-run after the changes: fmt clean, `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo test -p smg --lib tokenspeed` 33 passed, `cargo test -p engine-zmq-client` 94 passed.

PR #2541, opened 2026-09-15
feat(grpc): carry the engine's cache group on KV events

slin1237:
### Problem

Hybrid models keep several independently evicted caches per worker: vLLM calls them KV cache groups (full attention, sliding window, Mamba/GDN state). vLLM reports every block store and removal with the group it belongs to, the group's cache kind and, for a window group, the window. The bridge dropped all three. Every consumer therefore folded the groups into one position set per worker, where a window group freeing its old blocks looks like a hole in the full-attention prefix, and a recurrent group's sparse checkpoints look like a contiguous prefix. Both mis-score the worker.

### Solution

Carry the group on the wire and change nothing else. `KvBlocksStored` gains optional `cache_group`, `cache_kind` and `sliding_window`; `KvBlocksRemoved` gains optional `cache_group`. The vLLM converter sets them when the installed event schema has `group_idx` / `kv_cache_spec_kind` / `kv_cache_spec_sliding_window` (current vLLM main does), and leaves them unset for older schemas. Consumers that keep one position set per worker ignore the fields; the shared index (#2437) keeps one holder per (worker, group) and the gateway (#2438) applies the reuse rules. The alignment guard from #2495 still applies to group stores.

## Changes

- `crates/grpc_client/proto/common.proto`: three optional fields on `KvBlocksStored`, one on `KvBlocksRemoved` (additive; Go bindings untouched, unknown fields are ignored there).
- `grpc_servicer/smg_grpc_servicer/kv_events.py`: populate them from the vLLM event when present; malformed or absent group indices are not reported.
- `crates/mock_worker`, `model_gateway/src/worker/kv_event_monitor.rs`: struct literals gain the new fields (`None`).

## Test Plan

- [x] `grpc_servicer/tests/test_vllm_kv_events.py`: legacy schema reports no group; group fields ride the wire untouched (store and remove); a full-attention group has no window; malformed group indices (`None`, `-1`, `2**32`, `True`, `"0"`) are not reported; group stores still require dense alignment.
- [x] `cargo check -p smg-grpc-client -p mock-worker -p smg --all-targets`; `cargo +1.98.0 clippy -p mock-worker -p smg-grpc-client --all-targets -- -D warnings`; `cargo test -p mock-worker`.

https://claude.ai/code/session_01M7LDLMD8QeFy6MRwgxExmo


slin1237:
Closing: the vLLM feed side (event schema and converter) belongs to the group-events work in #2497, not to the index stack. The index keeps its engine-neutral lane/interval model without a wire change.

PR #2542, opened 2026-09-15
feat(radix-tree): coverage read returns each holder's runs along the query path

slin1237:
### Problem

`overlap` answers one number per holder: the consecutive covered depth from position 0. That is the shape of a contiguous prefix cache, and only that. An engine whose cache is several independently evicted position sets (full attention, a sliding window that frees its old blocks, recurrent state kept at checkpoints) stores fine in the tree, whose per-holder coverage is spans at arbitrary positions with holes allowed, and then reads back wrong: coverage that starts past position 0 scores zero, and coverage with holes scores only its first run.

Two snapshot readers had the same blind spot. A snapshot needs every held block's position on its lineage, and a holder's own keys cannot supply the uncovered head of a run that starts past 0 (a hole has no key).

### Solution

Three reads, no change to writes or invariants:

- `coverage(query, scratch, out)`: the same trie descent as `overlap` (shared as `match_path`), then per span × holder instead of the active-set intersection, emitting each holder's maximal `[start, end)` runs along the query path, sorted by (holder, start). Scratch-backed like `overlap`: a per-slot open-run index reset in O(touched), no per-query heap once warm. `overlap` is unchanged in behaviour and cost.
- `runs(holder)`: the holder's coverage as maximal runs per chain lineage, each with its keys in position order and the full content path from position 0 through the run's end, held or not. What a snapshot needs to rebuild the holder on another tree at the same positions.
- `path_contents(holder, key)`: the content path from 0 through the position of one held key.

## Changes

- `crates/radix_tree/src/chain.rs`: `match_path`, `coverage`, `runs`, `path_contents`.
- `crates/radix_tree/src/lib.rs`: `CoverageRun`, `HolderRun`, scratch fields.
- `crates/radix_tree/tests/api.rs`: six new tests.

## Test Plan

- [x] Coverage: holes, mid-path runs and divergence against `overlap`'s answer; scratch reuse across queries and retired holders; a randomized churn (shared prefixes, random removes across six holders) checking every holder's runs against the per-position key map on every stored sequence.
- [x] Runs: lineage split at a fork, the uncovered head carried on the path, keys at their positions, a sharing holder unaffected, retired holders empty; `path_contents` through forks and holes.
- [x] Full crate green (api 27; differential, fuzz, alloc gate, tree-compare unchanged); `cargo +1.98.0 clippy -p smg-radix-tree --all-targets -- -D warnings` clean.

Stacks on #2500. The index service uses all three reads in #2437.

https://claude.ai/code/session_01M7LDLMD8QeFy6MRwgxExmo


slin1237:
Closing: folded into #2437 as its first commit so the stack stays #2500 → #2437 → #2438 without a new PR.

PR #2557, opened 2026-09-16
test(grpc_client): use in-memory transport for unread frame tests

Moersity:
### Problem

`channel::tests::adaptive_window_goaways_on_unread_token_frames` consistently fails on the tested macOS machine: the h2 server observes `unexpected h2 error: connection reset` instead of a remote `GOAWAY ENHANCE_YOUR_CALM`. The same test passes on Linux with identical source and Cargo.lock. This makes the assertion depend on TCP teardown behavior while testing an HTTP/2 frame budget.

Baseline: `b669cf7990941be823acea6906085046a7d3c8db`; Rust 1.98.0 (`88d9e12ae`, 2026-08-18) on both machines.

| Platform | Before: 20 consecutive runs | After: 100 consecutive runs |
| --- | --- | --- |
| macOS 26.5.1, Darwin 25.5.0, arm64 | 20 failures, all `connection reset` | 100 passed |
| Ubuntu 24.04.4 LTS, Linux 6.8.0-124-generic, x86_64 | 20 passed | 100 passed |

### Solution

Run the unread-frame helper over `tokio::io::duplex` and connect the tonic channel through a one-shot connector. A 64 KiB buffer accommodates the 2,000 one-byte DATA frames and their headers, allowing the server to finish its write before reading the GOAWAY without depending on kernel TCP teardown behavior.

## Changes

- Replace the helper's loopback TCP transport with an in-memory duplex stream.
- Preserve the exact remote `GOAWAY ENHANCE_YOUR_CALM` assertion, the unread response body, and the fixed-window survival assertion.
- Return a connector error if the channel attempts to reconnect. No production changes or new dependencies.

## Test Plan

The baseline channel suite had 12 passed / 1 failed on macOS and 13 passed on Linux. After the change, both machines passed the entire library target, including both unread-frame tests:

```text
cargo test --locked -p smg-grpc-client --lib
macOS: test result: ok. 76 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
Linux: test result: ok. 76 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Reproduce the focused test with:

```sh
cargo test --locked -p smg-grpc-client --lib \
  channel::tests::adaptive_window_goaways_on_unread_token_frames -- --exact --nocapture
```

The repetition results above use sequential invocations of that exact test in the compiled library test executable (20 before / 100 after per platform).

Additional macOS validation:

```text
cargo +nightly fmt --all -- --check
(exit 0, no output)
cargo clippy --locked -p smg-grpc-client --lib --tests -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.69s
```

Cargo commands used `RUSTC_WRAPPER=` to bypass the local sccache/incremental compilation conflict. Linux validation used an isolated source copy and the existing build cache, preserving unrelated working-tree changes.

Validation is scoped to the affected crate and default features; the full workspace/all-features CI matrix was not run locally.

<details>
<summary>Checklist</summary>

- [ ] `cargo clippy --all-targets --all-features -- -D warnings` passes (scoped Clippy passed as recorded above)

</details>


Moersity:
Root cause confirmed with socket I/O tracing on both platforms: the macOS peer **does receive** `GOAWAY ENHANCE_YOUR_CALM` (`too_many_data_frames`). Its next read returns `ECONNRESET`, whereas Linux returns EOF. h2 propagates that later I/O error instead of returning the previously received GOAWAY.

The client closes with unread TCP data after shutting down its write side. A minimal Python socket reproducer shows the same platform-dependent reset/EOF behavior, independently of Rust and HTTP/2.

Using `tokio::io::duplex` removes this TCP teardown dependency while preserving the strict GOAWAY assertion. After the change, the test passed 100 consecutive runs on each platform, and all 76 crate library tests passed on both.

Issue #1533, opened 2026-05-24
[Bug]: WorkerService::create_worker orphans url_to_id entries; clients get 202 with worker_id that 404s on poll


CatherineSue:
## Bug Description

`WorkerService::create_worker` (`model_gateway/src/worker/service.rs:240`) reserves a `WorkerId` against the submitted URL string **before** the AddWorker workflow runs, so it can return `202 Accepted` synchronously with `Location: /workers/{worker_id}`. The 202 contract holds **only** when the workflow eventually registers the worker under the *same URL string* the reservation was keyed on. There are several real scenarios where it doesn't, and each one creates a permanently-orphaned entry in `WorkerRegistry::url_to_id` plus a `worker_id` in the 202 response that points at nothing.

Concretely:

1. **The reservation key and the registration key diverge.** `CreateWorkerStep` rewrites the URL via `normalize_url` (`workflow/steps/local/create_worker.rs:300-313`). Any input that does not already start with one of `http://`, `https://`, `grpc://`, `grpcs://` gets a scheme prepended. The reservation entry under the original URL becomes orphaned; a fresh `WorkerId` is minted for the canonical URL during `register_inner`; the client polls the original `worker_id` and gets 404.
2. **The AddWorker job submission fails after a successful reservation** (`service.rs:269-272`). The reservation is never reaped.
3. **The workflow itself fails downstream** (detection times out, gRPC build fails, etc.). The reservation is never reaped.

The orphan is only an `url_to_id → WorkerId` row with no matching `workers[WorkerId]`. It is small (one `String` key + one 16-byte `WorkerId`) but unbounded over process lifetime. The client-visible 404 is the more painful symptom.

Adjacent context: PR #1523 added a `resolve_url_to_id` canonicalization helper that makes downstream URL lookups *tolerate* the orphan (used by K8s service-discovery removal), but it does not address creation or cleanup of the orphan — and it has no effect on the 404-on-poll wart.

## Steps to Reproduce

```
1. Start SMG with default config.
2. curl -X POST http://localhost:<router-port>/workers \
        -H 'Content-Type: application/json' \
        -d '{"url":"10.0.0.5:8000"}'   # bare host:port, NOT http://...
3. Note the worker_id and Location header in the 202 Accepted response.
4. Wait for the AddWorker workflow to complete (a few seconds for detection).
5. curl http://localhost:<router-port>/workers/<worker_id_from_step_3>
   -> 404 Not Found
[...]

PR #2572, opened 2026-09-19
fix(worker): release the url reservation when AddWorker fails

aniruddh-krovvidi:
### Problem

Follow-up to #1533 (cases E1/E2, left open by #1532). `create_worker` reserves a
`WorkerId` in `url_to_id` before submitting the AddWorker job. If submission
fails, or the workflow fails later, nothing removes the entry: the client's 202
`worker_id` 404s forever and the orphan stays for the process lifetime.

### Solution

`WorkerRegistry::release_reservation(url)` drops a `url_to_id` entry only when
no worker exists under its id. `create_worker` calls it on both submission
failures; `JobQueue::process_job` calls it when an AddWorker job ends in error.
A live worker under the URL is never touched.

## Changes

- `worker/registry.rs`: `release_reservation`
- `worker/service.rs`: release on queue-unavailable and submit-failure paths; one test; a concurrent create for an in-flight URL returns 409 WORKER_CREATE_IN_PROGRESS instead of a second job
- `workflow/job_queue.rs`: release when an AddWorker job fails

Three files because the two failure paths live in two modules and share the one registry helper; 16 of the 44 changed lines are the test.

## Test Plan

`create_worker_releases_reservation_when_submission_fails`: reserve the id the
202 would carry, make `create_worker` fail at submission (uninitialized queue),
assert the reservation is gone. Fails before, passes after.

```
cargo test -p smg --lib worker::service::tests      7 passed
cargo test -p smg --lib worker::registry::tests     72 passed
cargo test -p smg --lib workflow::job_queue         7 passed
cargo fmt -p smg -- --check                         clean
cargo clippy -p smg --lib --tests -- -D warnings    0 warnings
```

<details>
<summary>Checklist</summary>

- [ ] `cargo +nightly fmt` passes (stable fmt clean; nightly not run)
- [x] `cargo clippy --all-targets -- -D warnings` passes for the crate (`--all-features` not run)

</details>


aniruddh-krovvidi:
c8e269a addresses all three review threads: `submit_if_idle` drops its claim when `submit` fails (test `failed_submit_if_idle_releases_its_claim`); `claim` only treats a same-type in-flight entry as competing, so a same-named MCP server or WASM module cannot block an AddWorker (covered in `claim_admits_one_in_flight_job_per_url`); the failure-path release is gated on `CreateOnly`, since only `POST /workers` reserves, and runs before the status turns terminal so a retry cannot reserve the same mapping mid-cleanup. A second create for an in-flight URL now gets 409 `WORKER_CREATE_IN_PROGRESS` with the id to poll instead of being merged into the first attempt, whose spec may differ. The `AppContext`-dropped branch is left alone: it only runs during shutdown, when no caller is left to observe the reservation; happy to add `Weak<WorkerRegistry>` plumbing as a follow-up if wanted.

slin1237 (approved):

